PR #67, opened 2023-01-14
Load vae and text encoder in the same way as stable-diffusion-webui

Fannovel16:
There are some models like AOM2 or Anything V4.0 don't zip all tensors into state_dict property. A error will be triggered when loading these models.


Fannovel16:
Sorry I edited the wrong file

PR #79, opened 2023-01-18
expand details in tensorboard logs

mgz-dev:
This PR adds some more information into the tensorboard logs saved during lora training

- Update tensorboard logging to track both unet and textencoder learning rates
- Update tensorboard logging to track both current and ~~moving~~ running average epoch loss
- Clean up tensorboard log variable names for dashboard formatting

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,11 @@ This repository contains training, generation and utility scripts for Stable Dif
 
 ## Updates
 
+- 17 Jan. 2023, 2023/1/17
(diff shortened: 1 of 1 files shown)


anon-1337:
Just discovered tensorboard because. Your PR is a great addition. Is it possible though to add tracking of loss per image?
Something like this (scroll to the very bottom): https://github.com/aria1th/Hypernetwork-MonkeyPatch-Extension/wiki

PR #120, opened 2023-01-27
Load train data outside of epochs loop

hitomi:
I'm trying to train some LoRA models and notice a long time gap between each epoch on my poor computer. I'm trying to findout where the time is spent and found the `train_dataloader` will load the data inside each epoch loop.

It doesn't seem necessary to reload the train data inside each epoch loop, I compared the data for each load and they appear to be equal, and the `train_dataset` and `train_dataloader` seems never change in the loop, so I try to move the load process outside the loop like this PR, it seems to work fine. (BTW, only test on `train_network.py` to train LoRA model)

I'm not yet familiar with this knowledge, so I'm not sure this thing is really correct, might need to check carefully about it.

### DIFF
--- a/fine_tune.py
+++ b/fine_tune.py
@@ -223,13 +223,16 @@ def train(args):
   if accelerator.is_main_process:
     accelerator.init_trackers("finetuning")
 
+  print(f"Loading train data")
+  preload_train_data = list(enumerate(train_dataloader))
+
   for epoch in range(num_train_epochs):
     print(f"epoch {epoch+1}/{num_train_epochs}")
     for m in training_models:
       m.train()
 
     loss_total = 0
-    for step, batch in enumerate(train_dataloader):
+    for step, batch in preload_train_data:
       with accelerator.accumulate(training_models[0]):  # 複数モデルに対応していない模様だがとりあえずこうしておく
         with torch.no_grad():
           if "latents" in batch and batch["latents"] is not None:
--- a/train_db.py
+++ b/train_db.py
@@ -200,6 +200,9 @@ def train(args):
   if accelerator.is_main_process:
     accelerator.init_trackers("dreambooth")
 
+  print(f"Loading train data")
+  preload_train_data = list(enumerate(train_dataloader))
+
   for epoch in range(num_train_epochs):
     print(f"epoch {epoch+1}/{num_train_epochs}")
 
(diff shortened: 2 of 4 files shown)


hitomi:
Thank you for your reply! I tried `--max_data_loader_n_workers` option but it still takes some loading time every time. To be honest, I don't have a clue where the time is spent, CPU or disk IO or other, I'm trying to find out about this.

And what you said about the dataset containing a lot of images will out of memory is surely a problem, I'm thoughtless about it. But maybe we can turn this into an option, like adding a `--keep-dataset-in-memory` option for some cases where we don't need to worry about memory issues?

hitomi:
Update:
Digging deep into the time usage, found that the main time cost is in https://github.com/pytorch/pytorch/blob/5976f0bdfed856ab48824bd41487318f3cf6e615/torch/utils/data/dataloader.py#L1030-L1044

```python
w = multiprocessing_context.Process(
    target=_utils.worker._worker_loop,
    args=(self._dataset_kind, self._dataset, index_queue,
          self._worker_result_queue, self._workers_done_event,
          self._auto_collation, self._collate_fn, self._drop_last,
          self._base_seed, self._worker_init_fn, i, self._num_workers,
          self._persistent_workers, self._shared_seed))
w.daemon = True
# NB: Process.start() actually take some time as it needs to
#     start a process and pass the arguments over via a pipe.
#     Therefore, we only add a worker to self._workers list after
#     it started, so that we do not call .join() if program dies
#     before it starts, and __del__ tries to join but will get:
#     AssertionError: can only join a started process.
w.start()
```

I guess maybe some arguments are really large and need more time to copy to the new process. (In my case, each sub-process will raise to 300MB+ memory usage, but I still do not understand why memory copy is so slow, maybe it's not copying simply, I had no knowledge about python multiprocessing).

And it seems like it can pass `num_workers=0` (`--max_data_loader_n_workers=0`)  to disable multiprocessing, at https://github.com/pytorch/pytorch/blob/5976f0bdfed856ab48824bd41487318f3cf6e615/torch/utils/data/dataloader.py#L384-L389 it decide to use `_SingleProcessDataLoaderIter` when `num_workers=0` else `_MultiProcessingDataLoaderIter`, and it will work fine as I expected, no process created, no time will cost, so this problem will be solved. I think this PR can close lol. Thanks for your amazing work!

hitomi:
Update:
I found `torch.utils.data.DataLoader` has an option called `persistent_workers`, if set it `True` it will reuse worker instead of create new worker each time. I test it and it can solved this problem too.

Consider to provide an option or just set it as default?

PR #140, opened 2023-02-01
Add persistent_workers options in DataLoader

hitomi:
Add persistent_workers option to reduce time gap between epoch, more detail comment in #120 

It seems have some memory problem in lower torch version (<1.10, https://github.com/pytorch/pytorch/issues/62066) so decide to make it into an option.


sashasubbbb:
Nice work, thank you for this addition

PR #165, opened 2023-02-08
Add support with multi-gpu train for train_newtork.py

Isotr0py:
Support DistributedDataParallel training for `train_newtork.py`.
Now `accelerate launch` with `--multi_gpu` argument can enable DDP training with multiple GPUs.

I've tested `train_newtork.py` with two Tesla T4 GPUs and tried to fix some bugs appeared in DDP training.
`unet`, `text_encoder`, `network` should be extracted from class `DistributedDataParallel` through attribute `.module` in DDP training.

I haven't used `train_db.py` and `train_TI.py` and didn't make any changes to them. I will test them later.

Some unnecessary code was inserted just to reduce RAM usage for my poor low-ram device, please ignore it, XD.

XLAP)% MD7YV )%SJKH@Z0C

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -1560,9 +1560,10 @@ def get_hidden_states(args: argparse.Namespace, input_ids, tokenizer, text_encod
   else:
     enc_out = text_encoder(input_ids, output_hidden_states=True, return_dict=True)
     encoder_hidden_states = enc_out['hidden_states'][-args.clip_skip]
-    if weight_dtype is not None:
-      # this is required for additional network training
-      encoder_hidden_states = encoder_hidden_states.to(weight_dtype)
+    # uncomment code may raise expected scalar type Half but found Float when using DDP
+    # if weight_dtype is not None:
+    #   # this is required for additional network training
+    #   encoder_hidden_states = encoder_hidden_states.to(weight_dtype)
     encoder_hidden_states = text_encoder.text_model.final_layer_norm(encoder_hidden_states)
 
   # bs*3, 77, 768 or 1024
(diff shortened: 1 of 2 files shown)


sheldonchiu (changes requested):
```python
from torch.cuda.amp import autocast
from torch.nn.parallel import DistributedDataParallel as DDP
```
```python
  text_encoder.requires_grad_(False)
  text_encoder.to(accelerator.device)
  if args.gradient_checkpointing:                       # according to TI example in Diffusers, train is required
    unet.train()
    text_encoder.train()

    # set top parameter requires_grad = True for gradient checkpointing works
    if type(text_encoder) == DDP:
      text_encoder.module.text_model.embeddings.requires_grad_(True)
    else:
      text_encoder.text_model.embeddings.requires_grad_(True)
  else:
    unet.eval()
    text_encoder.eval()

  # support DistributedDataParallel
  if type(text_encoder) == DDP:
      text_encoder = text_encoder.module
      unet = unet.module
      network = network.module
```
Doing a type checking should be a better method, and this little fix is need for gradient checkpointing to work.

estaesta:
how do you manage to run using multi gpu on Kaggle? I've updated my sd-script but it is still crashing because of the RAM

Isotr0py:
> how do you manage to run using multi gpu on Kaggle? I've updated my sd-script but it is still crashing because of the RAM

I have set my training notebook public in Kaggle. Maybe it can help you.
https://www.kaggle.com/code/isotr0py/lora-train-multi-gpus

Penguin0321:
I wonder how do you pass the parameter "multi_gpu" in a certain function since I don't see any "multi_gpu" in code

Isotr0py:
`--multi_gpu` is the hardware selection argument from `accelerate launch` instead of the script.
You can find reference from their document: [using-accelerate-launch](https://huggingface.co/docs/accelerate/v0.14.0/en/basic_tutorials/launch#using-accelerate-launch)

PR #173, opened 2023-02-10
Train inpainting model/LoRA/TI

Fannovel16:
Reference: https://github.com/huggingface/diffusers/blob/main/examples/research_projects/dreambooth_inpaint/train_dreambooth_inpaint.py


Fannovel16:
Note that this code wont be able to run, but it may be a good base ig.

PR #189, opened 2023-02-14
Show the moving average loss

shirayu:
This PR replaces the loss with the moving average of loss.
Current calculation of loss for logging resets at the begging of each epoch. 
This causes jumping of loss.

# Images

## Before
Before

## After
After

### DIFF
--- a/train_db.py
+++ b/train_db.py
@@ -206,6 +206,7 @@ def train(args):
   if accelerator.is_main_process:
     accelerator.init_trackers("dreambooth")
 
+  loss_list = []
   for epoch in range(num_train_epochs):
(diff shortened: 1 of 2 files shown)


shirayu:
`loss/epoch` is the same value to the moving average because window size is the number of steps of an epoch.

Surely, we can also watch the raw loss value and smoothed loss value in TensorBoard.

My first intention of this PR was the improvement of CLI loss indication.
I prefer that the loss value displayed on the screen with the progress bar be continuous.

shirayu:
I removed the call of `sum()` at 8aed5125deddef0a73fcb9a84c3bacefb0059a11.

Issue #181, opened 2023-02-11
LearningRate-Free Learning Algorithm


BootsofLagrangian:
Hi, how about D-adaptation? [https://github.com/facebookresearch/dadaptation](url)

This is a kind of algorithm that end-user doesn't need to set specific learning rate.

In short, D-adaptation use boundedness to find proper learning rate.

So, it might be useful to someone who hard to find hyperparameters.

Before I wrote this issue, I implement D-adaptation optimizer(Adam) for LoRA. It works!

A few code need to implementation. But I don't know all about sd-scripts code, there exists hard codings.

Requirement for D-dataptation is only torch>=1.5.1 and pip install dadaptation.

Here are codes.

In train_network.py
`
from torch.optim as optim # using for a raw learning rate scheduler
`
`
import dadaptation
`
and I hard-coded for applying optimizer.
`
optimizer = optimizer_class(trainable_params, lr=args.learning_rate)
`
to
`
optimizer = dadaptation.DAdaptAdam(trainable_params, lr=1.0, decouple=True, weight_decay=1.0)
`
Setting decople=True means that optimizer is AdamW not Adam. and weight_decay is for l2 penalty.

Other argumentation is not for end-user.(maybe)

And trainable_params doesn't need a specific learning rate, so replace
`
trainable_params = network.prepare_optimizer_params(args.text_encoder_lr, args.unet_lr)
`
to
`
trainable_params = network.prepare_optimizer_params(None, None)
`

In sd-scripts, lr_scheduler is a return of get_scheduler_fix function.

But I don't know why using get_scheduler_fix interrupt D-adaptation,

so I override lr_scheduler to LambdaLR. sorry for hard coding again :)

`
lr_scheduler = optim.lr_scheduler.LambdaLR(optimizer=optimizer, lr_lambda=[lambda epoch: 1, lambda epoch: 1], 
                                        last_epoch=-1, 
                                        verbose=False)
`

For monitoring dlr value,

`
logs['lr/d*lr'] = optimizer.param_groups[0]['d']*optimizer.param_groups[0]['lr']
`

might be needed. All things done.

![0c09e55bc847abe5d30bbcf1d03cdb19dd59e7e6760ff0576c66e35d6eae5600](https://user-images.githubusercontent.com/125134079/218263205-687945f8-6d53-480f-8df2-10e02e22ac7e.png)

_This image is d*lr-step graph when I use D-dadaptation._

[...]

PR #209, opened 2023-02-19
Dadaptation optimizer

BootsofLagrangian:
--use_dadaptation_optimizer args added.

with this option, dadaptation is avalilable.

while using dadaptation, it recommend 1-digit float numbers( 0.1 <= x <= 1.0) for learning rate, unet learning rate, textencoder learning rate.


BootsofLagrangian:
Related issue
#181

PR #236, opened 2023-02-26
Add Minimal Implementation of Masked Weight Loss

AI-Casanova:
This is my first ever Pull Request so bare with me, please. 

@cloneofsimo instituted weighted loss here: https://github.com/cloneofsimo/lora/discussions/96 based on a facial recognition mask. 

Following his work somewhat (I am no programmer) I came up with this implementation. It takes the alpha channel of a PNG, and converts it to a weight mask between 0-1. This is multiplied to `noise_pred` and `target` immediately before loss is calculated. 

I was unsure how you'd like to handle passing `args.masked_loss` to `train_util.py` so my PR loads masks from every image and stores them, regardless of whether the files have an alpha layer.

The new argument `--masked_loss` toggles the multiplication of the mask to `noise_pred` and `target`.

I have only tested this on `train_network.py` dreambooth, as I do not have any other datasets prepared, but I do not believe that the method I implemented will impact the other methods.

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -68,6 +68,8 @@ class ImageInfo():
     self.latents_flipped: torch.Tensor = None
     self.latents_npz: str = None
     self.latents_npz_flipped: str = None
+    self.mask: np.ndarray = None
+    self.mask_flipped: np.ndarray = None
 
 
 class BucketManager():
@@ -233,7 +235,6 @@ class BaseDataset(torch.utils.data.Dataset):
     self.dropout_rate: float = 0
     self.dropout_every_n_epochs: int = None
     self.tag_dropout_rate: float = 0
-
     # augmentation
     flip_p = 0.5 if flip_aug else 0.0
     if color_aug:
@@ -257,6 +258,7 @@ class BaseDataset(torch.utils.data.Dataset):
     self.image_data: Dict[str, ImageInfo] = {}
 
     self.replacements = {}
+    self.mask_max_attention = 1
 
   def set_current_epoch(self, epoch):
     self.current_epoch = epoch
@@ -465,11 +467,32 @@ class BaseDataset(torch.utils.data.Dataset):
     random.shuffle(self.buckets_indices)
     self.bucket_manager.shuffle()
 
+  def load_mask(self, path):
+    try:
+      mask_path = path[:path.rindex('.')]+".mask"
+      mask = np.array(Image.open(mask_path))
+      if len(mask.shape) > 2 and mask.max() <= 255:
+        print(mask.shape)
+        return np.array(Image.open(mask_path).convert("L"))
+      elif len(mask.shape) == 2 and mask.max() > 255:
+        print(mask.max())
+        return mask//(((2**16)-1)//255)
+      elif len(mask.shape) == 2 and mask.max() <= 255:
+        return mask
+      else:
+        print(f"{mask_path} has invalid mask format: Defaulting to no mask")
+        return np.ones_like(np.array(Image.open(path).convert("L")))*255
+    except:
+      print(f"{mask_path} not found: Defaulting to no mask")
+    return np.ones_like(np.array(Image.open(path).convert("L")))*255
+
   def load_image(self, image_path):
     image = Image.open(image_path)
-    if not image.mode == "RGB":
-      image = image.convert("RGB")
+    if not image.mode == "RGBA":
+      image = image.convert("RGBA")
     img = np.array(image, np.uint8)
+    #if img[:,-1].mean() == 255:
+    img[...,-1] = self.load_mask(image_path)
     return img
 
   def trim_and_resize_if_required(self, image, reso, resized_size):
@@ -508,16 +531,19 @@ class BaseDataset(torch.utils.data.Dataset):
 
       image = self.load_image(info.absolute_path)
       image = self.trim_and_resize_if_required(image, info.bucket_reso, info.resized_size)
-
+      mask = image[:,:,-1] #grab alpha channel
+      image = image[:,:,:3] #drop alpha channel
       img_tensor = self.image_transforms(image)
       img_tensor = img_tensor.unsqueeze(0).to(device=vae.device, dtype=vae.dtype)
       info.latents = vae.encode(img_tensor).latent_dist.sample().squeeze(0).to("cpu")
+      info.mask = mask/(255/self.mask_max_attention)
 
       if self.flip_aug:
         image = image[:, ::-1].copy()     # cannot convert to Tensor without copy
         img_tensor = self.image_transforms(image)
         img_tensor = img_tensor.unsqueeze(0).to(device=vae.device, dtype=vae.dtype)
         info.latents_flipped = vae.encode(img_tensor).latent_dist.sample().squeeze(0).to("cpu")
+        info.mask_flipped = mask[::-1]/(255/self.mask_max_attention)
 
   def get_image_size(self, image_path):
     image = Image.open(image_path)
@@ -606,6 +632,7 @@ class BaseDataset(torch.utils.data.Dataset):
     input_ids_list = []
     latents_list = []
     images = []
+    masks = []
 
     for image_key in bucket[image_index:image_index + bucket_batch_size]:
       image_info = self.image_data[image_key]
@@ -613,7 +640,9 @@ class BaseDataset(torch.utils.data.Dataset):
 
       # image/latentsを処理する
       if image_info.latents is not None:
-        latents = image_info.latents if not self.flip_aug or random.random() < .5 else image_info.latents_flipped
+        rand_flip = random.random()
+        latents = image_info.latents if not self.flip_aug or rand_flip < .5 else image_info.latents_flipped
+        mask = image_info.mask if not self.flip_aug or rand_flip < .5 else image_info.mask_flipped
         image = None
       elif image_info.latents_npz is not None:
         latents = self.load_latents_from_npz(image_info, self.flip_aug and random.random() >= .5)
@@ -622,6 +651,8 @@ class BaseDataset(torch.utils.data.Dataset):
       else:
         # 画像を読み込み、必要ならcropする
         img, face_cx, face_cy, face_w, face_h = self.load_image_with_face_info(image_info.absolute_path)
+        mask = img[:,:,-1]/(255/self.mask_max_attention) #grab alpha channel
+        img = img[:,:,:3] #drop alpha channel
         im_h, im_w = img.shape[0:2]
 
         if self.enable_bucket:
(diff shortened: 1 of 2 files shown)


AI-Casanova:
Quick and dirty example of `--masked_loss` vs normal loss. All hyper parameters were the same between training runs, including a fixed seed.

AI generated dataset and manual masks
Dataset

First Row Unmasked
xyz_grid-0141-RPGv4
Second Row Masked

bmaltais:
How is the mark used? Is it a seperate file? Pure whire mean full, black mean don't look at and 50% grey is look at it but not as much as the face?

Could you share the test dataset for testing, studying?

This is an interesting concept. More work but I think the results are well worth it.

AI-Casanova:
Here is the dataset I showed above
https://drive.google.com/drive/folders/1-3wc_zbzJiQ3NkshIL0M3fMtNf6DBykx

The masks are saved as the alpha channel of the PNGs, which as long as you don't pre-multiply saves the RGB channels intact. And prevents yet another instance of find the match like exists with captions and .npz

I made these masks in Gimp, but I've also experimented with depth maps on my private dataset.

You are correct, Full white areas experience normal loss, and black are completely removed from loss.

The gray in these masks is 75% I believe, will run again in the next day with 50% and 25% minimums. 

Feel free to check my math, and some things like the *8 in the reshape are straight from @cloneofsimo, and I don't know exactly how that operates.

bmaltais:
Preliminary results of a crude test loo impressive. Here is a graph showing the loss average of the non masked training vs the masked training:

image

Masked training is the one in orange. The non masked training is not yet completed... but the samples from the masked training are pretty good.

bmaltais:
And here is a quick comparison:

With masked face: 

grid-0005

Without masked face:

grid-0006

Same seed, same prompt. This is shocking.

bmaltais:
And one last comp before going to bed.

Masked:

grid-0009

Not Masked:

grid-0010

AI-Casanova:
> Hairs appear more defined on the non masked...

This is part of the reason I ran a gaussian blur on my alpha layers (though perhaps bloom might be more effective to only push higher values out, not meld evenly like blur)

My thought is that allowing some training at the interface between subject and background might help with those edge details.

AI-Casanova:
> the masking area are showing
@bmaltais 
Double check that your alpha channel was not pre-multiplied with the color channels. The first three layers should be exactly the same as the original image. 

If that was at fault, it makes the case for side loading the masks, to prevent people from doing the same. Side loading would also allow the use of 16bit grayscale depth masks natively, ie the outputs from https://github.com/thygate/stable-diffusion-webui-depthmap-script

AI-Casanova:
The other test that could be run is whether or not EXTRA attention is beneficial. 

DreamArtist used a non linear scale from 0-5 instead of the linear scale here of 0-1

https://github.com/7eu7d7/DreamArtist-sd-webui-extension#attention-mask

To test a linear scale from 0-2, line 521 could be changed to `info.mask = mask/128` and line 528 to `info.mask_flipped = mask[::-1]/128`

bmaltais:
I am testing creating the PNG using the depth map extension. Let see how this will compare with the result above... 1st observation, loss for the depth map mask is lower than with the manual mask... normal since the depth map has much more depth variation than the manual one which is pretty much an off, half and full mask.

Manual vs depth map mask learning:

image

But the result is not great...

grid-0014

Almost like it is way more rigid... perhaps the depth map produced PNG are not quite right...

AI-Casanova:
I just switched to sideloaded masks to avoid confusion with the alpha layers.
`train_util.py` now searches for `.mask` files in the same directory, pretty much the same logic as `.txt` caption files.
These files can be any of the filetypes accepted by the trainer, and has been tested on 16bit PNG from depths maps as well.

This is not backward with alpha layer masks.

Also added a `self.mask_max_attention` stub for scaling beyond 0-1
Setting it to 1.25 would make losses in white areas 25% stronger.

AI-Casanova:
@kohya-ss I enjoy the projects and sharing what I come up with for my own training, and I do like that you keep a very stable codebase, so I am by no means offended if I am outside of your intended scope. 

May I ask why you pass all of the `args` to 'FineTuningDataset` and `DreamBoothDataset` as positional arguments instead of passing `args` itself? I just noticed it when trying to send extra arguments through for the dataset handling I was doing, and had to add more positionals.

AI-Casanova:
Oops, didn't mean to close it, fat thumbs, but I will leave the fork as is, so it's available for review.

bmaltais:
@AI-Casanova I tried the new .mask feature and it worked very well. I was not able to get anywhere with the photoshop alpha channed yesterday for a model byt using depth map generated images named .mask has produced great results. Too bad this feature does not get integrated at the moment. Hopefully once the large change is implemented @kohya-ss might consider integrating it.

bmaltais:
In a way using gepth map for learning is like a controlnet for learning... It guide SD to learn what is important, maximising the learning potential of each images.

AI-Casanova:
>It guide SD to learn what is important, maximising the learning potential of each images.

That was my hope, to reduce reliance on captioning backgrounds. It is my (perhaps mistaken) feeling that the longer the caption, the more likely it is for the identity I'm trying to train to bleed out into the rest of the caption.

bmaltais:
> > It guide SD to learn what is important, maximising the learning potential of each images.
> 
> That was my hope, to reduce reliance on captioning backgrounds. It is my (perhaps mistaken) feeling that the longer the caption, the more likely it is for the identity I'm trying to train to bleed out into the rest of the caption.

I think you are right. I wonder if convnet depth_leres could be dynamically used at dataset image load time to generate a depth mash for training... using convnet as a tool to focus training vs manually creating the masks beforahand. I guess the convnet masks could be saved for reuse in other run...

flesnuk:
@AI-Casanova will this PR still be worked on? I don't know if this feature was added in other commit or not, or how to use it.
I saw that the original LoRA repo has something similar to this implemented but still doesn't clarify how to use it:
https://github.com/cloneofsimo/lora/discussions/96
Can you change the PR status to Open if is intended to be merged sometime, so there isn't any confusion?

AI-Casanova:
@flesnuk this PR needs to be reworked from scratch because of the major changes that have happened in the mean time, but it something that I should probably work on, as it showed some promise.

AI-Casanova:
@TingTingin Honestly it's quite fallen off my radar, in lieu of other projects. 

@kohya-ss Would you be interested if I take another crack at this? I'd likely put the mask loading as a TOML config option to select a folder and then search for matching names.

It occurs to me that it might be quite possible to use mediapipe to auto-mask faces, with face set at one value and everything else at another. This would be potentially useful both to train faces, and to ignore them when training a costume etc.

AI-Casanova:
Good idea. For every folder, if masked loss is enabled, assume folder/mask/* unless specified otherwise.

AI-Casanova:
I noticed there was a ControlNet branch, but never popped in, that sounds perfect.

I understand the point about auto-masking, I'll take some time to think about what that would best look like. 

As to the extra input and data loader changes, it's somewhat tangential to another proposal I had for you. With the inclusion of sliced VAE, and thus the ability to load larger images, it might be possible now to add data augmentations like random crop, random zoom, affine transforms etc to cached latents by loading them larger, and slicing in the data loader. This would have to be replicated for potential masks, but that isn't an insurmountable problem.

AI-Casanova:
I understand your concern about cropping latents, but a quick empirical test doesn't seem to show any worrying distortions at the cropped edges. 
Center crop:
`latents = latents[:,:2,rc:latents.shape[-2]-2),2:latents.shape[-1]-2]`
Corner Crop:
`latents = latents[:,:,:latents.shape[-2]-4,:latents.shape[-1]-4]`
Random crop: 
`rc= random.randint(0,4)
latents = latents[:,:,rc:latents.shape[-2]-(4-rc),rc:latents.shape[-1]-(4-rc)]` 
xyz_grid-0162-epicrealism_v1

Or are you worried about a tiled VAE encoding adding the seams to the base latent?

AI-Casanova:
Added one more crop method, this time loading though the VAE at 1.5x size (ie 768 and 1152 instead of 512 and 768) and then cropping to the original size.

I have a hunch that the issues with the VAE are on the decoder side, and not the encoder, but I'm not sure how to prove that beside that I'm empirically not seeing any distortions except for the fact that I've cropped the face in training, and thus in my results.
(slight color shift towards natural skin tone in the last column is probably max_norm which I forgot to disable)
image

recris:
@AI-Casanova I've submitted a slightly reworked version of your PR, it should work with current main branch

I've changed the mask loading logic to look for masks in a `mask` sub-folder, as previously suggested.

I've also tweaked the MSE loss calculation, mask values are now normalized using the mask mean value. I believe this makes the magnitude of the calculated loss to be less sensitive to the amount of non-black pixels in the mask - my testing agrees with this.

AI-Casanova:
@recris that's amazing! I'll pull as soon as life lets me and give it a try.

recris:
The relative attention of each pixel is still the same, grayscale value are mapped to the [0, 1] weight range.

As an example, lets say you have training set with 2 black and white masks, one has a very large white area (A) and the other has a small white area (B). Without re-scaling, the computed MSE loss will be proportional to the amount of white pixels in the overall image, meaning the calculated loss in A will be (statistically) greater than the loss in B, and this may skew the training. By re-scaling we eliminate this proportionality effect.

AI-Casanova:
I worry a bit about that rescaling. In my use case, I'd mask faces at 1, bodies at .5, and backgrounds at .25 perhaps.

I would think that any face should be backpropagated wrt the same per pixel loss, disregarding the reported loss, as I don't find loss graphs to be valuable.

recris:
I've done a few runs with re-scaling and haven't found negative effects yet. Keep in mind that the relative pixel weights stay the same since everything is scaled by the same factor.

The biggest improvement I found with this whole change is that certain details from the training data, like backgrounds, stopped leaking into the generated images. This was especially evident when the model is over-fitted. Previously I had to be smarter when choosing and labeling the images but now I can get the same quality with less effort put into the training set.

AI-Casanova:
Out of curiosity, are you fully zeroing your backgrounds, or leaving a bit for context?

recris:
My backgrounds are fully black, but I am still providing background descriptions in the captions (not sure if it matters)

recris:
Because creating human subject masks is very tedious, I've came up with a small script to automate this process: https://github.com/recris/subject-masker

It uses a combination of face detection, parsing and recognition models, plus instance segmentation model to generate subject masks. It supports providing distinct weight values for face, hair, body and background.

Overall the generated masks are pretty good, with the occasional mask that needs some additional cleaning in GIMP.

AI-Casanova:
All the things I was going to do """someday""" this boss already did.

PR #240, opened 2023-02-28
Partial library refactor

Ristellise:
Hi! In the process of looking at the library code so that I can add a potentially new method for training, I decided to refactor the model_util.py into smaller python files for ease of management.

The changes made is as follows:

  - Split up model_util.py into smaller files
  - Use `black` for code formatting.
  - 2 spaces to 4 spaces conversion.

Although it is longer in total overall length, I found the code to be more presentable in smaller chunks rather than having to scroll through the model_util.py previously which contained way too many functions and classes. 

Since this is a huge change, i'm marking this PR as a draft. Feel free to close the PR if it isn't the intended direction to take.


Ristellise:
I noticed that I included an extra dataset file for other purposes, I'll remove once I have access to a PC.

Ristellise:
Sorry to spy but I saw the comments on twitter relating to the PR. I'll close it as I believe it isn't the right direction.

Ristellise:
> Thank you for your consideration. I also understand that the code is messy and confusing. However, some of them are intended. I would rather spend time adding features than refactoring now, but in the future I will improve one by one.
> 
> `black` seems to be better than `autopep8`, I will consider switching to it in near future. Thank you for letting me know!

Hi! Thank you for the reply. I'll likely contribute a feature as well in the near future. The refactoring in part helped me understand how the code was structured, so it wasn't all for nothing.

PR #241, opened 2023-02-28
Load training arguments from .yaml, and other small changes

Linaqruf:
Hi, I want to propose some small changes about 2 things:

1. Some image scrapers, like [gallery-dl](https://github.com/mikf/gallery-dl), write tags or metadata using the `.jpg.txt` format instead of `.txt`. As a result, sometimes `merge_dd_tags_to_metadata.py` and `merge_captions_to_metadata.py` can't read that format. To address this, I have added a condition to read `.jpg.txt` if no `.txt` file is found.

2. It is becoming difficult to maintain training arguments because there are too many now. I suggest finding a way to pass training arguments to accelerate instead of writing them all down. My suggestion is to use `.yaml` as they are flexible and easy to maintain, compared to `.json`. 

Example: [lora_config.yaml](https://gist.github.com/Linaqruf/cab36a844293e311cf1ea1d54478f080)

i've only tested this on train_network.py
image

I know you can do better than mine for the implementation. Thanks!

### DIFF
--- a/fine_tune.py
+++ b/fine_tune.py
@@ -5,6 +5,7 @@ import argparse
 import gc
 import math
 import os
+import toml 
 
 from tqdm import tqdm
 import torch
@@ -362,4 +363,24 @@ if __name__ == '__main__':
   parser.add_argument("--train_text_encoder", action="store_true", help="train text encoder / text encoderも学習する")
 
   args = parser.parse_args()
+
+  if args.config_file:
+      config_path = args.config_file + ".toml" if not args.config_file.endswith(".toml") else args.config_file
+      if os.path.exists(config_path):
+          print(f"Loading settings from {config_path}...")
+          with open(config_path, "r") as f:
+              config_dict = toml.load(f)
+
+          ignore_nesting_dict = {}
+          for section_name, section_dict in config_dict.items():
+              for key, value in section_dict.items():
+                  ignore_nesting_dict[key] = value
+
+          config_args = argparse.Namespace(**ignore_nesting_dict)
+          args = parser.parse_args(namespace=config_args)
+          args.config_file = args.config_file.split(".")[0]
+          print(args.config_file)
+      else:
+          print(f"{config_path} not found.")
+          
   train(args)
--- a/finetune/merge_captions_to_metadata.py
+++ b/finetune/merge_captions_to_metadata.py
@@ -4,7 +4,7 @@ from pathlib import Path
 from typing import List
 from tqdm import tqdm
 import library.train_util as train_util
-
+import os
 
 def main(args):
   assert not args.recursive or (args.recursive and args.full_path), "recursive requires full_path / recursiveはfull_pathと同時に指定してください"
@@ -28,6 +28,9 @@ def main(args):
   for image_path in tqdm(image_paths):
     caption_path = image_path.with_suffix(args.caption_extension)
     caption = caption_path.read_text(encoding='utf-8').strip()
+
+    if not os.path.exists(caption_path):
+      caption_path = os.path.join(image_path, args.caption_extension)
 
     image_key = str(image_path) if args.full_path else image_path.stem
     if image_key not in metadata:
--- a/finetune/merge_dd_tags_to_metadata.py
+++ b/finetune/merge_dd_tags_to_metadata.py
@@ -4,7 +4,7 @@ from pathlib import Path
 from typing import List
 from tqdm import tqdm
 import library.train_util as train_util
-
+import os
 
 def main(args):
   assert not args.recursive or (args.recursive and args.full_path), "recursive requires full_path / recursiveはfull_pathと同時に指定してください"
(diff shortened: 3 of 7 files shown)


Linaqruf:
Hi, I don't know much about `.toml`, but I think `.yaml` is easier to use than `.toml`. `.yaml` is more widely used and simpler for end-users because it doesn't require strong typing like `.toml`. e.g. `data = "value"`, but instead uses `data: value`. One weakness of `.yaml` probably its indentation can be tricky.

But I think `.toml` is also a good choice for Python users out there because it uses similar syntax to Python when declaring variables and the data type, while `.yaml` is more similar to defining dict. Sometimes, I make mistakes with `.yaml` because it's too simple such as using `=` instead of `:`, which breaks my muscle memory.

green-s:
YAML is famously [over-complicated](https://github.com/cblp/yaml-sucks), but it does seem to be pretty standard for this kind of config in ML/Python. Coming from Rust where TOML is standard, I vote for that. It's not perfect, but it's pretty much just a stricter, extended .ini format, and since most ML configs tend to be flat/unstructured anyway, the (slightly controversial) nesting and array syntaxes are unlikely to be an issue. Either would work fine though.

fur0ut0:
I have some suggestions:

1. How about supporting TOML and YAML at the same time? However, it will be a concern that users may confuse with them.
1. How about unifying with existing dataset config feature? It may be straightforward to have only single config feature, as long as their options do not conflict with each other. However, there is another concern about how we organize priority of options (current priority: TOML dataset config > CLI options > YAML training config).

Linaqruf:
Hi, sure. I think I can do it, but do you think this PR still relevant? as it only a simple wrapper to wrap training args and pass them to training like .ps1 or .sh, because I've read new config_util and I think it has great potential and we can explore it more.

fur0ut0:
> It seems to me that the two configs should be kept separate rather than unified.
> 
> 1. we may want to share a dataset across multiple trainings, even for LoRA, DreamBooth or fine tuning.
> 2. we may want to train multiple datasets with common training config, for example train multiple characters one by one.
> 
> The training config might assume that the dataset config is used. This means that the training config doesn't have dataset configs. The CLI options always take priority.
> 
> (However, if the training config is a simple wrapper for CLI options, it can have dataset configs.)

I see.

I think it can point this by unifying the config schema and supporting splitted config files via CLI. However, this seems off topic of this PR. I will consider of this feature in future.

> I think I may merge this PR as it is. Whether to use it in `.toml` format is a difficult. But, because as Linaqruf says, it is a simple wrapper, the code would not be too complicated if both `.yaml` and `.toml` were supported in future (except for defining the schema).

Yes, adding config file format does not require hard work. Whether to adopt it depends on designing preference. I personally think it is better to provide several alternatives for config file format, but I understand decision not to provide many formats for clean design and clear usage.

Linaqruf:
Thanks, I'll do my best, but first I need to make sure latest commit running without error because I can't run it in colab notebook since the last update. 😅

bmaltais:
If this move forward it might be interesting to unify the config file with the one for the GUI so that the output of the GUI config could also be used with the CLI and vice versa... Will keep an eye on this. I personally use json to store the GUI config for sd-scripts as it is easy to deal with using python.

Linaqruf:
sdaddadh accidentally discarded commit

Linaqruf:
Hi, sorry for the long wait, I need to make sure the code is working because I don't really know much about `.toml` and the nesting stuffs. 

This is how training config should be passed from `.toml`. Nesting here is just a 'glorified' comments because I don't know how should I use it. So the code ignore it and just read the `key` and `value` of the config.

This is training config I use to train with this code : [config](https://huggingface.co/Linaqruf/locon-library/blob/main/config/config_file.toml)


```toml
[model_arguments]
v2 = false
v_parameterization = false
pretrained_model_name_or_path = "/content/pretrained_model/Anything-v3-1.safetensors"

[additional_network_arguments]
no_metadata = false
unet_lr = 0.0001
text_encoder_lr = 5e-5
network_module = "locon.locon_kohya"
network_dim = 64
network_alpha = 32
network_args = [ "conv_dim=64", "conv_alpha=32",]
network_train_unet_only = false
network_train_text_encoder_only = false

[optimizer_arguments]
optimizer_type = "AdamW8bit"
learning_rate = 0.0001
max_grad_norm = 1.0
optimizer_args = [ "weight_decay=0.6",]
lr_scheduler = "constant"
lr_warmup_steps = 0

[dataset_arguments]
cache_latents = true
debug_dataset = false

[training_arguments]
output_dir = "/content/LoRA/output"
output_name = "hitokomoru_locon"
save_precision = "fp16"
save_every_n_epochs = 1
train_batch_size = 6
max_token_length = 225
mem_eff_attn = false
xformers = true
max_train_epochs = 20
max_data_loader_n_workers = 8
persistent_data_loader_workers = true
gradient_checkpointing = false
gradient_accumulation_steps = 1
mixed_precision = "fp16"
clip_skip = 2
logging_dir = "/content/LoRA/logs"
log_prefix = "hitokomoru_locon"
noise_offset = 0.1
lowram = true

[sample_prompt_arguments]
sample_every_n_epochs = 1
sample_sampler = "ddim"

[dreambooth_arguments]
prior_loss_weight = 1.0

[saving_arguments]
save_model_as = "safetensors"
```

PR #243, opened 2023-02-28
Enable ability to resize lora dim based off sv ratios

mgz-dev:
Had the idea after looking at KohakuBlueleaf's implementation using [SV threshold](https://github.com/KohakuBlueleaf/LoCon/commit/76d0d282d69947d6310f616ea6eff10ac7dac0e0) to dynamically determine layer dimension while extracting dreambooth models to lora.

This uses maximum SV ratio instead of a threshold (all singular value ranks below the ratio are dropped. e.g. if ratio is 10 and the largest singular value is 5, all singular values below 0.5 are removed and the dim is dynamically calculated per layer).

LoRA resized this way currently do not work with [sd-webui-additional-networks](https://github.com/kohya-ss/sd-webui-additional-networks) since it will give size mismatch warnings but does funciton with built-in lora support in webui.

Same logic can be applied to extract_lora_from_models.

### DIFF
--- a/networks/resize_lora.py
+++ b/networks/resize_lora.py
@@ -38,10 +38,11 @@ def save_to_file(file_name, model, state_dict, dtype, metadata):
     torch.save(model, file_name)
 
 
-def resize_lora_model(lora_sd, new_rank, save_dtype, device, verbose):
+def resize_lora_model(lora_sd, new_rank, save_dtype, device, sv_ratio, verbose):
   network_alpha = None
   network_dim = None
   verbose_str = "\n"
+  ratio_flag = False
 
   CLAMP_QUANTILE = 0.99
 
@@ -57,9 +58,12 @@ def resize_lora_model(lora_sd, new_rank, save_dtype, device, verbose):
       network_alpha = network_dim
 
   scale = network_alpha/network_dim
-  new_alpha = float(scale*new_rank)  # calculate new alpha from scale
-
-  print(f"old dimension: {network_dim}, old alpha: {network_alpha}, new alpha: {new_alpha}")
+  if not sv_ratio:
+    new_alpha = float(scale*new_rank)  # calculate new alpha from scale
+    print(f"old dimension: {network_dim}, old alpha: {network_alpha}, new dim: {new_rank}, new alpha: {new_alpha}")
+  else:
+    print(f"Dynamically determining new alphas and dims based off sv ratio: {sv_ratio}")
+    ratio_flag = True
 
   lora_down_weight = None
   lora_up_weight = None
@@ -97,11 +101,24 @@ def resize_lora_model(lora_sd, new_rank, save_dtype, device, verbose):
 
         U, S, Vh = torch.linalg.svd(full_weight_matrix)
 
+        if ratio_flag:
+          # Calculate new dim and alpha for dynamic sizing
+          max_sv = S[0]
+          min_sv = max_sv/sv_ratio
+          new_rank = torch.sum(S > min_sv).item()
+          new_rank = max(new_rank, 1)
+          new_alpha = float(scale*new_rank)
+
         if verbose:
           s_sum = torch.sum(torch.abs(S))
           s_rank = torch.sum(torch.abs(S[:new_rank]))
-          verbose_str+=f"{block_down_name:76} | "
-          verbose_str+=f"sum(S) retained: {(s_rank)/s_sum:.1%}, max(S) ratio: {S[0]/S[new_rank]:0.1f}\n"
+          verbose_str+=f"{block_down_name:75} | "
+          verbose_str+=f"sum(S) retained: {(s_rank)/s_sum:.1%}, max(S) ratio: {S[0]/S[new_rank]:0.1f}"
+
+        if verbose and ratio_flag:
+          verbose_str+=f", dynamic| dim: {new_rank}, alpha: {new_alpha}\n"
+        else:
+          verbose_str+=f"\n"
 
         U = U[:, :new_rank]
         S = S[:new_rank]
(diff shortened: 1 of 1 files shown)


mgz-dev:
Sure. I grabbed the Arcane, Amber, and Makima LoRA from CivitAI since they are ranked as popular and processed them at sv_ratio of 5 (you can likely be more aggressive if you want). Listing file size reduction since dim is dynamic in new LoRAs.

- Arcane (295mb to 55mb)
1851130503 cfg7 steps20 1950

- Amber (144mb to 23mb)
1851130503 cfg7 steps20 1619 (Custom)

- Makima (148mb to 49mb)
1851130503 cfg7 steps20 1336

thojmr:
Was just about to mention Conv2d, nice!  Looking forward to testing extraction after Conv2d is implemented.  Should be way more accurate.

ashen-sensored:
K percentile(cumsum), then frobenius norm as precision error check may be a better way of doing it.
We can use precision error as controllable parameter input and do a sweep.

> Had the idea after looking at KohakuBlueleaf's implementation using [SV threshold](https://github.com/KohakuBlueleaf/LoCon/commit/76d0d282d69947d6310f616ea6eff10ac7dac0e0) to dynamically determine layer dimension while extracting dreambooth models to lora.
> 
> This uses maximum SV ratio instead of a threshold (all singular value ranks below the ratio are dropped. e.g. if ratio is 10 and the largest singular value is 5, all singular values below 0.5 are removed and the dim is dynamically calculated per layer).
> 
> LoRA resized this way currently do not work with [sd-webui-additional-networks](https://github.com/kohya-ss/sd-webui-additional-networks) since it will give size mismatch warnings but does function with built-in lora support in webui.
> 
> Same logic can be applied to extract_lora_from_models.

mgz-dev:
Thanks for the suggestion. It is easy to implement resizing based off any of these methods (cumulative sum, Frobenius norm, or sv ratio) so I went ahead and added them as options. I think they all have their strengths and weaknesses though and don't necessarily agree that cumulative sum or frobenius norm is strictly better.

- Singular value ratio should function well if the goal is to denoise lower ranking data. It likely will also perform best at extremely high levels of compression (though this should be checked on a case-by-case basis)
- Cumulative sum is similar to SV ratio, though the meaning of this value does not translate as well between different starting dims. It is important for people to understand that a "lower" cumulative sum is not necessarily bad. I originally avoided this because I was afraid people would misunderstand 0.6 as "only 60% as good as the original".
- Frobenius norm is likely the most intuitive number for most users. This will try to represent the original matrix as accurately as possible. This is a good choice for trying to recreate with the highest level of accuracy at the tradeoff of compression. This is not a good option for users is trying to salvage an overfit LoRA. It will have lower compression on average but the least variance per layer.

bmaltais:
I discovered an interesting thing... clamping is sort of degrading the output. To my surprise changing the `CLAMP_QUANTILE = 1 # 0.99` to 1 improves the results drastically... here is an example:

Original rank 256 model:

grid-0205

Resized with sv 5 clamp 0.99:

grid-0207


Resized with sv 5 clamp 1:

grid-0206

File size reduction for both was the same... from 295MB to 19.4 MB... but output of clamp 1 is much closer to original... perhaps the clamp value need to be exposed as a parameter?

mgz-dev:
@bmaltais good observation. I actually think clamping may not be required at all since SVD should always return norm 1 vectors within U and Vh but wasn't sure if there were some unique exceptions which caused problems to warrant its inclusion in the original dreambooth extraction code. It may be best to just remove the quantile/clamp processing entirely.

@TingTingin as a brief explanation, it is somewhat similar to the theory behind PCA. 
pca example
In our case the assumption is that dimensions with lower variance (you can think of smaller singular values as a proxy for less information) are less likely to contain useful data and more likely to contain noise from the training set which does not reflect what was actually desired to be learned during the training process. You can also think of it similarly to signal denoising.

As far as valid ranges, 
- sv_ratio should be greater than 1 (I'd recommend staying 2 or greater at a minimum)
- sv_cumulative is somewhat dependent on the original rank of a matrix but numbers are restricted to be between (0,1)
- sv_fro is also restricted to be between (0,1) but in practice should be much higher than sv_cumulative

You can use --verbose flag and compare numbers. I'd recommend doing some level of iteration to see what gives the best results for whichever LoRA you try to reduce since there is some variation in a case by case basis.

bmaltais:
Based on testing one is better to create the highest rank possible to get the most out of the model then resize it to a much smaller size with minimal loss. This provide way better results vs trying to train a small rank LoRA.

mgz-dev:
Apologies for so many additional changes. I was working to make the script modular to be able to recycle code to handle dreambooth extraction but it looks like Kohaku-Blueleaf pushed an update their repo which overlaps in that task. I'll table dreambooth extraction for now unless there is some reason to bring it back up.

Latest version should also be able to handle the conv2d layers. 

**important** `--new_rank` will now also limit the max rank of any individual lora layer during dynamic resizing. I added this because I noticed some layers could be disproportionately expensive space-wise. This will allow people to see if restricting those layers further result in quality loss (users can test on a model-by-model basis if they're trying to maximize space savings).

@bmaltais I have not done enough testing to say if 2-step LoRA is the best method, but I also have seen examples of trimming LoRA even improving quality. As a caveat though, I would think that trimming too many ranks runs the risk of losing important information in the model. There is probably some range that is best to stay within for resizing.

bmaltais:
@mgz-dev I will refer to your repo for updated code until you are ready to update the PR as to keep up with the improvements. Nice work. I like that it now support LoCon also.

I have a question for you... if one use SD1.5 as the model... what would you say it the highest rank one should use when creatien a LoRA... since the model use 768 vectors... would a ran 768 be the limit after whick there is no return?

I am asking because I think LoRA can be used as nice "frozen" model training. Once the LoRA has been created at full model potential resolution one simply use your tool to reduce it down to what really constitute the model. In my test this is providing the best LoRA.... but I wonder what the practical rank should be for a model...

woct0rdho:
Hi @kohya-ss , so you've removed the clamp in `resize_lora.py`. Would you also remove it in `extract_lora_from_models.py` and `svd_merge_lora.py`?

kohya-ss:
> Hi @kohya-ss , so you've removed the clamp in `resize_lora.py`. Would you also remove it in `extract_lora_from_models.py` and `svd_merge_lora.py`?

The actual work was done by mgz-dev, but clamp seems to have been removed. I think it may be an idea to make clamp optional in those scripts. In other words, allow omission (None) and do not clamp if not specified.

FurkanGozukara:
> > Hi @kohya-ss , so you've removed the clamp in `resize_lora.py`. Would you also remove it in `extract_lora_from_models.py` and `svd_merge_lora.py`?
> 
> The actual work was done by mgz-dev, but clamp seems to have been removed. I think it may be an idea to make clamp optional in those scripts. In other words, allow omission (None) and do not clamp if not specified.

is Clamp 1 same as None ?

woct0rdho:
@FurkanGozukara Yes it should be the same

PR #261, opened 2023-03-05
metadata |= to metadata.update

camenduru:
thanks for the 3.8 support @kohya-ss ❤


cerarslan:
No. I want to say big thanks both of you!

PR #271, opened 2023-03-09
Add lr_scheduler_type and lr_scheduler_args

Isotr0py:
- Add argument `lr_scheduler_type` and `lr_scheduler_args` to use lr_scheduler from another library

For example, to use `torch.optim.lr_scheduler.CosineAnnealingWarmRestarts` with t0=10 as lr_scheduler, we can just add 
`--lr_scheduler_type "CosineAnnealingWarmRestarts" --lr_scheduler_args "T_0=10"` or
`--lr_scheduler_type "torch.optim.lr_scheduler.CosineAnnealingWarmRestarts" --lr_scheduler_args "T_0=10"`

I haven't fully tested it yet since some schedulers will use more complex inputs. 
I will open it as soon as I finish my test.

### DIFF
--- a/fine_tune.py
+++ b/fine_tune.py
@@ -178,9 +178,7 @@ def train(args):
     print(f"override steps. steps for {args.max_train_epochs} epochs is / 指定エポックまでのステップ数: {args.max_train_steps}")
 
   # lr schedulerを用意する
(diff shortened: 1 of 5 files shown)


DarkAlchy:
Where do we add this for sd-scipts (not in Kohya_ss)?

PR #290, opened 2023-03-13
fix exception on training model in diffusers format

orenwang:
`sd-scripts` supports training models in `diffusers` format, but methods `model_hash()` and `calculate_sha256()` are throwing `IsADirectoryError`s on model that is not a directory.

This PR catches these exceptions and return 'IsADirectory' as hash instead.


orenwang:
Sounds good. You are way too kind!

PR #327, opened 2023-03-25
P+: Extended Textual Conditioning in Text-to-Image Generation

jakaline-dev:
Implemented from https://prompt-plus.github.io/
image

This method is training TI for each cross attention layer.

Comparable to LoRA? We will have to see.

### DIFF
--- a/XTI_hijack.py
+++ b/XTI_hijack.py
@@ -1,0 +1,209 @@
+import torch
+from typing import Union, List, Optional, Dict, Any, Tuple
+from diffusers.models.unet_2d_condition import UNet2DConditionOutput
+
+def unet_forward_XTI(self,
+        sample: torch.FloatTensor,
+        timestep: Union[torch.Tensor, float, int],
+        encoder_hidden_states: torch.Tensor,
+        class_labels: Optional[torch.Tensor] = None,
+        return_dict: bool = True,
+    ) -> Union[UNet2DConditionOutput, Tuple]:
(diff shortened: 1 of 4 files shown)


jakaline-dev:
> how many images how long was training what gpu did you use

~150 images, ~50 minutes on 4080
Memory usage is similar to plain TI training

jakaline-dev:
It's too complicated to fix for the sampling during training stages, so going dirty and disabled sampling for now. If it's not urgent, it can be fixed in the future.

jakaline-dev:
제목 없음
```
from safetensors import safe_open
from safetensors.torch import save_file

subsets = ["MID", "IN08", "OUT03", "IN07", "OUT04", "OUT05"]

for i in range(6):
    f1 = {}
    with safe_open("emu_prsk.safetensors", framework="pt", device="cpu") as f:
        for key in f.keys():
            f1[key] = f.get_tensor(key)

    f2 = {}
    with safe_open("hayate.safetensors", framework="pt", device="cpu") as f:
        for key in f.keys():
            f2[key] = f.get_tensor(key)

    for key in subsets[0:i+1]:
        f1[key] = f2[key]    
    save_file(f1, f"emu_hayate_subset_{i+1}.safetensors")
```
Meanwhile, here is some code for mixing layers (just as the paper did)

kgonia:
@jakaline-dev can prompt+ be used for standard finetuning?

PR #332, opened 2023-03-27
Reduce peak RAM usage

guaneec:
These are the changes I had to make to have LoRA training work in the Kaggle T4 x 2 environment.

What this PR does:
* Sequentially loads the SD model for each GPU so the peak usage doesn't grow linearly with number of GPUs
* Directly load state dicts into the correct device

Haven't tested if it breaks other stuff.


feffy380:
This seems to cause out of memory errors on gpus with 8GB VRAM (3060 ti):
image
(Screenshot not mine)

Issue #312, opened 2023-03-22
[Request] add caption words attention/emphasys for training


swfsql:
Is it possible to implement emphasys changes on words from the captions during training?  
This is the explanation of this feature in [auto](https://github.com/AUTOMATIC1111/stable-diffusion-webui/wiki/Features#attentionemphasis).

Example: "an astronaut riding a horse on the (moon:1.5)"
Thanks in advance!

Edit: I'm seeing that there is a regex related to this in [here](https://github.com/kohya-ss/sd-scripts/blob/main/library/lpw_stable_diffusion.py), but would someone know if this is applied during training? Thanks!

AI-Casanova:
The code from `lpw_stable_diffusion.py` doesn't look too hard to implement, and the negative prompts bits can be dropped. 

I'll give it a shot when I have time, and ping @swfsql in the PR if it works to help do some testing for effectiveness.

swfsql:
Thanks, please let me know if I can be of help with the testing.
I'm somewhat new to both python, AI and SD, and new to LoRa and training and so on, but I was trying to train a LoRa on gif images, and I had no luck so far.. and I also only got 6GB of vram.

But my original thought was that I could try separating the gif files in different "duration" concepts, where I would try to correlate the actual gif duration with the attention of the "duration" token(s) according to some formula. For example, saying `(gif duration:0.5)` could mean the gif tagets 0.5s of duration, and saying `(gif duration:1.5)` could mean the gif targets 5s of duration..

I'm pretty sure this would not work, as I only had failures so far, but anyway I couldn't help but try asking if this feature existed XD But coming back to you, I can try coming with other ways to test it in case you implement it, even if it's not related to the case I was thinking originally!

AI-Casanova:
> gif images

Unfortunately the only format that will load are `IMAGE_EXTENSIONS = [".png", ".jpg", ".jpeg", ".webp", ".bmp", ".PNG", ".JPG", ".JPEG", ".WEBP", ".BMP"]`

The way I see this feature working is to emphasize foreground tokens and de emphasize background tokens.

ie

`(SKS woman:1.1), (wearing a black dress:0.9), (sitting in a restaurant:0.8)`

In this way, we can hopefully learn the important features more strongly than the backgrounds, without ignoring the backgrounds completely.

Best practices for captioning will have to come from experimentation, but this will hopefully give us a new tool to experiment with.

AI-Casanova:
[...]

PR #336, opened 2023-03-28
Proof of Concept: Weighted captions

AI-Casanova:
### This PR is for testing the effect of weighted captions, not for imminent merging. 

Implementation of weighted captions requested here https://github.com/kohya-ss/sd-scripts/issues/312 and elsewhere on Discord etc.

With `--weighted_captions` one can now use captions with the standard (token:1.2) syntax, though one must be careful not to put a comma inside the parentheses if using `--caption_tag_dropout` `--shuffle_caption` etc as that would break the closing parens.

The entire embeddings are scaled to their previous mean after individual embeddings are scaled up, so increasing the weight of your foreground (subject) tokens effectively downscales the rest of them. 

I ran 2 quick runs with `--weighted_captions` one with standard unweighted captions to test the tokenization, and another with `(<token> woman:1.2)`

image

Losses track exactly with deterministic training, which is unsurprising to me. 

xyz_grid-0190-RPGv4
Images prompted only with `<token> woman` to proof the effect. Its fairly small, but my dataset has fairly short captions, so there's not a lot of comparative scaling perhaps.

This code was adapted from `lbw_stable_diffusion.py`

This is likely not the best implementation, and only implemented for `train_network.py` for tonight, (it could be handled natively in `train_util.py` for instance) but this disturbs the rest of the repo as little as possible.

### DIFF
--- a/library/custom_train_functions.py
+++ b/library/custom_train_functions.py
@@ -1,5 +1,8 @@
 import torch
 import argparse
+import re
+from typing import List, Optional, Union
+
 
 def apply_snr_weight(loss, timesteps, noise_scheduler, gamma): 
   alphas_cumprod = noise_scheduler.alphas_cumprod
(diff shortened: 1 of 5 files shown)


library/custom_train_functions.py
AI-Casanova, line 1: Oh, I think I can ditch a few of those imports, will confirm tomorrow.

library/custom_train_functions.py
```diff
@@ -16,3 +18,316 @@ def apply_snr_weight(loss, timesteps, noise_scheduler, gamma):
...
+        skip_weighting (`bool`, *optional*, defaults to `False`):
+            Skip the weighting. When the parsing is skipped, it is forced True.
+    """
+    max_length = (tokenizer.model_max_length - 2) * max_embeddings_multiples + 2
+    if isinstance(prompt, str):
+        prompt = [prompt]
+
+    prompt_tokens, prompt_weights = get_prompts_with_weights(tokenizer, prompt, max_length - 2)
+    prompt_tokens = [token[1:-1] for token in tokenizer(prompt, max_length=max_length, truncation=True).input_ids]
+    prompt_weights = [[1.0] * len(token) for token in prompt_tokens]
```
swfsql, line 287: @AI-Casanova  Please correct me if I'm wrong, but this appears to overwrite the previous value of `prompt_weights` to a fixed value?

AI-Casanova: Let me consult the source material.

AI-Casanova: Oof I think I screwed up an if/then deletion, will push a fix, and test as soon as I get home.

swfsql:
I've made some tests, and I don't have too many conclusions so far.

I've used a single image for training with a lr of 5e-4 (quite high I guess) for 1k steps, generating previews every 25 steps, and trained different lora models by changing the captions.

<details> 
  <summary>training image</summary>
 </img>
</details>

- lora-A image caption: `a photo of a (shs:1.0) (shoe:1.0)`
- lora-B image caption: `a photo of a (shs:1.8) (shoe:0.3)`
- lora-C image caption: `a photo of a (shs:0.3) (shoe:1.8)`
- lora-D image caption was two copies of the same image, one with a caption like lora-B and the other image with a caption like lora-C.  
Then I tried generating horses (not shoes), such as `A shs horse walking on the street`, to see what I would get.

<details> 
  <summary>loss graph</summary>
 </img>
</details>

The conclusions I got:
- All loras: They got very similar loss (except lora-D, which was a little bit different but ended up close to the others).
- lora-B: training a token (`shs`) with a high attention let you use that token also with a high attention without "exploding" the image.
- lora-D: the only one to completely overft. The last 20% of previews (last 200 training steps) generated basically the exact input image (the shoe), not a horse.

The models generated different images, but I couldn't see any clear difference when swapping the attentions like those tests that I did.

AI-Casanova:
Sorry about the dirty git log, I had to force the repo to merge. 

Thanks to @swfsql sharp eye, I caught the issue that was preventing weighted embeddings from calculating effectively.

> All loras: They got very similar loss

This is actually to be expected. Loss is driven almost entirely (per step at least) by the number of timesteps in that particular batch.

In fact, with a set seed, changing caption weights leaves loss within a rounding error, yet significantly changes the genned images
image
xyz_grid-0089-RPGv4
xyz_grid-0088-RPGv4



Captions were created with https://github.com/pharmapsychotic/clip-interrogator-ext and the "Best" setting which creates a fairly long caption. Captions were minimally cleaned, then the token and class were wrapped with `(token class:1)` and `(token class:1.2)` respectively. This makes background and putative style tags less important in comparison to the token and class.

AI-Casanova:
@TingTingin correct on both accounts. 

I believe you can also technically use negative weights, but I have not tested them, and they would not operate in the sense of a negative prompt during generation.

AI-Casanova:
@kohya-ss Awesome!

AI-Casanova:
Of course this is when I realize I didn't need to duplicate code, and could have just imported `lpw_stable_diffusion.py` directly. Well the code runs fine as is, so no need to refactor at this point.

enranime:
Sorry but is there a method to not include captions weight into the token length? because when I train with weight captions the token length its always exceed the limit (225)

CognitiveDiffusion:
This is amazing. 

When will we see this great feature for SDXL?

And would it be possible to ALSO add a Tag that let's the Training know which "tags" (1 or 2 long, natural language tags) are for TextEncoderG?

It could be (G: #this is a natural language prompt#).

And all other prompts (without (G: )) could be directed towards TextEncoderI (which is trained on tags).

Please give me feedback on this. I feel like this could be huge for SDXL training.
@bmaltais @AI-Casanova

AI-Casanova:
@CognitiveDiffusion honestly for SDXL LoRA training the best solution I've found (at least for people/characters) is training a TI, and then training a LoRA using the TI as a prompt (the two need to be used together in inference as well) 

I got this idea from cloneofsimo's PTI implementation, and it seems to work wonderfully for the two text encoders of SDXL, far better than text encoder training. 

I'd share my implementation, but it's completely hackneyed and hard coded. Unfortunately I'm on OT at work and don't have the time to develop it. 

@kohya-ss would you be interested in adding the ability to load TI embeddings into `train_network.py`? The results are quite impressive.

PR #400, opened 2023-04-11
Recursive support for captioning/tagging scripts

Linaqruf:
Hi! Thanks for the good work as always.

In this PR, I want to propose some small changes, mostly about recursive support for captioning/tagging scripts so users can annotate their dataset recursively. However, I am not sure if it is implemented correctly, so I hope for your review to make it better.

For recursive args, I borrowed `glob_images_pathlib()` from train_util.

1. tag_images_by_wd14_tagger.py
- Added `--recursive` to find and preprocess datasets inside sub-directories.
- Added `--remove_underscore` args.
- Added `--undesired_tags`, so users can delete undesired tags from the tagging process.
- Added character tags (category = `4`) as well as `--character_threshold`. I don't know if it's a good idea, but I think it might be helpful for character training. SmilingWolf released [SmilingWolf/wd-v1-4-convnextv2-tagger-v2](https://huggingface.co/SmilingWolf/wd-v1-4-convnextv2-tagger-v2), and I think it's a great and up-to-date model for character tagging.
- Renamed `--thresh` to `--general_threshold`.
- Added `--frequency_tags` to print tag frequency after the tagging process is done.
- Due to character tags added, I updated how `--debug` works with the following new template:
```
    {image_path} 
    Character Tags = {character_tags}
    General Tags = {general_tags}
```

2. make_captions_by_git.py
- Added `--recursive` to find and preprocess datasets inside sub-directories.

3. make_captions.py
- Added `--recursive` to find and preprocess datasets inside sub-directories.

4. prepare_buckets_latents.py
- Added `--recursive` to find and preprocess datasets inside sub-directories. I thought it was already covered by `--full_path`, but it cannot preprocess datasets inside sub-directories and make latents of them. It might not be useful for multi-concept training, but I think it's useful for multi-directories training. So users can keep the datasets inside the respective folder and preprocess them without needing to re-run the scripts every time.

5. convert_diffusers20_original_sd.py
- It might be a `QoL` rather than a new change, but I added `--save_precision_as` with choices of `["fp16", "bf16", "float"]`.

Thank you!

### DIFF
--- a/finetune/make_captions.py
+++ b/finetune/make_captions.py
@@ -4,6 +4,7 @@ import os
 import json
 import random
 
+from pathlib import Path
 from PIL import Image
 from tqdm import tqdm
 import numpy as np
@@ -72,7 +73,8 @@ def main(args):
     os.chdir('finetune')
 
   print(f"load images from {args.train_data_dir}")
-  image_paths = train_util.glob_images(args.train_data_dir)
+  train_data_dir_path = Path(args.train_data_dir)
+  image_paths = train_util.glob_images_pathlib(train_data_dir_path, args.recursive)
   print(f"found {len(image_paths)} images.")
 
   print(f"loading BLIP caption: {args.caption_weights}")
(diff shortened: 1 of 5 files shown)


Linaqruf:
I'm sorry for the late reply, and thanks for the merge!

I forgot emoji tags exists 😅 

Also I'm sorry but I cant use my pc right now for an hour or two

How about using the same `clean_tags_and_captions.py` approach? Like this:
```python
  tags = tags.replace('^_^', '^@@@^')
  tags = tags.replace('_', ' ')
  tags = tags.replace('^@@@^', '^_^')
```

PR #416, opened 2023-04-15
Add automatic ducument system for this project and push to the online readthedocs website

mio2333:
I would like to build a elegant document system for this. 
Preview can be seen from the [sd-scripts.readthedocs.io](https://sd-scripts.readthedocs.io/en/latest/index.html).
It's now under my account and built by myself,but I think it can add to a continuous github workflow. 
If you accept it, I would delete this url and it's content.


mio2333:
Next time I will add a script to automatically generate the corresponding API documentation from the function.
And I would like to translate this to English and Chinese.

mio2333:
Also, I would like to ask if you have an idea to release this project as a pip package.
And for better structure,I recommend creating a [recipe-like concept](https://github.com/espnet/espnet/tree/master/egs2) using espnet as a template.
With this idea,we can regard every finetune work as a recipe,which allow us share the whole reproducible work via huggingface.

For example,[this ](https://huggingface.co/mio/amadeus) is  a text-to-speech project trained by me using espnet. Not only it shares the trained model,but also shares the setting,command,and results.

mio2333:
Ok,thank you for your careful consideration and reply! I fully understand your statement,and I appreciate your work so I want to make this project better.

I want to add the documentation feature because I want to translate these markdown to my country's language. It will be better to do under the Sphinx since there are many previous similar work. But you are right that Python docstrings will take some time to fullfil this pipline.

And also I suggest the pip package distribution because I think the project should be organized better. For example, the subfolder `library` will be installed in the global 'site-packages` with the instructions in the README. This is not a good idea like you set a variable with the name `main` in python. It can work with other people will not do the same thing. And also for developer, if you edit any file under `library`,you should install the project with pip again, which is confusing. Athough it can be solved by using `pip instll -e` method.

Finally, thank you for your hard work. It maybe too early to do this as you said many features and bugs should be considered firstly.

mio2333:
I think are right,the two changes are correct and can be made without spending too much time.

Issue #443, opened 2023-04-24
[Feature]Support for Lion8bit


sdbds:
https://github.com/TimDettmers/bitsandbytes/pull/188

i try to update bitsandbytes to 0.38.1 and modified main.py it works.

but libbitsandbytes_cuda116.dll is too old so cant find new function for lib.clion32bit_g32

sdbds:
i find bitsandbytes 0.38.1 for windows but not test yet

PR #447, opened 2023-04-25
Support for Lion8bi

sdbds:
test bitsandbytes 0.38.1 for lion8bit
https://github.com/kohya-ss/sd-scripts/issues/443


sdbds:
bdc4543e86fc5a2b487b835ab5e2c30

ok，tested well.
lion8bit is a bit faster than lion.

anyone want to try this?

sdbds:
> Thank you for this! Lion8bit is quite interesting!
> 
> One concern I have is the stability of the new version of bitsandbytes. If there is some problem in training, it will affect many people. So I think it might be a good idea to make it an optional choice for a while. In other words, keep the default installation guide and add an optional guide for using the new bitsandbytes for those who want it..
> 
> I am also concerned about the source of the DLL. As for the license, I think the bitsandbytes binary is probably the same as the license of the original repository. But there might be some unexpected problem. (Maybe I'm being too paranoid, for example, malware.)
> 
> I would appreciate it if you could tell me where to find it.
> 
> I really wish I could build it myself, but it's not easy...

Thank you for replying!
yeah，new version maybe not stable,we can just put it for optional 

I found bitsandbytes 0.38.1 dll in here.
https://github.com/jllllll/bitsandbytes-windows-webui
Its author says he made form here
https://github.com/jllllll/bitsandbytes/tree/cmake_windows
Your concern is correct, it's better to compile a secure version from the above fork,it is esay to make with modified Windows version.

i tried to build it too,too hard...

sdbds:
> Thanks for all the updates!
> 
> I was able to successfully build the dll of bitsandbytes based on the repository you listed. However, I am not sure if the dll will work in all environments.
> 
> I may have been somewhat paranoid😅. It looks like we can use the dll from that repository.
> 
> Also, there seems to be a pre-built wheel for Windows in the repository from which the repository was forked. How about using this one? With this, we don't need copy dlls and modify sources.
> 
> https://github.com/acpopescu/bitsandbytes/releases

I think this is pretty good too, I did it this way, just using one of the dlls which biult form cuda118.

image

The only difference between these two forks is the correction of the CUDA loading path.

sdbds:
> Thank you for your comment and sorry for the delay. The difference seems to be related to compilation only, so I think we can use the wheel in acpopescu's repo.
> 
> In order to make the selection of the version of `bitsandbytes` optional, would it be possible for you to restore the contents of `bitsandbytes_windows` folder and the `requirements.txt`?
> 
> Then I think we can add the optional instruction for using the wheel.
> 
> Thank you again for your work!

OK！i had reversed them only change for lion8bit for trainuntil

Ashoka74:
Hey! Thank you for this issue! 

I am trying to install bitsandbytes on cuda 11.8 on my Windows for another app but am stuck here as well.
Is there a way you could share your bitsandbytes weights for cuda 11.8? @sdbds

sdbds:
> Hey! Thank you for this issue!
> 
> I am trying to install bitsandbytes on cuda 11.8 on my Windows for another app but am stuck here as well. Is there a way you could share your bitsandbytes weights for cuda 11.8? @sdbds

i dont have some special action for it , just install Windows version in [here](https://github.com/jllllll/bitsandbytes-windows-webui/blob/main/bitsandbytes-0.38.1-py3-none-any.whl)  
if you give some images about it , maybe i can give some suggestions

PR #455, opened 2023-04-26
Support for more Dadaptation

sdbds:
add more optimizer_type for DAdaptation
DAdaptation(DAdaptAdam)
DAdaptAdaGrad
DAdaptAdan
DAdaptSGD

i will test tomorrow


sdbds:
It works well.Even i try to use DAdaptAdan for block weight lr....
image

sdbds:
I found something interesting——pytorch_optimizer
included 47 optimizers, 6 lr schedulers and Many methods can be directly invoked.。
https://github.com/kozistr/pytorch_optimizer

PR #468, opened 2023-04-30
Add LoRA training supports for diffusers version from diffuser==0.10.x to latest 0.16.1

hellojixian:
majorly 2 things have been changed since v0.10
1, xformers op injection has been handled by diffusers nicely, after v0.10, and if using the latest 0.16.x the memory consumption is no longer any big difference in regardless turn on/off the xformers option

2, the LoRA create_network function make the network layer duplicated in the latest lora version.

Please read the details in the code change.
Thank you for a great job

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -1849,12 +1849,11 @@ def replace_unet_cross_attn_to_xformers(unet):
         out = self.to_out[1](out)
         return out
 
-    if hasattr(diffusers.models.attention, "CrossAttention") and \
+    if diffusers.__version__ >= "0.12":
+        unet.enable_xformers_memory_efficient_attention(attention_op=xformers.ops.MemoryEfficientAttentionFlashAttentionOp)
+    elif hasattr(diffusers.models.attention, "CrossAttention") and \
         hasattr(diffusers.models.attention.CrossAttention, "forward"):
         diffusers.models.attention.CrossAttention.forward = forward_xformers
-    elif hasattr(unet, "enable_xformers_memory_efficient_attention"):
-        unet.enable_xformers_memory_efficient_attention(attention_op=xformers.ops.MemoryEfficientAttentionFlashAttentionOp)
-        print(unet.enable_xformers_memory_efficient_attention())
     else:
         print('Do nothing...')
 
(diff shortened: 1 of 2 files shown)


hellojixian:
Thanks for the feedback, i have tested it with below environment 
xformers==0.0.19
transformers==4.28.1
diffusers==0.16.1 

torch==2.0.0
torchaudio==2.0.0
torchvision==0.15.0

(the diffuser will work since 0.12 - 0.16) i have tested them all 

the important thing is this patch works with torch 2.0,  in torch 2.0 they implemented its own improvement of the attention module, so natively without xformers and any other plugin, it runs 4 times faster than torch 1.x.
i would also recommend you try to switch to the latest torch v2, after this patch, the code works totally fine with torch 2.0

my LoRA training parameters are below:

accelerate launch --num_cpu_threads_per_process=2 "train_network.py" --enable_bucket --pretrained_model_name_or_path="runwayml/stable-diffusion-v1-5" --train_data_dir="/proj/suchka/image-generation/lora/jixian/img" --resolution=512,512 --output_dir="/proj/suchka/image-generation/lora/jixian/model" --logging_dir="/proj/suchka/image-generation/lora/jixian/log" --network_alpha="128" --save_model_as=safetensors --network_module=networks.lora --text_encoder_lr=5e-7 --unet_lr=5e-6 --network_dim=128  --output_name="jixianwang_v3" --lr_scheduler_num_cycles="10" --learning_rate="5e-6" --lr_scheduler="constant" --train_batch_size="24" --max_train_steps="2000" --save_every_n_epochs="1" --mixed_precision="bf16" --save_precision="bf16" --caption_extension=".txt" --cache_latents --optimizer_type="AdamW8bit" --max_data_loader_n_workers="0" --bucket_reso_steps=64 --xformers --bucket_no_upscale --noise_offset=0.1 --seed=1234

n00mkrad:
Still getting the same error on your branch, btw

feffy380:
This patch breaks the `--mem_eff_attn` option because of the CrossAttention module refactor. There are also reports that, at least on ROCM, torch 2.0's SDP attention actually *increases* vram usage rather than reducing it, so it's not a drop-in replacement.

PR #471, opened 2023-05-02
Multi-Resolution Noise

### LINKED ISSUE #334: Request: Multi-Resolution Noise
Similar to offset noise
https://wandb.ai/johnowhitaker/multires_noise/reports/Multi-Resolution-Noise-for-Diffusion-Model-Training--VmlldzozNjYyOTU2

pamparamm:
Support for multi-resolution noise ([pyramid noise implementation by Jonathan Whitaker](https://wandb.ai/johnowhitaker/multires_noise/reports/Multi-Resolution-Noise-for-Diffusion-Model-Training--VmlldzozNjYyOTU2))
Works fine with LoRA and LyCORIS, haven't tested with other training approaches.
Adds new arguments: `--multires_noise_iterations` to enable multires noise (recommended values are 6-10), and `--multires_noise_discount` ([recommended value](https://huggingface.co/johnowhitaker/pyramid_noise_test_600steps_08discount) by Jonathan is 0.8. I find that value too high for LoRAs with small datasets, so I would recommend something around 0.3-0.5).

Resolves #334

### DIFF
--- a/fine_tune.py
+++ b/fine_tune.py
@@ -21,7 +21,7 @@ from library.config_util import (
     BlueprintGenerator,
 )
 import library.custom_train_functions as custom_train_functions
-from library.custom_train_functions import apply_snr_weight, get_weighted_text_embeddings
+from library.custom_train_functions import apply_snr_weight, get_weighted_text_embeddings, pyramid_noise_like
 
 
 def train(args):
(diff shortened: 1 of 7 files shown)


bmaltais:
Queck question... when using multi-res should the original noise option be left to 0... or is it still required to be set?

pamparamm:
@bmaltais According to #477 you can't set both `noise_offset` and `multires_noise_iterations` as they both enable different noise techniques.

feffy380:
@pamparamm Can you post some examples showing the effect of this option?

johnowhitaker:
Oh cool! Fun to see this in use. 0.8 was from early tests and I've since (like you) figured out that much lower values work better :)

Issue #251, opened 2023-03-02
gradient accumulation forces to produce more epochs


specblades:
Please unlink num epochs from gradient accumulation steps

PR #483, opened 2023-05-07
fix num_train with gradient_accumulation_steps

sdbds:
https://github.com/kohya-ss/sd-scripts/blob/e6ad3cbc66130fdc3bf9ecd1e0272969b1d613f7/library/train_util.py#LL2685C4-L2685C4

someone mentions it in #251 .

it is strange that when use gradient_accumulation_steps and num_train also increase.
i found it and after researching a lot of information, when the gradient_accumulation_steps increase, the total number of steps should decrease instead of remaining unchanged.


sdbds:
i found some bug when use other scheduler,maybe i will more test it

sdbds:
> gradient accumulation simulates having a larger batch size it would be more confusing if it didn't increase imo

i know it but if we use more big gradient accumulation such as 64, num_train_steps will increase exponentially.

Generally speaking, when the batch size increases and num_train_steps increase simultaneously, the time should remain unchanged. However, in reality, due to gradient accumulation being a cumulative process, the time also increases proportionally.

Although I used the max_train_steps to avoid issues, there are still problems with correctly calculating the learning rate on the scheduler fix, such as not being able to reach 0 using cosine in the final max_step. After testing this code, there are still some bugs that need further testing.

sdbds:
> Thank you for raising the issue!
> 
> The relationship between gradient_accumulation_steps and the number of training steps is really annoying. It's also hard to see how it changes with multi-GPU training.
> 
> I am using the Diffusers DreamBooth sample as is for this part of the process, so it probably won't be a problem. It is not intuitive, but as TingTingIn said, it seems to make sense.

ok，i got it.

PR #489, opened 2023-05-09
fix pynoise code bug

sdbds:
I discussed pynoise with my friend, and he pointed out a coding bug in it. After the loop starts, w and h will be overwritten, requiring renaming new variable. Upon his testing, this issue indeed exists.

Before↓
image

After↓
image

I consulted with the author, and he agreed that the revised version better reflects the effects mentioned in the article, even though the original version also had good results. He has updated the article and added new notes, so I am submitting this new code to better align with the intended effect.

we can keep old code in a variant,too.

### DIFF
--- a/library/custom_train_functions.py
+++ b/library/custom_train_functions.py
@@ -347,7 +347,7 @@ def get_weighted_text_embeddings(
 
 # https://wandb.ai/johnowhitaker/multires_noise/reports/Multi-Resolution-Noise-for-Diffusion-Model-Training--VmlldzozNjYyOTU2
 def pyramid_noise_like(noise, device, iterations=6, discount=0.1):
-    b, c, w, h = noise.shape
+    b, c, w, h = noise.shape # EDIT: w and h get over-written, rename for a different variant!
     u = torch.nn.Upsample(size=(w, h), mode="bilinear").to(device)
     for i in range(iterations):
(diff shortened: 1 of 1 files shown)


sdbds:
The loop will break once when w or h becomes 1. Here is how I got that answer:

The initial values of w and h are 512.
The loop will iterate 6 times, with i ranging from 0 to 5.
In each iteration, r is a random number between 2 and 4.
w and h are divided by r^i and rounded down to the nearest integer.
If w or h becomes 1, the loop will break.
Let’s assume the worst case scenario, where r is always 4. Then we can calculate the values of w and h in each iteration as follows:

i = 0: w = 512 / 4^0 = 512, h = 512 / 4^0 = 512
i = 1: w = 512 / 4^1 = 128, h = 512 / 4^1 = 128
i = 2: w = 128 / 4^2 = 32, h = 128 / 4^2 = 32
i = 3: w = 32 / 4^3 = 8, h = 32 / 4^3 = 8
i = 4: w = 8 / 4^4 = 2, h = 8 / 4^4 = 2
i = 5: w = 2 / 4^5 = 0, h = 2 / 4^5 = 0
As you can see, when i is 5, both w and h become zero, which is less than one. So the loop will break at this point. Therefore, the loop will break once in the worst case scenario.

PR #507, opened 2023-05-15
Added support for Perlin noise in Noise Offset

HkingAuditore:
https://wandb.ai/johnowhitaker/multires_noise/reports/Multi-Resolution-Noise-for-Diffusion-Model-Training--VmlldzozNjYyOTU2

This article mentioned the possibility of adding Perlin noise to optimize the effect, so I added the Perlin noise function. It can be called by --perlin_noise, the parameter type is an int, which is used to specify the octaves of perlin noise

### DIFF
--- a/library/custom_train_functions.py
+++ b/library/custom_train_functions.py
@@ -373,3 +373,54 @@ def apply_noise_offset(latents, noise, noise_offset, adaptive_noise_scale):
 
     noise = noise + noise_offset * torch.randn((latents.shape[0], latents.shape[1], 1, 1), device=latents.device)
     return noise
+
+
+
+##########################################
+# Perlin Noise
+def rand_perlin_2d(device, shape, res, fade=lambda t: 6 * t ** 5 - 15 * t ** 4 + 10 * t ** 3):
+    delta = (res[0] / shape[0], res[1] / shape[1])
+    d = (shape[0] // res[0], shape[1] // res[1])
+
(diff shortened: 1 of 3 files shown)


HkingAuditore:
image
This is a comparison of the added noises and the pyramid noise

HkingAuditore:
> @HkingAuditore What is the recommended value for octaves? I would like to write it in the documentation.

1 for normal perlin noise, 4 for fractal noise

HkingAuditore:
> How about this code? I am not good at math, so I don't know if this is correct.
> 
> ```
> def perlin_noise(noise, device, octaves):
>     _, c, w, h = noise.shape
>     perlin = lambda: rand_perlin_2d_octaves(device, (w, h), (4, 4), octaves)
>     noise_perlin = []
>     for _ in range(c):
>         noise_perlin.append(perlin())
>     noise_perlin = torch.stack(noise_perlin).unsqueeze(0)   # (1, c, w, h)
>     noise += noise_perlin # broadcast for each batch
>     return noise / noise.std()  # Scaled back to roughly unit variance
> ```

image

Sorry, I may have made some unuploaded changes in my local environment, which caused the uploaded code to fail to run.

Your code works successfully for me! But the noise seems to be visually overexposed (pictured above). I'm not sure if this step of dividing by the standard deviation is necessary. The image below is the result of removing _/noise.std()_ in your code

image

HkingAuditore:
> Thank you for your comment!
> 
> I borrow the code from pyramid noise for the ` /noise.std()`, and totally unsure it is required or not, so your suggestion will be true.
> 
> I will update and make testing for training, and if it works, I will release the feature.

Thanks for your hard work. I'm really looking forward to this working!

Issue #540, opened 2023-05-26
Implementation of new optimizer: Sophia


KaraKaraWitch:
Github: https://github.com/kyegomez/Sophia

According to the paper on arxiv: https://arxiv.org/pdf/2305.14342.pdf It has a 2 time speed up compared to AdamW.

sdbds:
i will test it soon

sdbds:
![image](https://github.com/kohya-ss/sd-scripts/assets/8085926/ac377c3a-5f93-41c7-a77a-80668d6ccaa4)
it works but original pip pack has some bugs need to fix.

sdbds:
2 time speed up and 2x GPU memory required...
so i need to use batch size/2 when i use it

sdbds:
#541 Implemented using custom parameters

recris:
Has anyone benchmarked this? I am curious if the gains described in the paper transfer well to diffusion models in general.

PR #541, opened 2023-05-26
support for Sophia #540

sdbds:
#540 
i tested it but original pip pack have some bugs need to fixs
https://github.com/kyegomez/Sophia/pull/14

after fixed, i will test more

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -1940,7 +1940,7 @@ def add_optimizer_arguments(parser: argparse.ArgumentParser):
         "--optimizer_type",
         type=str,
         default="",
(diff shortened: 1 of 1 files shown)


sdbds:
> how is this on vram compared to 8bitAdam or daadaptation

2x speed up and almost 2x vram required....

sdbds:
> Thank you for this! Sophia optimizer is quite interesting.
> 
> However, Sophia does not seem to require special processing like D-Adaptation, so I wonder it will work if we specify the option like `--optimizer_type="Sophia.Sophia"`.

image

i tested but error...

sdbds:
Since the original repository hasn't released a new package in a month, I decided to use another repository. 
After testing, no additional modifications are needed, so this pr is closed. 
If you want to use this optimizer, ↓
`pip install -U --no-deps pytorch-optimizer`
`--optimizer_type=pytorch_optimizer.SophiaH`

PR #545, opened 2023-05-29
Dropout and Max Norm Regularization for LoRA training

AI-Casanova:
This PR adds Dropout and Max Norm Regularization [[Paper]](https://www.cs.toronto.edu/~rsalakhu/papers/srivastava14a.pdf) to `train_network.py`

Dropout randomly removes some weights/neurons from calculation on both the forward and backward passes, effectively training many neural nets successively like so:
image
This encourages the LoRA to diversify its training, instead of only picking a few weights to continuously update, hopefully reducing overtraining. 

Max Norm Regularization calculates the L2 norm of the weights at each key and if they exceed the cutoff, scales the entire key by a factor to bring them in line, (mentioned in section 5.1 of the paper) This works because the relationships between weights in a layer seem to be more important than the total magnitude. 
When enabled, adds logging for TensorBoard, and an average norm value and number of keys scaled each step to the progress bar. 

Either option can be used independently:

- Dropout suggested setting >0.3
- Max Norm suggested setting = 1 (You can also set it high enough to never trigger ie 10 to watch Tensor Board and see where a good point to set it at might be)

Example of training with dropout [0.5,0.25,0.10.05,0] and Max Norm 1 all other settings deterministic
dropout

Notes for @kohya-ss 
Dropout requires Xformers, and I didn't know how you wanted to do the assertion for that
Dropout requires the Cutlass kernel for GPUs with Capability <8 (A100, 4090 etc) this has been tested on a Colab T4 with xformers 0.0.19, no idea minimum requirements.
I believe I passed dropout in a way that won't interfere with the other trainers `(dropout=None)` in the function call, but should be checked.
Also I'm currently scaling lora_up and lora_down by  `ratio**0.5` as a scalar should be commutative when multiplied to a matrix multiplication (ie `matmul(r*A, B) = matmul(A, B*r), matmul(sqrt(r)*A, sqrt(r)*B)`) , will do further testing to confirm whether to remain this way or only multiply up or down by the full ratio.


AI-Casanova:
You may be correct on the issue of dropout. The main thrust of this PR was originally Max Norm, as I hadn't yet figured out that Cutlass kernel supported other GPUs. I will look into your suggestions there.

I'm fairly confident that Max Norm will work as expected, as I've been experimenting in post-processing LoRA and full merged models, and have found that key-wise rescaling can help reduce the effects of overtraining on both. 

[This Publicly available LoRA](https://civitai.com/models/69108/samuraipunkai) originally had a max key norm of about 6, when rescaled to 1 (center columns) the effect was definitely reduced, but its interaction with other LoRA (not shown) was improved greatly. When applied during training, a norm constraint doesn't seem as drastic, as other weights are encouraged to be explored. 
norm

As a side note, in full models, Add Difference merges seem to blow up norms, leading to burning at lower and lower CFG, another motivation for keeping a LoRA layer from being overpowered.

IdiotSandwichTheThird:
> [This Publicly available LoRA](https://civitai.com/models/69108/samuraipunkai) originally had a max key norm of about 6, when rescaled to 1 (center columns) the effect was definitely reduced, but its interaction with other LoRA (not shown) was improved greatly.

Do you have any code available for rescaling max key norms on pre-trained lora? It sounds pretty interesting.

XZiar:
I remmeber when applying lora, weights will be scaled with `alpha/dim`, so maybe the maxnorm should also consider this?

AI-Casanova:
I see rank and module dropout, very nice! Looking forward to exploring the new hyperparameter space this creates.

dill-shower:
Can we use dropout when training the sd model and not just lora?

DuroCuri:
Dropout requires the Cutlass kernel for GPUs with Capability <8 (A100, 4090 etc) this has been tested on a Colab T4 with xformers 0.0.19, no idea minimum requirements. Is this means, if I am using rtx 3090, I can not use dropout? It is depressed for me...

feffy380:
@rkfg If you're adjusting settings so it never scales keys, max norm regularization isn't doing anything since scaling keys is its whole thing

AI-Casanova:
@feffy380 he's using it as a gauge, which also can be done by setting a higher norm, and I believe tensorboard should be collecting max norm data, but it's been a while since I set it up.

Being aware of the max norm is just as valid, if not perhaps more so than forcing it down via my regularization

AI-Casanova:
@rkfg there's some interesting discussion about loss over here if you're interested in the nitty gritty https://github.com/kohya-ss/sd-scripts/discussions/294#discussioncomment-7125083

rkfg:
@AI-Casanova yes, I use it exactly as a progress indicator and it works wonderfully! Thanks for the link, I'll give it a read.

mykeehu:
> In my experience 1 is the best setting, but most importantly I can tell the lora is overtrained when it starts scaling keys! So I then adjust the alpha in such a way that max_key_norm hits the ceiling (which is 1 in this case) but doesn't push further causing key scaling. I use cosine with restarts scheduling and it takes a few attempts to choose the right alpha so that max_key_norm stops growing roughly when the learning rate drops low. But then the lora comes out absolutely perfect, I've never had such great results before! Very flexible and yet perfectly captures the subject.

If I understand correctly, the max_norm/average_key_norm is the value to look at on Tensorboard, so if it reaches 1, is Lora good?
image

rkfg:
I was talking only about `max_key_norm`, not average. In your case it hits 1 immediately and this feature starts scaling a lot of keys down, I try to keep those at minimum, around 1-3 tops (you get 100 and more). So while I only talked about my experience which isn't a dogma, try to lower the alpha value. If you hit 100 scaled keys this fast, divide the alpha by 100 for example or even more, so `max_key_norm` hits 1 at the very end of training.

mykeehu:
It's cosine_with_restarts training with AdamW8bit, and here the scaling always went up. Should I find another method?

rkfg:
I don't think the scheduler matters much, it's about your learning rate and Network Alpha that multiplies the LR by alpha/rank (so if your rank is 128 and alpha is 1, your LR would be 128 times less than specified). Again, my observation doesn't apply to all subjects, some come out undertrained and I need to raise the scaling ceiliing to 2 or even 3.

mykeehu:
At 64/1 I get this scaling and it still looks a bit strong.
image

rkfg:
Try below 1 then, alpha is fractional.

This all isn't a rule or a guide, just my observation. Since we don't have any meaningful indicators of the training progress, this one turned out to work best. Loss is meaningless for loras, it shows almost the same values no matter what parameters you set, and if you change the seed these values also change unpredictably and it's impossible to tell on which epoch you under/overtrained it. So this idea could be a total coincidence for me but this parameter at least depends on the total learning *progress*, it grows even after the scheduler restarts and it reflects the important values (LR, Alpha at least)

mykeehu:
Yesterday I looked at normalization, but it seems to work fine for me only with cosine scheduler, with the others it generates huge losses for some reason. But with cosine it gives nice results under AdamW, Prodigy or Adafactor optimizer.

Seedmanc:
I'd like to just interject for a moment, why is it suggested to modify alpha to reduce learning rate instead of directly reducing the rate itself? Are there more factors at play? In fact I don't understand the point of this alpha parameter at all if all it does is linearly reducing the learning rate. Is there more?

WetOnTheWater:
@AI-Casanova do you know if these dropout regularization techniques would work well on Flux?

PR #551, opened 2023-05-31
add controlnet training

ddPn08:
Add ControlNet training code.
As soon as all the tests are finished, I will write the details and open it.

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -401,6 +401,54 @@ class FineTuningSubset(BaseSubset):
         return self.metadata_file == other.metadata_file
 
 
+class ControlNetSubset(BaseSubset):
+    def __init__(
+        self,
+        image_dir: str,
+        conditioning_data_dir: str,
+        caption_extension: str,
+        num_repeats,
+        shuffle_caption,
+        keep_tokens,
+        color_aug,
+        flip_aug,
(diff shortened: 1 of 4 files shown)


library/train_util.py
```diff
@@ -1787,6 +2103,8 @@ def backward(ctx, do):
 
 def replace_unet_modules(unet: diffusers.models.unet_2d_condition.UNet2DConditionModel, mem_eff_attn, xformers):
     # unet is not used currently, but it is here for future use
+    unet.enable_xformers_memory_efficient_attention()
+    return
```
ddPn08, line 2107: It should be fixed.

sdbds:
Can we use original ControlNet repository's preprocesser to generate
conditioning data for conditioning_data_dir ?
I think this is very convenient, and it also allows for updates when new preprocessors are added to Controlnet in the future.

ddPn08:
If you plan to upgrade the version of diffusers, I'll use ControlNet implemented with diffusers as it is.
And I realized that before this PR, I needed a PR to bump the version of diffusers.

ddPn08:
I will make it open once I write the documentation.

ddPn08:
I got it. I open this PR.

PR #589, opened 2023-06-14
add masked loss implementation

recris:
This is mostly a rebase of https://github.com/kohya-ss/sd-scripts/pull/236

Relevant differences:
- Instead of looking for `.mask` files it will now look for a matching PNG file in the `mask` sub-directory
- When computing MSE loss, mask values are rescaled by the mask mean value.

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -1092,10 +1092,10 @@ class BaseDataset(torch.utils.data.Dataset):
                 image = None
             elif image_info.latents_npz is not None:  # FineTuningDatasetまたはcache_latents_to_disk=Trueの場合
                 latents, original_size, crop_ltrb, flipped_latents = load_latents_from_disk(image_info.latents_npz)
-                mask = load_mask(image_info.absolute_path, image_info.resized_size) / 225
+                mask = load_mask(image_info.absolute_path, image_info.resized_size) / 255
                 if flipped:
                     latents = flipped_latents
-                    mask = mask.flip(mask, dims=[3])
+                    mask = np.flip(mask, axis=1)
                     del flipped_latents
                 latents = torch.FloatTensor(latents)
 
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -1083,8 +1083,10 @@ class BaseDataset(torch.utils.data.Dataset):
                 image = None
             elif image_info.latents_npz is not None:  # FineTuningDatasetまたはcache_latents_to_disk=Trueの場合
                 latents, original_size, crop_ltrb, flipped_latents = load_latents_from_disk(image_info.latents_npz)
+                mask = load_mask(image_info.absolute_path, image_info.resized_size) / 225
                 if flipped:
                     latents = flipped_latents
+                    mask = mask.flip(mask, dims=[3])
                     del flipped_latents
                 latents = torch.FloatTensor(latents)
 
@@ -1093,9 +1095,6 @@ class BaseDataset(torch.utils.data.Dataset):
                 # 画像を読み込み、必要ならcropする
                 img, face_cx, face_cy, face_w, face_h = self.load_image_with_face_info(subset, image_info.absolute_path)
                 im_h, im_w = img.shape[0:2]
-                # loss mask is alpha channel, separate it
-                mask = img[:, :, -1] / 255
-                img = img[:, :, :3]
 
                 if self.enable_bucket:
                     img, original_size, crop_ltrb = trim_and_resize_if_required(
@@ -1131,9 +1130,12 @@ class BaseDataset(torch.utils.data.Dataset):
                 if flipped:
                     img = img[:, ::-1, :].copy()  # copy to avoid negative stride problem
 
+                # loss mask is alpha channel, separate it
+                mask = img[:, :, -1] / 255
+                img = img[:, :, :3]
+
                 latents = None
                 image = self.image_transforms(img)  # -1.0~1.0のtorch.Tensorになる
-                mask = torch.from_numpy(mask)
 
             images.append(image)
             latents_list.append(latents)
@@ -2157,9 +2159,10 @@ def load_mask(image_path, target_shape):
 
     if os.path.exists(mask_path):
         try:
-            mask = np.array(Image.open(mask_path))
+            mask_img = Image.open(mask_path)
+            mask = np.array(mask_img)
             if len(mask.shape) > 2 and mask.max() <= 255:
-                result = np.array(Image.open(mask_path).convert("L"))
+                result = np.array(mask_img.convert("L"))
             elif len(mask.shape) == 2 and mask.max() > 255:
                 result = mask // (((2 ** 16) - 1) // 255)
             elif len(mask.shape) == 2 and mask.max() <= 255:
@@ -2175,8 +2178,7 @@ def load_mask(image_path, target_shape):
 
     # stretch mask to image shape
     if result.shape != target_shape:
-        print(f"{mask_path} does not match image dimensions, resizing")
-        result = cv2.resize(result, dsize=target_shape, interpolation=cv2.INTER_NEAREST)
+        result = cv2.resize(result, dsize=target_shape, interpolation=cv2.INTER_LINEAR)
 
     return result
 
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -135,6 +135,9 @@ class ImageInfo:
         self.text_encoder_outputs1: Optional[torch.Tensor] = None
         self.text_encoder_outputs2: Optional[torch.Tensor] = None
         self.text_encoder_pool2: Optional[torch.Tensor] = None
+        # Masked Loss
+        self.mask: np.ndarray = None
+        self.mask_flipped: np.ndarray = None
 
 
 class BucketManager:
@@ -1050,6 +1053,7 @@ class BaseDataset(torch.utils.data.Dataset):
         input_ids2_list = []
         latents_list = []
         images = []
+        masks = []
         original_sizes_hw = []
         crop_top_lefts = []
         target_sizes_hw = []
@@ -1071,8 +1075,10 @@ class BaseDataset(torch.utils.data.Dataset):
                 crop_ltrb = image_info.latents_crop_ltrb  # calc values later if flipped
                 if not flipped:
                     latents = image_info.latents
+                    mask = image_info.mask
                 else:
                     latents = image_info.latents_flipped
+                    mask = image_info.mask_flipped
 
                 image = None
             elif image_info.latents_npz is not None:  # FineTuningDatasetまたはcache_latents_to_disk=Trueの場合
@@ -1087,6 +1093,9 @@ class BaseDataset(torch.utils.data.Dataset):
                 # 画像を読み込み、必要ならcropする
                 img, face_cx, face_cy, face_w, face_h = self.load_image_with_face_info(subset, image_info.absolute_path)
                 im_h, im_w = img.shape[0:2]
+                # loss mask is alpha channel, separate it
+                mask = img[:, :, -1] / 255
+                img = img[:, :, :3]
 
                 if self.enable_bucket:
                     img, original_size, crop_ltrb = trim_and_resize_if_required(
@@ -1124,9 +1133,11 @@ class BaseDataset(torch.utils.data.Dataset):
 
                 latents = None
                 image = self.image_transforms(img)  # -1.0~1.0のtorch.Tensorになる
+                mask = torch.from_numpy(mask)
 
             images.append(image)
             latents_list.append(latents)
+            masks.append(torch.tensor(mask))
 
             target_size = (image.shape[2], image.shape[1]) if image is not None else (latents.shape[2] * 8, latents.shape[1] * 8)
 
@@ -1218,7 +1229,7 @@ class BaseDataset(torch.utils.data.Dataset):
         else:
             images = None
         example["images"] = images
-
+        example["masks"] = torch.stack(masks) if masks[0] is not None else None
         example["latents"] = torch.stack(latents_list) if latents_list[0] is not None else None
         example["captions"] = captions
 
@@ -2132,10 +2143,42 @@ def load_arbitrary_dataset(args, tokenizer) -> MinimalDataset:
 
 def load_image(image_path):
     image = Image.open(image_path)
-    if not image.mode == "RGB":
-        image = image.convert("RGB")
+    if not image.mode == "RGBA":
+        image = image.convert("RGBA")
     img = np.array(image, np.uint8)
+    img[..., -1] = load_mask(image_path, img.shape[:2])
     return img
+
+
+def load_mask(image_path, target_shape):
+    p = pathlib.Path(image_path)
+    mask_path = os.path.join(p.parent, 'mask', p.stem + '.png')
+    result = None
+
+    if os.path.exists(mask_path):
+        try:
+            mask = np.array(Image.open(mask_path))
+            if len(mask.shape) > 2 and mask.max() <= 255:
+                result = np.array(Image.open(mask_path).convert("L"))
+            elif len(mask.shape) == 2 and mask.max() > 255:
+                result = mask // (((2 ** 16) - 1) // 255)
+            elif len(mask.shape) == 2 and mask.max() <= 255:
+                result = mask
+            else:
+                print(f"{mask_path} has invalid mask format: using default mask")
+        except:
+            print(f"failed to load mask: {mask_path}")
+
+    # use default when mask file is unavailable
+    if result is None:
+        result = np.full(target_shape, 255, np.uint8)
+
+    # stretch mask to image shape
+    if result.shape != target_shape:
+        print(f"{mask_path} does not match image dimensions, resizing")
+        result = cv2.resize(result, dsize=target_shape, interpolation=cv2.INTER_NEAREST)
+
+    return result
 
 
 # 画像を読み込む。戻り値はnumpy.ndarray,(original width, original height),(crop left, crop top, crop right, crop bottom)
@@ -2184,12 +2227,17 @@ def cache_batch_latents(
     latents_original_size and latents_crop_ltrb are also set
     """
     images = []
+    masks = []
     for info in image_infos:
         image = load_image(info.absolute_path) if info.image is None else np.array(info.image, np.uint8)
         # TODO 画像のメタデータが壊れていて、メタデータから割り当てたbucketと実際の画像サイズが一致しない場合があるのでチェック追加要
         image, original_size, crop_ltrb = trim_and_resize_if_required(random_crop, image, info.bucket_reso, info.resized_size)
+        # alpha channel contains loss mask, separate it
+        mask = image[:, :, -1] / 255
+        image = image[:, :, :3]
         image = IMAGE_TRANSFORMS(image)
         images.append(image)
+        masks.append(mask)
 
         info.latents_original_size = original_size
         info.latents_crop_ltrb = crop_ltrb
@@ -2207,7 +2255,7 @@ def cache_batch_latents(
     else:
         flipped_latents = [None] * len(latents)
 
-    for info, latent, flipped_latent in zip(image_infos, latents, flipped_latents):
+    for info, latent, flipped_latent, mask in zip(image_infos, latents, flipped_latents, masks):
         # check NaN
         if torch.isnan(latents).any() or (flipped_latent is not None and torch.isnan(flipped_latent).any()):
             raise RuntimeError(f"NaN detected in latents: {info.absolute_path}")
@@ -2216,8 +2264,10 @@ def cache_batch_latents(
             save_latents_to_disk(info.latents_npz, latent, info.latents_original_size, info.latents_crop_ltrb, flipped_latent)
         else:
             info.latents = latent
+            info.mask = mask
             if flip_aug:
                 info.latents_flipped = flipped_latent
+                info.mask_flipped = mask.flip(mask, dims=[3])
 
     # FIXME this slows down caching a lot, specify this as an option
     if torch.cuda.is_available():
@@ -3157,6 +3207,10 @@ def add_dataset_arguments(
     )
     parser.add_argument(
         "--bucket_no_upscale", action="store_true", help="make bucket for each image without upscaling / 画像を拡大せずbucketを作成します"
+    )
+
+    parser.add_argument(
+        "--masked_loss", action="store_true", help="Enable masking of latent loss using grayscale mask images"
     )
 
     parser.add_argument(
--- a/fine_tune.py
+++ b/fine_tune.py
@@ -32,6 +32,7 @@ from library.custom_train_functions import (
     get_weighted_text_embeddings,
     prepare_scheduler_for_custom_training,
     scale_v_prediction_loss_like_noise_prediction,
+    get_latent_masks
 )
 
 
@@ -338,6 +339,11 @@ def train(args):
                     target = noise_scheduler.get_velocity(latents, noise, timesteps)
                 else:
                     target = noise
+
+                if args.masked_loss and batch['masks'] is not None:
+                    mask = get_latent_masks(batch['masks'], noise_pred.shape, noise_pred.device)
+                    noise_pred = noise_pred * mask
+                    target = target * mask
 
                 if args.min_snr_gamma or args.scale_v_pred_loss_like_noise_pred:
                     # do not mean over batch dimension for snr weight or scale v-pred loss
--- a/library/custom_train_functions.py
+++ b/library/custom_train_functions.py
@@ -457,6 +457,21 @@ def apply_noise_offset(latents, noise, noise_offset, adaptive_noise_scale):
     return noise
 
 
+def get_latent_masks(image_masks, latent_shape, device):
+    masks = (
+        image_masks
+        .to(device)
+        .reshape(latent_shape[0], 1, latent_shape[2] * 8, latent_shape[3] * 8)
+    )
+    # resize to match latent
+    masks = torch.nn.functional.interpolate(
+        masks.float(),
+        size=latent_shape[-2:],
+        mode="nearest"
+    )
+    return masks
+
+
 """
 ##########################################
 # Perlin Noise
--- a/library/custom_train_functions.py
+++ b/library/custom_train_functions.py
@@ -458,6 +458,12 @@ def apply_noise_offset(latents, noise, noise_offset, adaptive_noise_scale):
 
 
 def get_latent_masks(image_masks, latent_shape, device):
+    # given that masks lower the average loss this will counteract the effect
+    factor = torch.sqrt(image_masks.mean([1, 2]))
+    factor = torch.where(factor != 0.0, factor, 1.0)
+    factor = factor.reshape(factor.shape + (1,) * 2)
+    image_masks = image_masks / factor
+
     masks = (
         image_masks
         .to(device)
--- a/sdxl_train.py
+++ b/sdxl_train.py
@@ -34,6 +34,7 @@ from library.custom_train_functions import (
     prepare_scheduler_for_custom_training,
     scale_v_prediction_loss_like_noise_prediction,
     add_v_prediction_like_loss,
+    get_latent_masks
 )
 from library.sdxl_original_unet import SdxlUNet2DConditionModel
 
@@ -547,6 +548,11 @@ def train(args):
                     noise_pred = unet(noisy_latents, timesteps, text_embedding, vector_embedding)
 
                 target = noise
+
+                if args.masked_loss and batch['masks'] is not None:
+                    mask = get_latent_masks(batch['masks'], noise_pred.shape, noise_pred.device)
+                    noise_pred = noise_pred * mask
+                    target = target * mask
 
                 if args.min_snr_gamma or args.scale_v_pred_loss_like_noise_pred or args.v_pred_like_loss:
                     # do not mean over batch dimension for snr weight or scale v-pred loss
--- a/train_db.py
+++ b/train_db.py
@@ -35,6 +35,7 @@ from library.custom_train_functions import (
     pyramid_noise_like,
     apply_noise_offset,
     scale_v_prediction_loss_like_noise_prediction,
+    get_latent_masks
 )
 
 # perlin_noise,
@@ -325,6 +326,11 @@ def train(args):
                     target = noise_scheduler.get_velocity(latents, noise, timesteps)
                 else:
                     target = noise
+
+                if args.masked_loss and batch['masks'] is not None:
+                    mask = get_latent_masks(batch['masks'], noise_pred.shape, noise_pred.device)
+                    noise_pred = noise_pred * mask
+                    target = target * mask
 
                 loss = torch.nn.functional.mse_loss(noise_pred.float(), target.float(), reduction="none")
                 loss = loss.mean([1, 2, 3])
--- a/train_network.py
+++ b/train_network.py
@@ -43,6 +43,7 @@ from library.custom_train_functions import (
     prepare_scheduler_for_custom_training,
     scale_v_prediction_loss_like_noise_prediction,
     add_v_prediction_like_loss,
+    get_latent_masks
 )
 
 
@@ -795,6 +796,11 @@ class NetworkTrainer:
                         target = noise_scheduler.get_velocity(latents, noise, timesteps)
                     else:
                         target = noise
+
+                    if args.masked_loss and batch['masks'] is not None:
+                        mask = get_latent_masks(batch['masks'], noise_pred.shape, noise_pred.device)
+                        noise_pred = noise_pred * mask
+                        target = target * mask
 
                     loss = torch.nn.functional.mse_loss(noise_pred.float(), target.float(), reduction="none")
                     loss = loss.mean([1, 2, 3])
--- a/train_textual_inversion.py
+++ b/train_textual_inversion.py
@@ -32,6 +32,7 @@ from library.custom_train_functions import (
     prepare_scheduler_for_custom_training,
     scale_v_prediction_loss_like_noise_prediction,
     add_v_prediction_like_loss,
+    get_latent_masks
 )
 
 imagenet_templates_small = [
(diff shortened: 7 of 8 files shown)


recris:
No worries, if you decide to merge the other PR first I will rebase again

Elevory:
I'm getting some errors with `load_mask()` when it encounters unconventional image sizes. I had to wrap line 727 of `train_util.py` in a try/except as a workaround:

```
img[..., -1] = self.load_mask(image_path)
```

As well as line 1037:

```
example["masks"] = torch.stack(masks) if masks[0] is not None else None
```

Is there a better way to ensure compatibility with bucketing?

Thanks!

Elevory:
Hi,

I moved the initialization of `mask` below the `trim_and_resize_if_required()` function that is used with bucketing. Updated copy of `train_util.py` is attached. It will also catch a couple errors as described in my previous post.

[train_util.zip](https://github.com/kohya-ss/sd-scripts/files/11791347/train_util.zip)

recris:
> Hi,
> 
> I moved the initialization of `mask` below the `trim_and_resize_if_required()` function that is used with bucketing. Updated copy of `train_util.py` is attached. It will also catch a couple errors as described in my previous post.
> 
> [train_util.zip](https://github.com/kohya-ss/sd-scripts/files/11791347/train_util.zip)
> 
> Seems to work in my (brief) testing. **EDIT:** Hmm, it's blowing out the exposure on my DreamBooth models. Not sure if it's an issue with DB or my bucketing fix. Is it normal for masking to cause the model to trend toward white?

I haven't observed anything like this, do you have an example you could share?

Elevory:
Sure, let me provide more context. :)

I am using the `--enable_buckets` flag which produces the following error at the start of training:

```
  File "T:\code\python\kohya-ss\train_db.py", line 507, in <module>
    train(args)
  File "T:\code\python\kohya-ss\train_db.py", line 334, in train
    .reshape(
RuntimeError: shape '[1, 1, 704, 320]' is invalid for input of size 1323504
```

If I understand correctly, this is because on line 965 of `train_util.py`, you are initializing the `mask` variable before applying `trim_and_resize_if_required()` for bucketing on line 970.

But in case I've misdiagnosed the issue, here is my entire configuration:

```
set PROJECT="project_name"
call ./venv/Scripts/activate

accelerate launch --num_cpu_threads_per_process=12 train_db.py ^
  --pretrained_model_name_or_path="T:/code/python/automatic-stable-diffusion-webui/models/Stable-diffusion/general/reliberate_v10.safetensors" ^
  --train_data_dir="T:/code/python/kohya-ss/_in/%PROJECT%/train" ^
  --reg_data_dir="T:/code/python/kohya-ss/_in/%PROJECT%/reg" ^
  --output_dir="T:/code/python/kohya-ss/_out_db/%PROJECT%" ^
  --save_every_n_steps=100 ^
  --sample_every_n_epochs=5 ^
  --train_batch_size=8 ^
  --prior_loss_weight=1.0 ^
  --output_name=%PROJECT% ^
  --resolution=512,512 ^
  --learning_rate=1e-6 ^
  --lr_scheduler="constant" ^
  --lr_warmup_steps=0 ^
  --max_train_steps=5000 ^
  --mixed_precision="fp16" ^
  --gradient_accumulation_steps=1 ^
  --gradient_checkpointing ^
  --save_model_as=safetensors ^
  --clip_skip=1 ^
  --seed=1 ^
  --color_aug ^
  --xformers ^
  --use_8bit_adam ^
  --sample_prompts="./prompts.txt" ^
  --persistent_data_loader_workers ^
  --enable_bucket ^
  --random_crop

pause
```

Now, as far as the image exposure problem: I spliced in your `train_network.py` code into `train_db.py` because I want to use masking with DreamBooth. I haven't actually tried this PR with Loras, so it's possible that this is specifically an issue with DB.

I can see the sample images becoming gradually whiter/brighter/more washed out over the course of training, but it's most evident if I prompt for a dark scene with the finished model. For example, here's _"Will Smith at the nighttime fireworks show"_ in the non-masked model:



And here's the same prompt in the masked version (identical inference settings and pretty similar training settings):



Finally, here's the output from a checkpoint trained twice as long:



> I've been using this with square images and masks, of various sizes. The important part is that mask must be in grayscale format, dimensions exactly matching the corresponding training input.

I can confirm that the masks are all grayscale 8-bit PNGs. Just to be sure, I even tried generating them with your awesome masking tool, which unfortunately didn't make a difference.

Elevory:
Hi @AI-Casanova ,

Thank you for the suggestion. Unfortunately, removing the `--color_aug` flag does not eliminate the issue. Non-masked:



Masked:




I may try preparing another dataset with images at a single resolution to see if bucketing is somehow responsible, but I find it unlikely.

Do you think regularization images might be at fault here? DreamBooth requires hundreds of pictures of the object class, and I haven't created masks for those. Are they perhaps having a stronger-than-intended effect on training?

Elevory:
I think I got it working!

When I moved the initialization of the `mask` var below `trim_and_resize_if_required()`, I neglected to move the "drop alpha channel" line with it. So I was dropping the alpha channel too early.

I'm attaching my updated copies of `train_util.py` and `train_db.py` below. By the way, I'm seeing a noticeable improvement in the quality of my sample images, so masking may be a gamechanger after all - great work @AI-Casanova and @recris!

[masking_for_db.zip](https://github.com/kohya-ss/sd-scripts/files/11830677/masking_for_db.zip)

briansemrau:
I've just reimplemented this on top of the sdxl branch: https://github.com/briansemrau/kohya-ss-sd-scripts/tree/sdxl_lora_mask

recris:
Update:
 - Rebased PR on latest main
 - Made mask loading a bit more robust
 - Added latent masking to all training methods, not just LoRA

I've tested this with LoRA for SD 1.5 and SDXL. Have not yet tested with other methods, if someone would volunteer it would be great :)

recris:
I have removed mask rescaling by mean value. I am concerned that it can make loss magnitude fluctuate in a hard to predict manner because the loss would be scaled differently for each image and cause an "uneven" learning process.

Rescaling would give an increase to loss magnitude but I think we can achieve a similar effect just by adjusting the learning rate.

sashasubbbb:
So what's the dataset configuration for masks? 
Subfolder "mask" and greyscale .png with same file names inside?

What about multiple concepts:

```
dataset/
--5_obj1/
----------mask/
--10_obj2/
----------mask/
```

Is fine?

For some reason i'm running into this error using this configuration.
>   File "B:\AIimages\sd-scripts\sd-scripts\venv\lib\site-packages\torch\utils\data\dataset.py", line 243, in __getitem__
>     return self.datasets[dataset_idx][sample_idx]
>   File "B:\AIimages\sd-scripts\sd-scripts\library\train_util.py", line 1140, in __getitem__
>     masks.append(torch.tensor(mask))
> UnboundLocalError: local variable 'mask' referenced before assignment

sashasubbbb:
I somewhat figured it out, that error occured when using caching latents.

There were other errors about tensor size mismatch when i used all kinds of resolutions of pictures and masks (res of corresponding pictures and masks matched), but disappeared when i rescaled them all to 768,768. It's like it was trying to hook up different mask and got a resolution mismatch? Maybe that one occured because i use batch size of 3, instead of 1?

Also does training resolution must match resolution of dataset? Because when i tried to train on 512,512 with 768,768 dataset i've ran into an error.

So, to start a training i had to:

1) Disable cache latents;
2) Resize all dataset to training resolution.

recris:
> I somewhat figured it out, that error occured when using caching latents.
> 
> There were other errors about tensor size mismatch when i used all kinds of resolutions of pictures and masks (res of corresponding pictures and masks matched), but disappeared when i rescaled them all to 768,768. It's like it was trying to hook up different mask and got a resolution mismatch? Maybe that one occured because i use batch size of 3, instead of 1?
> 
> Also does training resolution must match resolution of dataset? Because when i tried to train on 512,512 with 768,768 dataset i've ran into an error.
> 
> So, to start a training i had to:
> 
>     1. Disable cache latents;
> 
>     2. Resize all dataset to training resolution.

Right now it does not work with _disk_ caching of latents, but in-memory caching should work.

Also I've been testing with a dataset having images of different sizes and is working fine. When the mask does not exactly match the corresponding image then it also should resize the mask automatically.

cheald:
One suggestion: when using masks, consider dividing loss by the mean of the mask. The idea is that you don't want a masked image's loss computation relative to the other samples affected by the portion of the image that is masked. For loss, you really care about how close the predicted noise got to the actual noise for only the masked pixels.

In the current implementation, the mask "knocks out" the contribution to loss for anything but the masked pixels. This means that all else equal, images with less masking will have higher loss. By dividing the loss by the mean of the mask, you boost the observed loss by the same amount that the loss scalar is reduced due to zeroing out the masked pixels.

In my empirical tests, this vastly improved the results of my training. My implementation looks like this:

```python
loss_div = 1.0
if args.masked_loss and batch['masks'] is not None:
    mask = get_latent_masks(batch['masks'], noise_pred.shape, noise_pred.device)
    noise_pred = noise_pred * mask
    target = target * mask
    loss_div = mask.mean()
    if loss_div == 0:
      loss_div =1.0

loss = torch.nn.functional.mse_loss(noise_pred.float(), target.float(), reduction="none")
loss = loss.mean([1, 2, 3])

loss_weights = batch["loss_weights"]  # 各sampleごとのweight
loss = loss * loss_weights / loss_div
```

recris:
> One suggestion: when using masks, consider dividing loss by the mean of the mask. The idea is that you don't want a masked image's loss computation relative to the other samples affected by the portion of the image that is masked. For loss, you really care about how close the predicted noise got to the actual noise for only the masked pixels. Right now, the more masked an image is, the more its loss is damped, due to all the 0 MSE pixels included in the calculation.
> 
> In the current implementation, the mask "knocks out" the contribution to loss for anything but the masked pixels. This means that all else equal, images with less masking will have higher loss. By dividing the loss by the mean of the mask, you boost the observed loss by the same amount that the loss scalar is reduced due to zeroing out the masked pixels.
> 
> In my empirical tests, this vastly improved the results of my training. My implementation looks like this:
> 
> ```python
> loss_div = 1.0
> if args.masked_loss and batch['masks'] is not None:
>     mask = get_latent_masks(batch['masks'], noise_pred.shape, noise_pred.device)
>     noise_pred = noise_pred * mask
>     target = target * mask
>     loss_div = mask.mean()
>     if loss_div == 0:
>         loss_div = 1.0
> 
> loss = torch.nn.functional.mse_loss(noise_pred.float(), target.float(), reduction="none")
> loss = loss.mean([1, 2, 3])
> 
> loss_weights = batch["loss_weights"]  # 各sampleごとのweight
> loss = loss * loss_weights / loss_div
> ```

This is something I had in a previous commit, but decided to roll back. I am not entirely sure dividing by the mean is the best "ratio", I guess we need further experimentation.

AI-Casanova:
My thought on dividing by mean:

Sure the overall loss will be lower with masks, but the goal is to speed up convergence by eliminating extraneous input, and step size can be compensated for by raising the overall learning rate. 

Besides, adjusting a squared loss by a linear factor is far from perfect.

AI-Casanova:
> black = ignore
white = include
grey = ignore 50%

Precisely

Though I now think the mask should be applied to the loss itself instead of the noise, because `(0.5x -0.5y)^2 != 0.5(x-y)^2`

recris:
We can apply the mask to the noise, we just have to use `sqrt(mask)` for the same effect.

For gray regions we're just reducing the loss magnitude in those regions, so in practice what we have is kind of a dynamic learning rate. That is, a gray value of 0.5 means we cutting the LR in half.

AI-Casanova:
Yes we *can* sqrt mask

But I'm looking at clean implementation. 

We can just take the unreduced loss, and call a `function(loss, masks) ... return loss`

recris:
> I somewhat figured it out, that error occured when using caching latents.
> 
> There were other errors about tensor size mismatch when i used all kinds of resolutions of pictures and masks (res of corresponding pictures and masks matched), but disappeared when i rescaled them all to 768,768. It's like it was trying to hook up different mask and got a resolution mismatch? Maybe that one occured because i use batch size of 3, instead of 1?
> 
> Also does training resolution must match resolution of dataset? Because when i tried to train on 512,512 with 768,768 dataset i've ran into an error.
> 
> So, to start a training i had to:
> 
>     1. Disable cache latents;
> 
>     2. Resize all dataset to training resolution.

Can you retest with latest changes? Now it should work as intended regardless of cache being used and mask resizing should be automatic in all circumstances.

sashasubbbb:
Yeah, it seems to work just fine. Thanks, @recris.

library/train_util.py
```diff
@@ -1071,14 +1075,18 @@ def __getitem__(self, index):
...
                     latents = image_info.latents
+                    mask = image_info.mask
                 else:
                     latents = image_info.latents_flipped
+                    mask = image_info.mask_flipped
 
                 image = None
             elif image_info.latents_npz is not None:  # FineTuningDatasetまたはcache_latents_to_disk=Trueの場合
                 latents, original_size, crop_ltrb, flipped_latents = load_latents_from_disk(image_info.latents_npz)
+                mask = load_mask(image_info.absolute_path, image_info.resized_size) / 225
```
cheald, line 1086: This should probably be 255?

Of note: loading masks breaks when loading a latent that was cached to disk if the original image was resized and cropped, because the loaded mask doesn't undergo the same load/crop routine, and ends up mismatched to the latent. When loading latents from disk, the original image is never considered. I think that the current implementation of `trim_and_resize_if_required` prevents replication of the original crop if random_crop is specified, since the coordinates of the crop are not passed back; crop_ltrb is something other than the coords actually used for cropping the image.

I've solved this in my local copy by just persisting the mask in the .npz, but that obviously blows up the cache size.

library/train_util.py
```diff
@@ -1071,14 +1075,18 @@ def __getitem__(self, index):
...
                 if flipped:
                     latents = flipped_latents
+                    mask = mask.flip(mask, dims=[3])
```
cheald, line 1089: `mask` is an np.array here and doesn't have a `flip` method. `np.flip(mask, axis=1)` is probably what's desired here.

library/train_util.py
recris: Whoops! nice catch.

I guess right now it is a limitation - I don't see a reliable way of preserving the crop parameters other than serializing the mask like you did (or a latent sized crop), or at least embed the crop parameters into the npz.

cheald: Keeping, returning, persisting, and retrieving a second set of crop parameters from `trim_and_resize_if_required` would be the much lighter option. Either way, a change to the npz format is likely necessary, or latent disk caching should fail loudly when using masked loss and images which need cropping.

araleza:
Hi, I just wanted to ask why the mask files have to be a separate image, instead of just using the alpha channel of the training image?  It would seem easier to work with that way.

AI-Casanova:
@araleza that's actually how I first implemented it, problem is pre multiplied alphas

recris:
This change has been progressing for months and I'd like to move this into a mergeable state. @kohya-ss I'd appreciate your input on what you think will be needed to eventually merge this PR.

Here are a few things that I still want to implement:
- Instead of relying on an implicitly defined `mask` folder, use a new subset parameter `loss_mask_dir` so that users can set a folder explicitly. Same file name rules apply. if `loss_mask_dir` is unspecified then we can assume `<image_dir>/mask` by default.
- Update the latent disk cache format to include an optional "mask" tensor (and "mask_flipped" too). This tensor should have same shape as the latent tensor. The cache generation script needs to be updated to embed the latent mask when writing npz files. This disk format change should be backwards compatible, however users that want use the masking feature would have to regenerate their cache files.
- Right now mask files are always read even if the `masked_loss` parameter is not set, we should not be reading files that will not be used.

araleza:
I'm looking forward to this change, it should make training easier when you have training images with good foregrounds, but irrelevant hard-to-describe stuff in the background.

rockerBOO:
Tried this branch out. One note is I failed multiple times (after noticing the code) to be using the masks. I made a mistake calling the directory `masks` instead of `mask` (I thought the plural was the directory name). There was no indication that the masks were not being used or found. This can be very frustrating as others won't be able to notice that this is happening as it is silent completely if it's working or not. 

I want to suggest adding a check for the masks when making the Dataset subsets. Check each image if the mask file exists when making the subset. Then add the count to the dataset output. We have a "mask" in `ImageInfo` but we'd have to check if that is all ones? 

Sorry if i came across poorly. A little frustrated myself but wanted to save others the frustration in at least being able to confirm that the masks are working and how many are working. 

Thanks for adding this! :) I am looking forward to this. So much as to try it before it's released. :)

rockerBOO:
For example I made the following just to help me to confirm it was working. (I made a branch including my validation loss PR)

```python
        if args.masked_loss:
            masked = 0
            for b in train_dataloader:
                if torch.count_nonzero(b['masks']) != torch.numel(b['masks']):
                    masked += 1

            val_masked = 0
            for b in val_dataloader:
                if torch.count_nonzero(b['masks']) != torch.numel(b['masks']):
                    val_masked += 1

            accelerator.print("Using masked loss")
            accelerator.print(f"Masked: {masked}/{len(train_dataloader)}")
            accelerator.print(f"Validation masked: {val_masked}/{len(val_dataloader)}")
```

Prints out like this

```
Using masked loss
Masked: 228/241
Validation masked: 57/60
```

rockerBOO:
Been using this a lot this weekend and had an idea for a "masked min". This would set the masked value to a minimum value so the background can be masked freely to extract the subject but we can set a minimum value to apply. 

```
# Assuming new value `--masked_min=0.2` up to say 0.5 seems pretty good. 
if args.masked_min and torch.count_nonzero(mask) != torch.numel(noise_pred):
    minimum = torch.tensor([args.masked_min]).to(mask.device)
    minimum_mask = torch.gt(minimum, mask)
    mask = mask + (minimum * minimum_mask)
```

I'm still a little rough around the code here but hopefully it's enough to express what I mean. This helps to reduce a certain haloing effect that the trainings get around the masked subjects. It also helps that we don't need to make specific masks to enforce these minimums.

araleza:
Yeah, I've been thinking about how to keep some background too, @rockerBOO.  I think some people won't realise that you have to keep some background, or reflective parts of your character (including just the ambient color of the light in the scene) won't be learned correctly, and the resulting learned subject won't seem to fit properly into scenes that you make with the model.

Even just the scale of the foreground object needs some background for context for the learning to work at its best.

IdiotSandwichTheThird:
> Been using this a lot this weekend and had an idea for a "masked min". This would set the masked value to a minimum value so the background can be masked freely to extract the subject but we can set a minimum value to apply.

This isn't any different from manually adjusting black to some value of gray in the mask itself, right? Just wondering.

recris:
> Been using this a lot this weekend and had an idea for a "masked min". This would set the masked value to a minimum value so the background can be masked freely to extract the subject but we can set a minimum value to apply.
> 
> ```python
> # Assuming new value `--masked_min=0.2` up to say 0.5 seems pretty good. 
> if args.masked_min and torch.count_nonzero(mask) != torch.numel(noise_pred):
>     minimum = torch.tensor([args.masked_min]).to(mask.device)
>     minimum_mask = torch.gt(minimum, mask)
>     mask = mask + (minimum * minimum_mask)
> ```
> 
> I'm still a little rough around the code here but hopefully it's enough to express what I mean. This helps to reduce a certain haloing effect that the trainings get around the masked subjects. It also helps that we don't need to make specific masks to enforce these minimums.
> 
> I also have this little bit of code to see the masks
> 
> ```python
> def save_mask(mask, tag):
>     import torchvision.transforms as T
>     image = T.ToPILImage()(mask)
>     image.save(f'{tag}.mask.png')
> ```

I recommend doing this kind of mask processing outside the training code, it is easy to do with a small python script. If we are going to add a new parameter every time we need a new operation the list would grow indefinitely.

recris:
> Yeah, I've been thinking about how to keep some background too, @rockerBOO. I think some people won't realise that you have to keep some background, or reflective parts of your character (including just the ambient color of the light in the scene) won't be learned correctly, and the resulting learned subject won't seem to fit properly into scenes that you make with the model.
> 
> Even just the scale of the foreground object needs some background for context for the learning to work at its best.

I've been testing masks will an all black background and it seems to be working fine. However I do provide lighting "cues" in my captions which is probably helping.

rockerBOO:
> I recommend doing this kind of mask processing outside the training code, it is easy to do with a small python script. If we are going to add a new parameter every time we need a new operation the list would grow indefinitely.

I do agree with this approach but in practice it's prohibitively complicated. 

We'd make masks for the things we want but then would require creating new masks for the different minimums. Which requires some modifications of directories or putting originals elsewhere. It gets more complicated if you have many subset directories and need to modify each one separately. It can also cause false positive not knowing if you did all the masks correctly to the same value for testing purposes. I use upwards of 30 different dataset subsets which can make the process of individually modifying all the masks much more complicated and error prone. Scripting against my dataset_config toml is an option though but still need to manage different minimum masks for each subset. 

I believe Kohya supports many options already for different things to allow a lot of experimentation to happen. If we happen along a path of indefinite growth of arguments the option of specific masked arguments `--mask_args` could be an option to allow the flexibility to happen more specifically in the masked direction. I do not think we should limit the options in these cases if they have good value.

In this case I'm recommending 1 option that makes a clear testable, reproducible result that reduces complication, false positives, and other mistakes that can confuse the trainer. Provided the code works properly of course. 

Ultimately it's a suggestion so I'd be down with whatever is chosen. Thanks.

recris:
I understand your concern, in fact it is something I also struggled with... that is why I am building a set of tools to make preparing and modifying training data very easy. This is something I am planning to open-source soon.

twister77:
@recris 

Am i wrong or not, in a Runpot notebook my dataset is organized like this
`/workspace/dataset
01.jpg
02.jpg`

Does that mean that my masks should be organized like:
`/workspace/dataset/mask
01.png
02.png`

recris:
> @recris
> 
> Am i wrong or not, in a Runpod notebook my dataset is organized like this **/workspace/dataset** 01.jpg 02.jpg
> 
> Does that mean that my masks should be organized like: **/workspace/dataset/mask** 01.png 02.png

correct

araleza:
> I do agree with this approach but in practice it's prohibitively complicated.

Hello, I thought I'd throw in my opinion here.  Which is easy for me to do, as I'm not the person doing any of the work.  And I really appreciate that you people are actually doing the work here, so please don't mind me saying stuff here.

I also think that the masks subdirectory seems like it'll be a burden to work with.  I do realize that the idea of 'use the alpha channel of the training image' had difficulties when it came to whether premultiplied alpha was being used in the source .png files or not.

But, premultiplied alpha does not affect any pixel that has an alpha of 0.0 or an alpha of 1.0.  I doubt many people will care about alpha values that are not 0 or 1 - they mostly want to just delete noisy objects from their training images.

If you just assumed either premult alpha was present in the .png files, or not present, either way would be good.  Then a command line parameter could be added to override that default if that was actually important to them, but I imagine no-one would ever care enough to use that flag.

Thank you for listening to my suggestion.  And thank you for the work you've all done so far.

araleza:
Great to see this moving ahead, I feel it's a valuable feature for the training scripts. Thanks for getting this working. :)

Do you know if it is difficult to port it to sdxl_train.py as well, for fine tuning instead of generating a LoRA?

jordoh:
> I've update the branch to support masked loss with `sdxl_train.py`, but not tested yet.

I ran a test test with SDXL, but the fine-tuned model produces (fully) black images. I'll try the same config without masking today to see if this is an issue with my config, but posting as a heads up in the meantime.

Using the follow command line (formatted for easier reading):
```
accelerate launch --num_cpu_threads_per_process=2
    ./sdxl_train.py
    --enable_bucket --min_bucket_reso=256 --max_bucket_reso=1024
    --pretrained_model_name_or_path="C:\Users\demo\ComfyUI\models\checkpoints\juggernautXL_v9Rundiffusionphoto2.safetensors"
    --dataset_config="C:\Users\demo\dataset.toml"
    --caption_extension=".txt"
    --resolution="1024,1024"
    --output_dir="C:\Users\demo\output" --logging_dir="C:\Users\demo\log"
    --save_model_as=safetensors --full_bf16 --output_name="hta-juggernautxl9-3"
    --max_token_length=225 --max_data_loader_n_workers="0"
    --learning_rate_te1="1e-05" --learning_rate_te2="1e-05" --learning_rate="1.0"
    --lr_scheduler="cosine" --lr_scheduler_num_cycles="10"
    --train_batch_size="1" --max_train_steps="6600"
    --save_every_n_epochs="1"
    --mixed_precision="bf16" --save_precision="bf16"
    --optimizer_type="Adafactor" --optimizer_args scale_parameter=False relative_step=False warmup_init=False weight_decay=0.01
    --min_snr_gamma=5 --noise_offset=0.0
    --log_with wandb --wandb_api_key="..."
    --gradient_checkpointing --cache_latents --cache_latents_to_disk
    --masked_loss
```
and `dataset.toml`:
```
[general]
batch_size = 1
enable_bucket = true
bucket_reso_steps = 64


[[datasets]]
resolution = [1024, 1024]
min_bucket_reso = 512
max_bucket_reso = 1024

	[[datasets.subsets]]
	image_dir = "C:\\Users\\demo\\img\\10_hta woman"
	caption_extension = ".txt"
	conditioning_data_dir = "C:\\Users\\demo\\mask"
	num_repeats = 10
```
It could be that I've mis-configured the dataset.toml, as this is the first time I've used one, but training did appear to be proceeding as expected, with W&B showing a reasonable average loss at 1k steps, though there was a very high initial spike that I haven't seen before.

jordoh:
Testing today, I spotted the mix of Adafactor/DAdaptation settings I had in above command line (`learning_rate`, in particular). 🤦 

SDXL support appears to be working well - training still running now but early checkpoints clearly showing learning of masked areas only. 👍

araleza:
Hi, I'm testing it out now for SDXL fine tuning.

One thing I think might help people get used to the feature is adding something to the debug output. Maybe an INFO line of 'no mask images found' or '<x> mask images found', maybe next to the line about there being no regularization images found.

Nice work getting it in. And that future ControlNet training sounds interesting too.

araleza:
So I gave this code a much stronger test today, and placed specific objects in a series of 10 test images, and then masked them out. And I found that they were being learned, so my masks were not working.

This was because I was not passing --masked_loss as a parameter to sdxl_train.py.  But when I added this parameter, I just got an error message that didn't explain what to do:

```
  File "/home/zmx/m.2/Dev/sdxl/sd-scripts_mask/./sdxl_train.py", line 810, in <module>
    train(args)
  File "/home/zmx/m.2/Dev/sdxl/sd-scripts_mask/./sdxl_train.py", line 586, in train
    mask_image = batch["conditioning_images"].to(dtype=weight_dtype)[:, 0].unsqueeze(1)  # use R channel
KeyError: 'conditioning_images'
```
I then read @jordoh 's setup, above. It turns out that if you want to use the masked images, it seems that you _must_ use a .toml file to set up the conditioning_data_dir, which should be set to the mask image directory. With that in place, the specific objects I had masked out stopped showing up in my sample output, so the masks do work.

I don't generally use .toml files, so would it be possible to add a command line option, --conditioning_data_dir, to achieve the same result? Or, just look in the 'mask' subdirectory by default for the masks? That's how I thought it worked when I was doing training yesterday, but my masks weren't being used at all. It's easy to talk yourself into believing that they're having an effect, even when they're not.  I really do think some debug INFO lines are important to clarify if masks have been found or not.

araleza:
After that, I tried copying that .toml from my directory with 10 test images to my real training dataset.  I updated the paths inside the .toml for the images and masks, but when I run sdxl_train.py, I just get:
```
  File "/home/zmx/m.2/Dev/sdxl/sd-scripts_mask/./sdxl_train.py", line 810, in <module>
    train(args)
  File "/home/zmx/m.2/Dev/sdxl/sd-scripts_mask/./sdxl_train.py", line 166, in train
    blueprint = blueprint_generator.generate(user_config, args, tokenizer=[tokenizer1, tokenizer2])
  File "/home/zmx/m.2/Dev/sdxl/sd-scripts_mask/library/config_util.py", line 395, in generate
    sanitized_user_config = self.sanitizer.sanitize_user_config(user_config)
  File "/home/zmx/m.2/Dev/sdxl/sd-scripts_mask/library/config_util.py", line 358, in sanitize_user_config
    return self.user_config_validator(user_config)
  File "/home/zmx/m.2/Dev/sdxl/sd-scripts_mask/venv/lib/python3.10/site-packages/voluptuous/schema_builder.py", line 272, in __call__
    return self._compiled([], data)
  File "/home/zmx/m.2/Dev/sdxl/sd-scripts_mask/venv/lib/python3.10/site-packages/voluptuous/schema_builder.py", line 595, in validate_dict
    return base_validate(path, iteritems(data), out)
  File "/home/zmx/m.2/Dev/sdxl/sd-scripts_mask/venv/lib/python3.10/site-packages/voluptuous/schema_builder.py", line 433, in validate_mapping
    raise er.MultipleInvalid(errors)
voluptuous.error.MultipleInvalid: extra keys not allowed @ data['datasets'][0]['subsets'][0]['conditioning_data_dir']
```
I have no idea what this error means.  It goes away if I delete the conditioning_data_dir = "[...]" bit from my .toml file, but then the masks won't work of course.

Any thoughts what the cause / fix for this error message is?

cheald:
Can you show your dataset toml?

Here's an example of what's working for me:

```
# dataset.toml
[[datasets]]
  [[datasets.subsets]]
  num_repeats = 2
  caption_extension = ".txt"
  image_dir = "/path/to/images/subdir1"
  conditioning_data_dir = "/path/to/images/subdir1/masks"

  [[datasets.subsets]]
  num_repeats = 4
  caption_extension = ".txt"
  image_dir = "/path/to/images/subdir2"
  conditioning_data_dir = "/path/to/images/subdir2/masks"
```

You could then add `--dataset_config /path/to/dataset.toml` to the "Additional Parameters" field in the advanced tab.

araleza:
Thanks for asking, @cheald .  Here's my dataset.toml:
```
[[datasets]]
 
resolution = [1024, 1024]
min_bucket_reso = 32
max_bucket_reso = 1024

	[[datasets.subsets]]
	image_dir = "/home/zmx/m.2/Dev/sdxl/training/terra/kohya/img/1_terra planet/"
	caption_extension = ".txt"
	conditioning_data_dir = "/home/zmx/m.2/Dev/sdxl/training/terra/kohya/img/1_terra planet/mask/"
	num_repeats = 1
```
I've tried it both with and without the trailing slashes on the ends of the paths. Also, I've copied and pasted those directories into Ubuntu Dolphin, to check they definitely exist, and they do.

This same .toml worked for my test dataset of 10 images with 10 mask images.  It just doesn't work for my actual dataset, which has far more images.  I don't have masks for all of them yet, but that's not supposed to be necessary.

cheald:
If it's not expecting `conditioning_data_dir` then that'd suggest to me that you don't have `--masked_loss` set. Are you sure you set that for your real data run? ConfigSanitizer only accepts `conditioning_data_dir` when `support_controlnet` is set, and during lora training that's set by the `args.masked_loss` flag.

The trainer will expect masks for each of your inputs (and no extras), but you'll get an error message about that once you get past the schema checks.

araleza:
> If it's not expecting conditioning_data_dir then that'd suggest to me that you don't have --masked_loss set.

Aha yep that's it. I had that --masked_loss parameter set for my 10 image test, but I forgot to copy it across to my real training set. Thanks, @cheald .

It might be worth upgrading that error message, or having --masked_loss set implicitly when people pass in a conditioning_data_dir.

araleza:
I found an issue that if the training image is a .jpg file, it seems that the mask image also has to be a .jpg, which is not likely to be something you want to do.

You get this error message if you have a .jpg for the training image and a .png for the mask image:
```
AssertionError: missing conditioning data for 1 images: ['terra_landscape_1.jpg']
```

CerVieel:
Is it possible to use masked loss with a regularization subset? Ideally masks for regularization images also, but for now I'd be happy if it worked just for the dataset when there are also regularization images.

Calling the test script (from a notebook):
```
!accelerate launch --num_cpu_threads_per_process=2 "sd-scripts/train_network.py" \
    --dataset_config=$dataset"/dataset.toml" \
    --pretrained_model_name_or_path=$models"/dreamshaper_8.safetensors" \
    --resolution=768,768 \
    --output_dir=$output"/model" \
    --logging_dir=$output"/log" \
    --save_model_as=safetensors \
    --network_module=networks.lora \
    --xformers \
    --output_name="msk00" \
    --optimizer_type="AdamW8bit" \
    --lr_scheduler="constant" \
    --masked_loss \
    --seed=4321
```

Toml:
```
[[datasets]]
  [[datasets.subsets]]
  num_repeats = 2
  caption_extension = ".txt"
  image_dir = "/home/studio-lab-user/sagemaker-studiolab-notebooks/datasets/test_img"
  conditioning_data_dir = "/home/studio-lab-user/sagemaker-studiolab-notebooks/datasets/test_img/mask"
  
  [[datasets.subsets]]
  num_repeats = 1
  caption_extension = ".txt"
  is_reg = true
  image_dir = "/home/studio-lab-user/sagemaker-studiolab-notebooks/datasets/test_reg"
  conditioning_data_dir = "/home/studio-lab-user/sagemaker-studiolab-notebooks/datasets/test_reg/mask"
```


When the reg subset has a `conditioning_data_dir`:
```
raise er.MultipleInvalid(errors)
voluptuous.error.MultipleInvalid: extra keys not allowed @ data['datasets'][0]['subsets'][1]['is_reg']
```

When `conditioning_data_dir` is removed from reg:
```
raise er.MultipleInvalid(errors)
voluptuous.error.MultipleInvalid: extra keys not allowed @ data['datasets'][0]['subsets'][0]['conditioning_data_dir']
```
It runs when the entire reg subset is removed.

recris:
I've done some tests using the `dev` branch, and it seems to be working well for me, at least for SDXL LoRA training.

I will close this PR once the feature lands in `main` branch.

Serallan:
> Unfortunately the dataset with `conditioning_data_dir` doesn't support `is_reg`. Please set `num_repeats` of each dataset so that 'number of images * num_repeats' are the same. This is very similar to use reg images.

Thanks for clearing this up! Once I tried to use reg images as training images and results were pretty different from a run with them as reg, but I didn't alter repeats. When you mean to set repeats so they're the same you mean that if I originally have something like:

150 images in 2_trainingImages  (150 * 2 = 300)
50 images in 1_regImages (50 * 1 = 50)

To achieve the closest result (and be able to use masks!) then I must turn them into:
2_trainingImages (150 * 2 = 300)
6_regImages (now as training imgs) (50 * 6 = 300)
?

Dumb example with weird unnecessary repeats I know, I just want to make sure I understood it correctly.

PR #604, opened 2023-06-21
Multi-GPU (DDP and deepspeed) fixes

jamarju-fp:
- Fixed DDP bug where forward/backward are called on the *unwrapped* modules. As a result gradient synchronization won't happen, everything may look okay but you will actually get diverging (and undertrained) models on each GPU.
 See: https://github.com/kohya-ss/sd-scripts/blob/c7fd336c5d5c0f05378dd34e7b5ca6ac8e972773/fine_tune.py#L243

- `accelerator.save_state` needs to be called from all processes so that rank-specific data is saved, fixed.

- Also implemented multi-gpu inference, a new --sample_before_training flag and upgraded bitsandbytes to 0.39 for Lion8bit, Ada + Hopper, etc.

- Note only `fine_tune.py` was fixed, no LoRA changes. Comments more than welcome.

### DIFF
--- a/fine_tune.py
+++ b/fine_tune.py
@@ -148,10 +148,10 @@ def train(args):
         train_util.replace_unet_modules(unet, args.mem_eff_attn, args.xformers)
 
     # 学習を準備する
+    vae.to(accelerator.device, dtype=weight_dtype)
+    vae.requires_grad_(False)
+    vae.eval()
     if cache_latents:
-        vae.to(accelerator.device, dtype=weight_dtype)
-        vae.requires_grad_(False)
-        vae.eval()
         with torch.no_grad():
             train_dataset_group.cache_latents(vae, args.vae_batch_size, args.cache_latents_to_disk, accelerator.is_main_process)
         vae.to("cpu")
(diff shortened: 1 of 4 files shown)


pekinghk:
> * Fixed DDP bug where forward/backward are called on the _unwrapped_ modules. As a result gradient synchronization won't happen, everything may look okay but you will actually get diverging (and undertrained) models on each GPU.
>   See: https://github.com/kohya-ss/sd-scripts/blob/c7fd336c5d5c0f05378dd34e7b5ca6ac8e972773/fine_tune.py#L243
> * `accelerator.save_state` needs to be called from all processes so that rank-specific data is saved, fixed.
> * Also implemented multi-gpu inference, a new --sample_before_training flag and upgraded bitsandbytes to 0.39 for Lion8bit, Ada + Hopper, etc.
> * Note only `fine_tune.py` was fixed, no LoRA changes. Comments more than welcome.

Why would we unwrap the model before saving it? Wouldn't this stop the gradient synchronization across multiple GPUs, essentially turning it into a single-GPU training? I don't think this is a good adjustment.

jamarju-fp:
> Why would you unwrap the model before training it? 
So that the PR blends in it with minimal differences (the original code unwraps the models)

> Wouldn't this stop the gradient synchronization across multiple GPUs, 
No, because both forward and backward passes are called on the accelerated (wrapped) model.

> essentially turning it into a single-GPU training?
This is exactly what this PR attempts to fix. You are actually training N separate models with 1/N the data without this.

PR #618, opened 2023-07-03
A script to revert the new format safetensors checkpoint (diffusers 0.17.) to the old format

Joanne0720:
I am also facing the issue about the Missing key(s) and Unexpected key(s) in state dict of VAE in diffusers 0.17.+ .
It happens when I'm trying to fine-tuning a checkpoint by LoRA, and checkpoint comes from the latest diffusers fine-tune script.
So I've written a script to revert the new format safetensors file to the old format to solve the problem. 
It works for me when loading the sd1.5-based models.

### DIFF
--- a/tools/revert_safetensors_to_old_format.py
+++ b/tools/revert_safetensors_to_old_format.py
@@ -1,0 +1,43 @@
+import argparse
+from safetensors.torch import load_file, save_file
+
+vae_conversion_map_attn = [
+    # (old format, new format)
+    (".q.", ".to_q."),
+    (".k.", ".to_k."),
+    (".v.", ".to_v."),
+    (".proj_out.", ".to_out.0."),
+]
+
+def convert_vae_attn_state_dict(state_dict):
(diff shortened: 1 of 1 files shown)


feffy380:
IMO for ease of use this would be better suited as an automatic conversion step when loading a checkpoint, similar to how diffusers 0.17 automatically converts to the new format.

shadowlocked:
I would really appreciate it if this problem, [which has impacted me lately](https://github.com/kohya-ss/sd-scripts/issues/723), could be resolved by integration into the core.

shadowlocked:
Is it possible to give some instructions for how to run this script? I don't know how to construct the arguments.

shadowlocked:
> If you can checkout sdxl branch, please use the branch for a while.

Ach, won;t be ready for SDXL for a while. I'll have to wait.

shadowlocked:
Is there someplace I can look to see when this particular issue will be fixed?

PR #621, opened 2023-07-05
20230705 add local model path

godlockin:
in case user download the model before using the scripts, add a param to define the local path for model files' loading


godlockin:
yeah, but each time to set the 'tokenizer_cache_dir' on the script run? the tokenizer (maybe), vocabulary and other odds and ends are small to download soon, but the base model for GB may not

godlockin:
emmm, seems like we'd set 2 params `--tokenizer_cache_dir=$local_path_to_tokenizer` and `--model_cache_dir=$local_path_to_model` which default to empty means to load/download the model remotely

Issue #651, opened 2023-07-16
Cache latents slow down after a certain number of steps have passed


aearone:
I don't know what happened, it's never happened before. I am trying to train the model at 576x576 resolution using regularization images (I have 1960 of them and 98 training images), but when I check the cache after 42 steps the speed drops a lot, i.e. it was 5-8s/it and becomes 1.4, cache read time is 1 hour. What can be the problem? I checked the images, they are fine, they all have the same format and size.

![Screenshot_50](https://github.com/kohya-ss/sd-scripts/assets/93353660/31461f3a-6986-4d02-b0fa-1ffeebf12a91)

aearone:
> @aearone Check your VRAM usage in task manager. The training/caching becomes very laggy if it starts to use shared memory. Also I believe there is a memory leak during latent caching stage. If it was working fine before on your device with that many images, you can test this PR #658 and check for any improvements,

During caching my VRAM is fully utilized (8gb out of 8), so probably there really is a memory leak or just not enough memory. As I said, if I use 512x512 training resolution and 512x512 regularization images, then caching happens quite quickly, in 5-7 minutes, without slowing down the speed (for 2000 images). And the same if I use training resolution 768x768 and regularization images 768x768, caching in this case also goes fast enough and there is no "stuck". Only if I use some non-standard values like 576x576. But I will consider that most likely it is a memory leak.

aearone:
> @aearone Check your VRAM usage in task manager. The training/caching becomes very laggy if it starts to use shared memory. Also I believe there is a memory leak during latent caching stage. If it was working fine before on your device with that many images, you can test this PR #658 and check for any improvements,

Just as I thought, same problem on a completely new operating system. Is it a problem on the script side or on my video card side (not enough video memory for caching), as the video card is 100% loaded and using 7.7 GB VRAM out of 8. It seems there is indeed a memory leak when the script is trying to use RAM to run, this is the reason why the caching speed is slowing down. 
![Снимок экрана 2023-07-18 052939](https://github.com/kohya-ss/sd-scripts/assets/93353660/7f897d1f-fa8b-459b-8645-1ee69b77c1f3)

PR #658, opened 2023-07-17
Cache latents VRAM leak fix

pamparamm:
Probably related to #651. I also had a problem with caching images from a small dataset (50-100 images) at 512 resolution, causing VRAM to exceed 24GB.

This should fix problems with vram leakage during "caching latents..." stage (tested on Windows).

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -2141,6 +2141,8 @@ def cache_batch_latents(
             info.latents = latent
             if flip_aug:
                 info.latents_flipped = flipped_latent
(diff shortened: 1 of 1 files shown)


ArrowM:
This fixed it for me. Thanks!

pamparamm:
@kohya-ss Thank's for merging. Fix still works correctly outside the for loop.
BTW, here's what it used to look like: (left spike is VRAM usage with fix applied, right without):
vram-leak

PR #666, opened 2023-07-20
Make things work for MPS.

hameerabbasi:


hameerabbasi:
Not much really, just had to get the device and use it instead of CUDA.

hameerabbasi:
The assignment is needed for all backends. This was a bug I fixed, otherwise it moves the device and discards the result instead of using it.

hameerabbasi:
I will adjust this PR soon. ;)

Issue #354, opened 2023-03-30
[Error] 'device cuda is invalid'


zakinp:
This error occur when using the lastest version of this scripts. The messages are:

Traceback (most recent call last):
  File "/root/autodl-tmp/sd-scripts/train_network.py", line 719, in <module>
    train(args)
  File "/root/autodl-tmp/sd-scripts/train_network.py", line 134, in train
    text_encoder, vae, unet, _ = train_util.load_target_model(args, weight_dtype, accelerator.device)
  File "/root/autodl-tmp/sd-scripts/library/train_util.py", line 2649, in load_target_model
    text_encoder, vae, unet = model_util.load_models_from_stable_diffusion_checkpoint(args.v2, name_or_path, device)
  File "/root/autodl-tmp/sd-scripts/library/model_util.py", line 869, in load_models_from_stable_diffusion_checkpoint
    _, state_dict = load_checkpoint_with_text_encoder_conversion(ckpt_path, device)
  File "/root/autodl-tmp/sd-scripts/library/model_util.py", line 844, in load_checkpoint_with_text_encoder_conversion
    state_dict = load_file(ckpt_path, device)
  File "/root/miniconda3/envs/kohya/lib/python3.10/site-packages/safetensors/torch.py", line 98, in load_file
    with safe_open(filename, framework="pt", device=device) as f:
Exception: device cuda is invalid
Traceback (most recent call last):
  File "/root/miniconda3/envs/kohya/bin/accelerate", line 8, in <module>
    sys.exit(main())
  File "/root/miniconda3/envs/kohya/lib/python3.10/site-packages/accelerate/commands/accelerate_cli.py", line 45, in main
    args.func(args)
  File "/root/miniconda3/envs/kohya/lib/python3.10/site-packages/accelerate/commands/launch.py", line 1104, in launch_command
    simple_launcher(args)
  File "/root/miniconda3/envs/kohya/lib/python3.10/site-packages/accelerate/commands/launch.py", line 567, in simple_launcher
    raise subprocess.CalledProcessError(returncode=process.returncode, cmd=cmd)
[...]

PR #676, opened 2023-07-23
Fix RAM leak when loading model in lowram device

Isotr0py:
- Fix RAM leak on low RAM device like colab when `lowram = true`
- Fix `safetensors_rust.SafetensorError: device cuda:0 is invalid`. Same to #354

For low-ram device like Colab,  the usual workflow to load U-net may blow the RAM since it costs about 12.7Gib RAM to load the checkpoint and initialize the `SdxlUNet2DConditionModel()`.

Using `init_empty_weights()` context manager can initialize a model without using any RAM. It can solve the RAM leak on low RAM device.

However, since we didn't use `load_state_dict()` when loading U-Net from checkpoint, `_IncompatibleKeys` will not be warned for corrupted checkpoint with `missing_keys` and `unexpected_keys`...

### DIFF
--- a/library/sdxl_model_util.py
+++ b/library/sdxl_model_util.py
@@ -141,7 +141,10 @@ def load_models_from_sdxl_checkpoint(model_version, ckpt_path, map_location):
     # Load the state dict
     if model_util.is_safetensors(ckpt_path):
         checkpoint = None
-        state_dict = load_file(ckpt_path, device=map_location)
+        try:
+            state_dict = load_file(ckpt_path, device=map_location)
+        except:
+            state_dict = load_file(ckpt_path)   # prevent device invalid Error
         epoch = None
         global_step = None
     else:
--- a/library/sdxl_model_util.py
+++ b/library/sdxl_model_util.py
@@ -1,4 +1,6 @@
 import torch
+from accelerate import init_empty_weights
+from accelerate.utils.modeling import set_module_tensor_to_device
 from safetensors.torch import load_file, save_file
 from transformers import CLIPTextModel, CLIPTextConfig, CLIPTextModelWithProjection, CLIPTokenizer
 from diffusers import AutoencoderKL, EulerDiscreteScheduler, StableDiffusionXLPipeline, UNet2DConditionModel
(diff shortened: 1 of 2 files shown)


Isotr0py:
```python
from accelerate import init_empty_weights, load_checkpoint_and_dispatch
from accelerate.utils.modeling import set_module_tensor_to_device

from diffusers import StableDiffusionXLPipeline
import torch
import gc

from library import model_util, sdxl_model_util, train_util, sdxl_original_unet


def load_target_model(device):
    pipe = StableDiffusionXLPipeline.from_pretrained(
        "stabilityai/stable-diffusion-xl-base-1.0", torch_dtype=torch.float16, variant="fp16", use_safetensors=True
    )

    text_encoder1 = pipe.text_encoder
    text_encoder2 = pipe.text_encoder_2
    vae = pipe.vae
    unet = pipe.unet
    del pipe
    gc.collect()

    print("convert U-Net to original U-Net")

    state_dict = sdxl_model_util.convert_diffusers_unet_state_dict_to_sdxl(unet.state_dict())

    # original u-net
    original_unet = sdxl_original_unet.SdxlUNet2DConditionModel()
    state_dict = sdxl_model_util.convert_diffusers_unet_state_dict_to_sdxl(unet.state_dict())
    original_unet.load_state_dict(state_dict)
    unet = original_unet
    
    # pr's u-net
    with init_empty_weights():
        unet = sdxl_original_unet.SdxlUNet2DConditionModel()
    for k in list(state_dict.keys()):
        set_module_tensor_to_device(unet, k, device, value=state_dict.pop(k))

    print("U-Net converted to original U-Net")
    return text_encoder1, text_encoder2, unet

text_encoder1, text_encoder2, unet = load_target_model("cpu")
unet.to("cuda")
text_encoder1.to("cuda")
text_encoder2.to("cuda")
```

I tested with the code upon. PR's and original method both cost about 12.2G VRAM (10.6G for U-net only) to load unet and text_encoders.

It seems that VRAM usage didn't change in model loading. I wonder OOM occurs in which step to figure out what is going wrong.

Isotr0py:
BTW, for `set_module_tensor_to_device`, if we pass `dtype=torch.float16` or `dtype=torch.bfloat16`, the model will directly load to `fp16/bf16` instead of `fp32` than `fp16/bf16`.

Maybe we can integrate `full_fp16/full_bf16` into model loading to reduce VRAM peak.🤔

Isotr0py:
Well, `_load_state_dict` would work like `model.load_state_dict()` with `strict=True` to find out wrong keys and raise an error before loading.

It seems that it did work for a corrupted SDXL-1.0 checkpoint created by adding/deleting some keys.

PR #687, opened 2023-07-28
[Fix text_model confinement][Unexpected key(s) in state_dict]

tomguluson92:
Fix `Unexpected key(s) in state_dict: "text_model.embeddings.position_ids".` problem aroused by loading **text_model**.

### DIFF
--- a/library/model_util.py
+++ b/library/model_util.py
@@ -896,7 +896,7 @@ def load_models_from_stable_diffusion_checkpoint(v2, ckpt_path, device="cpu", dt
             transformers_version="4.25.0.dev0",
         )
         text_model = CLIPTextModel._from_config(cfg)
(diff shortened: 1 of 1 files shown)


tomguluson92:
> Thank you for this! But I'd like to keep `strict=True` to find unexpected errors. If some error occurs in loading the state dict, I will add a conversion code for it. Could you please tell me the location (url) of the checkpoint which causes the error?

Thanks, This is only for info `text_model`, which is not used in real training. The base model I found not compatible is [civitai-awportrait](https://civitai.com/models/61170/awportrait)

vladmandic:
btw, this is also happening if using `transformers` any version higher than `4.30.2`.

QianquanChina:
When I upgraded `transformers==4.36.2` this error disappeared

PR #755, opened 2023-08-13
add lora_fa

kohya-ss:


feffy380:
Is that system memory or vram?

evshiron:
@kohya-ss 

Greetings. Do you have an ETA for merging the dev branch to main? A PR https://github.com/vladmandic/automatic/pull/2009 is currently waiting for it.

feffy380:
I saw vram usage increase from 10.7GB to 11.3GB. Speed increase was only about 3-4%. It doesn't seem to be worth using in the current state

feffy380:
@kohya-ss I printed out the `network.parameters()` right before the training loop and it looks like the lora_down modules aren't getting frozen (True is output of requires_grad)
image

Additionally, `LoRAModule.requires_grad_()` never gets called. The `network.requires_grad_()` call doesn't seem to propagate to it.

PR #760, opened 2023-08-15
Update resize_lora.py

Symbiomatrix:
I was [referred by bmaltais](https://github.com/bmaltais/kohya_ss/pull/1393) here. 
This is a (partial?) bugfix for when lora layers are period delimited instead of some other char; split gets only a fraction of the block and errs when the modified block name is not found. Proposed rsplit method splits exactly according to the location of the lora_down and last period (after which is presumed to appear weight name). It might be preferable instead to just change lora_down to lora_up via a substring replace call.
I see the code hasn't been updated in 3 months, so I guess it's well overdue for an overhaul in general, supporting lycoris and nonstandard layers.


Symbiomatrix:
Certainly, thanks. I still have no idea where that model could've come from, just wanted to share a small nonintrusive patch on my end.

Issue #788, opened 2023-08-25
Are there any way to reduce system RAM usage? I am trying to make a tutorial for free colab SDXL LoRA training 


FurkanGozukara:
Hello. I am trying to make a tutorial for free colab

It has good GPU but the problem is system ram is only 13 GB

We need some options or way to reduce RAM usage

FurkanGozukara:
Kaggle provides 2 GPU

can we load unet and text encoder into one of the GPUs with --lowram option?

![image](https://github.com/kohya-ss/sd-scripts/assets/19240467/8337a6ac-d885-43b8-941d-67964c94c288)


@kohya-ss

FurkanGozukara:
```
loading U-Net from checkpoint
IOStream.flush timed out
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ /opt/conda/bin/accelerate:8 in <module>                                      │
│                                                                              │
│   5 from accelerate.commands.accelerate_cli import main                      │
│   6 if __name__ == '__main__':                                               │
│   7 │   sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])     │
│ ❱ 8 │   sys.exit(main())                                                     │
│   9                                                                          │
│                                                                              │
│ /opt/conda/lib/python3.10/site-packages/accelerate/commands/accelerate_cli.p │
│ y:45 in main                                                                 │
│                                                                              │
│   42 │   │   exit(1)                                                         │
│   43 │                                                                       │
│   44 │   # Run                                                               │
│ ❱ 45 │   args.func(args)                                                     │
│   46                                                                         │
│   47                                                                         │
│   48 if __name__ == "__main__":                                              │
│                                                                              │
│ /opt/conda/lib/python3.10/site-packages/accelerate/commands/launch.py:909 in │
│ launch_command                                                               │
│                                                                              │
│   906 │   elif args.use_megatron_lm and not args.cpu:                        │
│   907 │   │   multi_gpu_launcher(args)                                       │
│   908 │   elif args.multi_gpu and not args.cpu:                              │
│ ❱ 909 │   │   multi_gpu_launcher(args)                                       │
│   910 │   elif args.tpu and not args.cpu:                                    │
│   911 │   │   if args.tpu_use_cluster:                                       │
│   912 │   │   │   tpu_pod_launcher(args)                                     │
│                                                                              │
[...]

PR #791, opened 2023-08-27
Intergrate fp16/bf16 support to sdxl model loading

Isotr0py:
Maybe related issue: #788 
- Specify fp16/bf16 dtype when loading model state_dict

This should reduce VRAM usage peak when enable `--full_fp16`/`--full_bf16` and `--lowram` at the same time during loading model.

It also reduced RAM usage when loading checkpoint from `safetensors` format.


Isotr0py:
Yes, `dtype` is not passed to `_load_state_dict_on_device` for `text_encoders`. So `text_encoders` will always load as fp32, no matter whether full_fp16/bf16 is enabled or not.

FurkanGozukara:
could this have introduced a major bug?

My SDXL LoRAs are super trained now with same settings compared to before this change

Actually I am testing right now effect of this option

PR #798, opened 2023-09-02
add input perturbation noise

vvern999:
Method from https://arxiv.org/abs/2301.11706
Recommended value is 0.1. Values >0.2 are too high. 
Example: https://files.catbox.moe/okgry9.jpg

Very simple, but no one has added this yet.


sanshibayuan:
What's the difference between input perturbation noise and offset noise? The inplementation seems the same to me.

sanshibayuan:
For those who also wondered, I found the explanation from the author: https://github.com/forever208/DDPM-IP/issues/7#issuecomment-1563108075

> What's the difference between input perturbation noise and offset noise? The inplementation seems the same to me.

PR #823, opened 2023-09-16
update bitsandbytes for 0.41.1 and fixed bugs with generate_controlnet_subsets_config for training

sdbds:


sdbds:
> Thank you for this!
> 
> However, in sdxl (and dev) branch, `bitsandbytes` is now optional, and the user can choose the version of `bitsandbytes` from 0.35.0 and 0.41.1. In addition to that, if the user uses 0.41.1, I recommend to use the pre-built wheel form [jllllll](https://github.com/jllllll). So I think the dlls and main.py should be kept as is.
> 
> If it is fine, could you please make this only for the generate_controlnet_subsets_config issue?

OK，i revert it and just fix config issue

sdbds:
> Thank you for updating, it is much appreciated. I will review the PR sooner.
> 
> I will be happy if you provide some description of the root issue when you have time.

controlnet dataset use old code so it judge reg and happend errors.

i changed it then use new dirs formatting with toml datasets(without num repeats)

PR #825, opened 2023-09-19
Intel ARC support with IPEX

Disty0:
Initial Intel ARC GPU support with IPEX on Linux / WSL.
This PR presents IPEX as CUDA.

Users needs to set these manually:
Mixed precision to BF16,
Attention to SDPA,
Optimizer to: AdamW (or any other non 8 bit one).

### DIFF
--- a/XTI_hijack.py
+++ b/XTI_hijack.py
@@ -1,4 +1,11 @@
 import torch
+try:
+    import intel_extension_for_pytorch as ipex
+    if torch.xpu.is_available():
+        from library.ipex import ipex_init
+        ipex_init()
+except Exception:
+    pass
 from typing import Union, List, Optional, Dict, Any, Tuple
 from diffusers.models.unet_2d_condition import UNet2DConditionOutput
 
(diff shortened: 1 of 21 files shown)


Disty0:
I can maintain IPEX side. You can tag me with IPEX specific issues. That's what other repos do most of the time.

For new scripts, adding these lines after importing torch should enable IPEX support and you can continue with using torch.cuda:

```
try:
    import intel_extension_for_pytorch as ipex
    if torch.xpu.is_available():
        from library.ipex import ipex_init
        ipex_init()
except Exception:
    pass
```

I went with converting torch.xpu to torch.cuda route because using torch.xpu needs a lot of maintenance.
I also manage my IPEX library externally here:
https://github.com/Disty0/ipex_to_cuda

FurkanGozukara:
when installing this via [bmaltais](https://github.com/bmaltais) GUI don't select this option if you don't have it and know what you are doing

I have 13900 K and it caused me to not to be able to do any training

Disty0:
Because it's trying to use the iGPU and IPEX currently doesn't work with iGPUs.

PR #832, opened 2023-09-24
Sdxl branch: Added setting to save checkpoint only after X number of steps

yushan777:
Added setting `save_every_n_steps_after_x` to save checkpoint only after X number of steps (if `save_every_n_steps` is used).  Default=0


araleza:
Perhaps a 'save every x minutes' box could be useful, as the time per step depends on the batch size, and calculating how many steps you need to save roughly every half an hour is a pain.

yushan777:
> Thank you for this! This seems to be useful and intuitive. However I think this option could be substituted with the functionality of the option `--save_last_n_steps`. What do you think about that?

Sorry for the late reply.  I was not familiar with `--save_last_n_steps` at the time.   So if I am correct, 

if i have max steps = 3000, but only want to save from say, 1000 onwards, i would use --save_last_n_steps=2000?

PR #858, opened 2023-10-05
Add append_captions feature to wd14 tagger

a-l-e-x-d-s-9:
This feature allows for appending new tags to the existing content of caption files.
If the caption file for an image already exists, the tags generated from the current
run are appended to the existing ones. Duplicate tags are checked and avoided.


a-l-e-x-d-s-9:
Thank you.

PR #859, opened 2023-10-05
Update train_network.py

DevArqSangoi:
Hello,

I've noticed that the `neuron_dropout` argument in the function call to `create_network` is not being correctly mapped to the underlying Lycoris library. In the Lycoris implementation, the argument for specifying dropout is called `dropout`, not `neuron_dropout`.

Because of this discrepancy, the user-defined value for dropout is not being effectively passed to the Lycoris library, resulting in the library defaulting to 0.0 for dropout even when the user specifies a different value.

To fix this issue, I've updated the argument name from `neuron_dropout` to `dropout` to ensure proper mapping and to allow the user-defined dropout value to take effect.

Thank you for considering this change.

### DIFF
--- a/train_network.py
+++ b/train_network.py
@@ -291,7 +291,7 @@ class NetworkTrainer:
                 vae,
                 text_encoder,
                 unet,
-                neuron_dropout=args.network_dropout,
+                dropout=args.network_dropout,
                 **net_kwargs,
             )
         if network is None:


DevArqSangoi:
> Unfortunately, this PR will stop networks.lora from working. In principle, it would be appropriate to request an update from LyCORIS repository. However, it is possible to add a workaround that supports LyCORIS, so I will do in this case.

Are you sure? I made this modification locally and it's working fine. I only submitted the PR after testing it. 👀

DevArqSangoi:
> Because `create_network` in lora.py doesn't have `dropout` argument.
> 
> https://github.com/kohya-ss/sd-scripts/blob/0faa350175cc9f6d848c5bbce3fabde5ba02fe36/networks/lora.py#L410
> 
> I've just updated and I believe LyCORIS works with dropout now.

Ah, I see the confusion here. The issue is that the `create_network` called in LyCORIS doesn't come from `lora.py`; it comes from the Kohya/LyCORIS library.

I don't know how to quote here, but it's on line 44 of 'C:\kohya_ss\venv\Lib\site-packages\lycoris\kohya\__init__.py'.

DevArqSangoi:
My dropout was already working with the argument change, but, ok 🤷‍♂️ thanks, I think.

Issue #854, opened 2023-10-04
I get an error when trying to learn using google colab


StupidGame:
I get an error when trying to learn using google colab.
Error Description:

```Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/transformers/utils/import_utils.py", line 1086, in _get_module
    return importlib.import_module("." + module_name, self.__name__)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/usr/local/lib/python3.10/dist-packages/transformers/models/clip/image_processing_clip.py", line 22, in <module>
    from ...image_transforms import (
  File "/usr/local/lib/python3.10/dist-packages/transformers/image_transforms.py", line 48, in <module>
    import tensorflow as tf
  File "/usr/local/lib/python3.10/dist-packages/tensorflow/__init__.py", line 38, in <module>
    from tensorflow.python.tools import module_util as _module_util
  File "/usr/local/lib/python3.10/dist-packages/tensorflow/python/__init__.py", line 37, in <module>
    from tensorflow.python.eager import context
  File "/usr/local/lib/python3.10/dist-packages/tensorflow/python/eager/context.py", line 29, in <module>
    from tensorflow.core.framework import function_pb2
  File "/usr/local/lib/python3.10/dist-packages/tensorflow/core/framework/function_pb2.py", line 5, in <module>
    from google.protobuf.internal import builder as _builder
ImportError: cannot import name 'builder' from 'google.protobuf.internal' (/usr/local/lib/python3.10/dist-packages/google/protobuf/internal/__init__.py)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/diffusers/utils/import_utils.py", line 684, in _get_module
    return importlib.import_module("." + module_name, self.__name__)
[...]

PR #864, opened 2023-10-08
Add `--onnx` to wd14 tagger

Isotr0py:
Related issue: #854

- Add `--onnx` for wd14 tagger to run onnx model
- Make `tensorflow` optional for wd14 tagger

Though the `onnx` model in [SmilingWolf/wd-v1-4-convnext-tagger-v2](https://huggingface.co/SmilingWolf/wd-v1-4-convnext-tagger-v2) can only use `batch_size=1`, it still has a faster tagging speed than `keras`.

### DIFF
--- a/finetune/tag_images_by_wd14_tagger.py
+++ b/finetune/tag_images_by_wd14_tagger.py
@@ -2,15 +2,14 @@ import argparse
 import csv
 import glob
 import os
-
-from PIL import Image
-import cv2
-from tqdm import tqdm
-import numpy as np
-from tensorflow.keras.models import load_model
-from huggingface_hub import hf_hub_download
-import torch
 from pathlib import Path
+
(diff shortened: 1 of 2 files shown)


Isotr0py:
Yes, changing the batch size of the model on the fly will cause wrong results, because we may break the model graph on the fly.
For other tagger, change the batch size of the model on the fly will fail because they have different calculation graph.

It seems that the dynamic shape onnx model exported from original tensorflow model using [onnx/tensorflow-onnx](https://github.com/onnx/tensorflow-onnx) can work well:
```
python -m tf2onnx.convert --saved-model ./wd-v1-4-convnext-tagger-v2/ --output convnext-tagger-v2.onnx --opset 13 --verbose --inputs input_1:0[-1,448,448,3]
```

Issue #890, opened 2023-10-20
--train_text_encoder in sdxl_train.py does literally NOTHING


araleza:
Hi!

I've been trying to perform Dreambooth training of the SDXL text encoders without affecting the unet at all.  There doesn't seem to be an option in sdxl_train.py to specifically target only the text encoder, so I've achieved that by using these options:

`--learning_rate="0.001"
`
`--train_text_encoder --block_lr 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
`

Watching the Python code execution, all looks well.  Just the text encoder parameters get marked for training, the two text models are set to train=true, the training rate of 0.001 is attached to the text parsers, and loss is generated and backpropagated.

The only problem is that _nothing happens_.  No matter what learning rate I set the network to (even crazy high values like --learning_rate="0.001", such as in my example here to make it obvious if any training is occurring), the sample output images remain exactly the same quality as if no training is being performed.  They neither get worse, nor better.

I can train only the text encoder without a problem when I'm making a LoRA, but it just seems non-functional for Dreambooth training.  I can easily see the text encoder changing the sample output images as my LoRA trains.  But it doesn't work for Dreambooth.

Here's my full command, in case there is anything interesting in there:

[...]

PR #895, opened 2023-10-24
Better implementation for te autocast

KohakuBlueleaf:
When we disable the training for TE, we will not prepare it and in this case we will need explicit convert it to the target dtype (or it will remain in fp32 which may not the expected behavior)

So basically I do 2 things here:
1, explicitly convert the TE to target dtype/device when it is not be trained
2, explicitly add autocast for TE since sometime it may not be prepare. (or definitely will not be prepared eg: cached TE)

### DIFF
--- a/sdxl_train_network.py
+++ b/sdxl_train_network.py
@@ -97,7 +97,8 @@ class SdxlNetworkTrainer(train_network.NetworkTrainer):
         if "text_encoder_outputs1_list" not in batch or batch["text_encoder_outputs1_list"] is None:
             input_ids1 = batch["input_ids"]
             input_ids2 = batch["input_ids2"]
-            with torch.enable_grad():
+            # Should not always enable grad
+            with (torch.enable_grad() if self.is_train_text_encoder(args) else torch.no_grad()):
                 # Get the text embedding for conditioning
                 # TODO support weighted captions
                 # if args.weighted_captions:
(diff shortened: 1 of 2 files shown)


KohakuBlueleaf:
> Thank you for this! It seems to be very good. I will review and merge sooner.
> 
> Perhaps, when we cache Text Encoder outputs, another way might be fine to change the dtype of Text Encoders to fp16/bf16 in advance...

Yeah I also add "te.to(weight_dtype)" too

KohakuBlueleaf:
@kohya-ss Oh I got what you said.
Basically we cache TE with weight_dtype is better?
I also found that sdxl_train.py cache TE in fp32.
I think this is bad (since cached context is actually huge, and will slow down the caching process)

KohakuBlueleaf:
@kohya-ss I added autocast TE caching into sdxl_train.py and also add seperated LR settings for TE in sdxl_train.py which is requested by @Linaqruf

KohakuBlueleaf:
> > Basically we cache TE with weight_dtype is better?
> 
> I think so. In my understanding, it is same to apply autocast. (However, we convert back TEs to float32 and move to cpu for sampling images.)
> 
> > @kohya-ss I added autocast TE caching into sdxl_train.py and also add seperated LR settings for TE in sdxl_train.py which is requested by @Linaqruf
> 
> This is really nice!

autocast and changing dtype is actually different.
autocast will convert your weight into target dtype(if backend accept it) and then do computation.
So convert weight directly still need autocast to ensure all the operation can run normally

Linaqruf:
> @kohya-ss I added autocast TE caching into sdxl_train.py and also add seperated LR settings for TE in sdxl_train.py which is requested by @Linaqruf

Nice 😆

KohakuBlueleaf:
> When I apply this PR and train with sdxl_train.py including Text Encoder, it seems that neither U-Net nor Text Encoder is trained. When training only U-Net, there is no problem.
> 
> I will investigate further. But I think that perhaps multiple models may not work when specified like: `{"params": ..., "lr": ...}`...

Ok this makes sense
Will also check that!

KohakuBlueleaf:
@kohya-ss it is weird.
I checked my last training with this fork
it is definitely learning things.

but it cannot run fp16 (full fp16 not work, full bf16 work)
said
```
AssertionError: No inf checks were recorded for this optimizer.
```

araleza:
Hi, I think you may need to make the fix to the Text Encoder learning rate code in the --block_lr section that I found here:

https://github.com/kohya-ss/sd-scripts/issues/890#issuecomment-1779601676

KohakuBlueleaf:
@araleza @kohya-ss This is weird for me
In my implementation, both TE and UNet only put their generator into the params dict
But When I tried to train it, It can definitely learn things. (Expected: don't learn anything since all the param generator are used)

KohakuBlueleaf:
But I still made a fix with it

KohakuBlueleaf:
@kohya-ss I also add a manual timeout settings for DDP since for some large dataset with multi gpu training. It is very likely to runout the default timeout(30min) when caching the latents or TextEncoder output.
(For example: the dataset I'm using need literally 3days to cache the Latents)

FurkanGozukara:
when we can expect this to be merged thank you so much

FurkanGozukara:
> @KohakuBlueleaf Thanks for the various updates! The script seems to be working fine.
> 
> I would like to add some features after the merge, such as specifying independent learning rates for Text Encoder 1 and 2, excluding from the optimizing parameters for models with 0 specified for the learning rate, etc.

so this will support both SD 1.5 and SDXL?

PR #900, opened 2023-10-27
Add PagedAdamW32bit

xzuyn:


oliverban:
This is great, but isn't the 32bit part still expensive in terms of memoy? Or is this different for images? I really like this way of optimization but if we can get a PagedAdamW8Bit, wouldn't that be even better? I read that but they were talking about LLMs.

xzuyn:
> This is great, but isn't the 32bit part still expensive in terms of memory?

In terms of memory usage it's more expensive than using 8bit, but since its `Paged` that memory isn't eating into GPU VRAM, only system RAM. So if you have the RAM for it, using 32bit is higher precision. No idea how much that translates into real world quality though.

> I really like this way of optimization but if we can get a PagedAdamW8Bit, wouldn't that be even better?

[It's already implemented.](https://github.com/kohya-ss/sd-scripts/blob/95ae56bd22c285ccb2fe5fca96d92f39842bb99b/library/train_util.py#L3461) I added `PagedAdamW32bit` here since *only* `PagedAdamW8bit` was added.

There may also be a 16bit paged version just named `PagedAdamW` but that may just be an alias for `PagedAdamW32bit` or `PagedAdamW8bit`. I'll have to check into this later.

PR #902, opened 2023-10-27
Check format by black

shirayu:
- Add black format check every commit
- This helps keep code readability at all times

- Currently, the formatting has not been done. So the test will be failed. Please perform formatting at the appropriate time.
    - This is an example of successful test after the format: https://github.com/shirayu/sd-scripts/actions/runs/6669255174/job/18126620573


akx:
I would suggest [`ruff format`](https://docs.astral.sh/ruff/formatter/) instead; it's compatible with `black` but much faster.

`ruff` is also a powerful linter; running linting might avoid some bugs.

Adding a [`pre-commit`](https://pre-commit.com/) configuration would make it easier for contributors to run the formatting automatically as they're committing code:

```yaml
repos:
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.1.10
    hooks:
      - id: ruff-format
```

shirayu:
Hello, thanks for the suggestion!
I didn't know the ruff.
It seems to be a nice tool.

I've created a new pull request #1055 and close this now.

PR #905, opened 2023-10-29
Replace print with logger if they are logs

shirayu:
This PR replaces print with logger if they are logs.

This helps distinguish logs from other outputs.
In addition, this replacement can also determine whether logs are warnings, information, or errors, and can understand what line of the file they come from.

### DIFF
--- a/library/utils.py
+++ b/library/utils.py
@@ -1,6 +1,21 @@
 import threading
 from typing import *
-
+import logging
 
 def fire_in_thread(f, *args, **kwargs):
     threading.Thread(target=f, args=args, kwargs=kwargs).start()
+
+def get_my_logger(name: str):
+    logger = logging.getLogger(name)
+    logger.setLevel(logging.INFO)
+
+    stream_handler = logging.StreamHandler()
+    stream_handler.setLevel(logging.INFO)
+
+    myformat = '%(asctime)s\t[%(levelname)s]\t%(filename)s:%(lineno)d\t%(message)s'
+    date_format = '%Y-%m-%d %H:%M:%S'
+    formatter = logging.Formatter(myformat, date_format)
+    stream_handler.setFormatter(formatter)
+    logger.addHandler(stream_handler)
+
+    return logger
--- a/fine_tune.py
+++ b/fine_tune.py
@@ -20,6 +20,8 @@ except Exception:
 from accelerate.utils import set_seed
 from diffusers import DDPMScheduler
 
+from library.utils import get_my_logger
+logger = get_my_logger(__name__)
 import library.train_util as train_util
 import library.config_util as config_util
 from library.config_util import (
@@ -51,11 +53,11 @@ def train(args):
     if args.dataset_class is None:
         blueprint_generator = BlueprintGenerator(ConfigSanitizer(False, True, False, True))
         if args.dataset_config is not None:
-            print(f"Load dataset config from {args.dataset_config}")
+            logger.info(f"Load dataset config from {args.dataset_config}")
             user_config = config_util.load_user_config(args.dataset_config)
             ignored = ["train_data_dir", "in_json"]
             if any(getattr(args, attr) is not None for attr in ignored):
-                print(
+                logger.warning(
                     "ignore following options because config file is found: {0} / 設定ファイルが利用されるため以下のオプションは無視されます: {0}".format(
                         ", ".join(ignored)
                     )
@@ -88,7 +90,7 @@ def train(args):
         train_util.debug_dataset(train_dataset_group)
         return
     if len(train_dataset_group) == 0:
-        print(
+        logger.error(
             "No data found. Please verify the metadata file and train_data_dir option. / 画像がありません。メタデータおよびtrain_data_dirオプションを確認してください。"
         )
         return
@@ -99,7 +101,7 @@ def train(args):
         ), "when caching latents, either color_aug or random_crop cannot be used / latentをキャッシュするときはcolor_augとrandom_cropは使えません"
 
     # acceleratorを準備する
-    print("prepare accelerator")
+    logger.info("prepare accelerator")
     accelerator = train_util.prepare_accelerator(args)
 
     # mixed precisionに対応した型を用意しておき適宜castする
@@ -460,7 +462,7 @@ def train(args):
         train_util.save_sd_model_on_train_end(
             args, src_path, save_stable_diffusion_format, use_safetensors, save_dtype, epoch, global_step, text_encoder, unet, vae
         )
-        print("model saved.")
+        logger.info("model saved.")
 
 
 def setup_parser() -> argparse.ArgumentParser:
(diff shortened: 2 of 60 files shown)


library/utils.py
```diff
@@ -1,6 +1,21 @@
...
+    stream_handler = logging.StreamHandler()
+    stream_handler.setLevel(logging.INFO)
+
+    myformat = '%(asctime)s\t[%(levelname)s]\t%(filename)s:%(lineno)d\t%(message)s'
+    date_format = '%Y-%m-%d %H:%M:%S'
+    formatter = logging.Formatter(myformat, date_format)
+    stream_handler.setFormatter(formatter)
+    logger.addHandler(stream_handler)
+
+    return logger
```
akx, line 21: I would advise against a `get_my_logger` function like this.

The preferred way to use `logging` is to just use `logging.getLogger(__name__)` in all modules as suitable, and configure the root logger, not each logger by itself.

The log messages will then be propagated up from the named loggers into the root logger, which will print them.

akx (changes requested):
Changing `print()`s to logging calls is a great idea, but I think this PR proposes a slight misuse of `logging` (please see my review comment within).

library/utils.py
shirayu: Yes, all calls of the function `get_my_logger` are like this.

```python
logger = get_my_logger(__name__)
```

So, ``name`` is ``__name__`` in ``logging.getLogger(name)``.


I think this is what you mentioned.
Please let me know if I misunderstood your intent.

akx: Sorry if I was unclear! I did mean that you should just use `logging.getLogger(__name__)` instead of a function that gets a logger and configures that particular logger "by hand".

Instead, it's better practice to configure the root logger, either by hand or with `logging.basicConfig()`, once per application. You can see an example of that in my PR to `stable-diffusion-webui` here: https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/14538 (though it's a bit more complex than it needs to be).

An additional issue with modifying a single logger like this PR is currently doing is that if you call `get_my_logger("something")` twice, the `"something"` logger will have two handlers, i.e. all messages logged with it will end up being printed twice.

shirayu: Thank you for your comment with an example!
I got it.
You mean that the call of logger can be like this.

```python
setup_logging() # call configuration of the root logger first
logging.getLogger(__name__) # call logger as usual way
```

akx: Yes, that's the idea. 😄 

`setup_logging()` would be called only once per program (so in the case of this repository, e.g. `train_network.py` and so on).

shirayu:
- Improvised initialization of logger ( Thanks to @akx 's comments )
- Added `rich`

akx:
> If we want to achieve this, should we initialize the logging framework in `__main__` function immediately after argparse?

Logging is now initialized early in each script by 

```python
from library.utils import setup_logging
setup_logging()
```

I think a follow-up PR can change that to happen after `argparse` (to allow setting via options), if needed.

akx:
> However, just curious for me, what is the advantage of initializing the log at the beginning of the script instead of initializing it in main?

No difference - unless something logs something "too early", in which case `logging` may fall back to its internal last resort logger for critical messages. For this project, that's not a big concern :)

shirayu:
@kohya-ss The following is an example of ignoring errors in the log initialization section.
This way, we can simply copy the code to another project and not have to change it.

> It appears that the utility scripts are copied and used in other projects (including my own). The log output in that case seems to be troublesome. Do you have any good ideas?

```python
# current
from library.utils import setup_logging
 setup_logging()
````

```python
# an idea
try:
    from library.utils import setup_logging
    setup_logging()
except ImportError:
    pass
```

> In addition, it seems that sd-scripts code is sometimes incorporated into other projects. Proper use of the logging framework may actually be a difficult task...

``logging`` is a standard python module, so I think it should not be a problem.
However, if users want to maintain compatibility with conventional ``print``,  adding the following statement is an option.

```python
# current
logger = logging.getLogger(__name__)
```

```python
# an idea
logger = logging.getLogger(__name__)
for f in [logger.debug, logger.info, logger.warning, logger.error, logger.critical]:
     f = print
```

akx:
> I just noticed that the logging framework is initialized at the beginning of utility scripts as well as scripts executed directly from the command line.

`setup_logging()` getting called multiple times should not be an issue, but I don't think any module in `library/` should be calling it. (I didn't check if anything in there is doing that right now.)

> It appears that the utility scripts are copied and used in other projects (including my own). The log output in that case seems to be troublesome. 

Mm, troublesome how? Using `logging` instead of `print` is more easily controlled, since an application embedding those scripts can selectively turn logger levels to what they like:

```python
logging.getLogger("library.hoge").setLevel(logging.ERROR)  # quiesce all hoge except errors
logging.getLogger("library.puyo").setLevel(logging.DEBUG)  # make puyo louder!
```

Whereas with `print()`, there's no easy way to filter it at all 😁 

> However, if users want to maintain compatibility with conventional print, announce to add the following statement is an option.

Unfortunately the suggested replacements are not a good idea – `logger.debug` and friends have a different signature than `print`.

Users who want log output to look like `print`s should just configure `logging` to output to `stdout` without formatting – and since loggers cascade, they can even do

```python
logging.getLogger("library").addHandler(...)
```
and all `library.whatever` logging will get handled by that handler instead of the root logger.

akx:
Oh, and by the way – to avoid people needing to copy-paste these scripts into their other projects, I will gladly lend a hand to make this repository installable as a Python package (like I did yesterday with https://github.com/huchenlei/HandRefinerPortable/pull/2) – but that's a matter for a different PR, of course 😄

shirayu:
@kohya-ss 

> So that, would the following code do it?

I think this is what you expect.


```python
import logging

try:
    from library.utils import setup_logging

    setup_logging()
except ImportError:

    class PrintHandler(logging.Handler):
        def emit(self, record):
            print(self.format(record))

    handler = PrintHandler()
    logging.root.addHandler(handler)
    logging.root.setLevel(logging.DEBUG) # print ALL logs

logger = logging.getLogger(__name__)
```

akx:
Ah, @shirayu, I'm not quite following why there would ever be an import error for `library.utils`? 😅  

Also, could you remove the `setup_logging()` calls from `library` (and other modules that are supposed to be only imported and not run directly)? The idea with the `logging` module is that importing a library module in general should not be configuring logging.

> Hmm... If I use other libraries, I don't like the detailed formatted log output from these libraries. Therefore, I personally would like the default behavior of each scripts to be the same as the current print.

If I understand you correctly here, you'd want `logger` output from the main scripts (e.g. `sdxl_train`) to look like regular prints (when run stand-alone)? Then we could

```python
logger = logging.getLogger(__name__)
# ...
if __name__ == "__main__":
    from library.utils import configure_logger_to_print
    configure_logger_to_print(logger)
    # ... other initialization ...
```
and something like
```python
def configure_logger_to_print(logger):
    h = logging.StreamHandler(sys.stdout)
    h.setLevel(logging.DEBUG)
    h.propagate = False  # prevent parent loggers from handling this
    logger.addHandler(h)
```
?

PR #907, opened 2023-10-29
Add option --sample_at_first

shirayu:
- :sparkles: Generate samples before training
    - This is useful for comparison before and during the training

## :spiral_notepad: Note

There is a possibility that the images generated by this and the images generated by ``sdxl_gen_img.py`` (or other generation scripts) are slightly different, even with the same parameters. Whether this is true or not has not yet been investigated.

### DIFF
--- a/fine_tune.py
+++ b/fine_tune.py
@@ -303,6 +303,9 @@ def train(args):
         accelerator.print(f"\nepoch {epoch+1}/{num_train_epochs}")
         current_epoch.value = epoch + 1
 
+        # For --sample_at_first
+        train_util.sample_images(accelerator, args, epoch, global_step, accelerator.device, vae, tokenizer, text_encoder, unet)
+
         for m in training_models:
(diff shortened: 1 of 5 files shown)


shirayu:
I have noticed that the reason for the difference in results may be due to ``--full_bf16`` or ``--full_bf16`` in training.
It would be better to add ``is_full`` argument to ``sdxl_train_util.load_target_model()`` and convert ``unet``, `text_encoder1`, `text_encoder2` to fp16 or bf16.

shirayu:
I have noticed that the reason for the difference in results may be due to ``--full_bf16`` or ``--full_bf16``.
It would be better to add ``is_full`` argument to ``sdxl_train_util.load_target_model`` and convert the unet, text_encoder1, text_encoder2 in it.

Related to #628

araleza:
This seems like a good feature.  I found myself wanting it just last week.

PR #913, opened 2023-10-30
Add custom seperator for shuffle caption

KohakuBlueleaf:
as title


KohakuBlueleaf:
> Sorry for the long delay. I've also updated wd14 tagger script.

That's not a problem!

PR #949, opened 2023-11-14
Add AttentionProcessor support

rockerBOO:
Background: I was working to add [DAAM](https://github.com/castorini/daam) attention mapping heatmaps into the sampling pipeline (and possibly into the training process). For my own purposes but also to propose to you. To do this I needed to add AttentionProcessor like support and many small changes to be able to have DAAM use it's diffusers interface to work with the Kohya original_unet and sampling pipelines. 

Diffusers made some changes and added a AttentionProcessor which abstracts out the attention mechanism and allows one to add their own processor in there. The DAAM library hooks into this processor to do cross attention mapping into a heatmap. https://github.com/castorini/daam/blob/v0.1.0/daam/trace.py#L281-L282

This also adds some additional code from the [Diffusers Attention class](https://github.com/huggingface/diffusers/blob/4b45a1e147e4f8fab31b888f50db74b30a1aa105/src/diffusers/models/attention_processor.py#L38) to further support the manipulations of attention that DAAM requires. 

I added the dropout architecture back in because DAAM was looking for the last node and it missing caused it to error out. We could make it `nn.Identity` instead if that's faster. 

The `encoder_hidden_states` was being used on Diffusers side which I believe is related to the `context` term so I mapped that over. This allows other libraries to pass diffusers like variables or the original `context` and should work similarly. This happens due to a lot of **kwargs like mappings. 

Ultimately I made these changes to limit any further refactoring into maybe a diffusers like architecture and better allow sd-scripts to work with diffusers. I know this is a lot of changes and additions. I can refactor it down if necessary. Thank you!

### DIFF
--- a/library/original_unet.py
+++ b/library/original_unet.py
@@ -790,7 +790,7 @@ class AttnProcessor:
 
         # linear proj
         hidden_states = attn.to_out[0](hidden_states)
-        hidden_states = attn.to_out[1](hidden_states)
+        # hidden_states = attn.to_out[1](hidden_states) # No dropout
         return hidden_states
 
 
@@ -849,6 +849,7 @@ class FlashAttentionProcessor:
         x = hidden_states
         h = attn.heads
         q = attn.to_q(x)
+        context = encoder_hidden_states if encoder_hidden_states is not None else context
         context = context if context is not None else x
         context = context.to(x.dtype)
         k = attn.to_k(context)
--- a/library/original_unet.py
+++ b/library/original_unet.py
@@ -548,11 +548,18 @@ class CrossAttention(nn.Module):
         heads: int = 8,
         dim_head: int = 64,
         upcast_attention: bool = False,
+        upcast_softmax: bool = False,
+        cross_attention_norm: Optional[str] = None,
+        cross_attention_norm_num_groups: int = 32,
+        added_kv_proj_dim: Optional[int] = None,
+        dropout: float = 0.0,
     ):
         super().__init__()
         inner_dim = dim_head * heads
         cross_attention_dim = cross_attention_dim if cross_attention_dim is not None else query_dim
         self.upcast_attention = upcast_attention
+        self.upcast_softmax = upcast_softmax
+        self.dropout = dropout
 
         self.scale = dim_head**-0.5
         self.heads = heads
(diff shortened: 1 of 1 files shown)


rockerBOO:
@ThereforeGames It works but still a little rough. if you can figure out the following steps you could probably try it out. 

https://github.com/rockerBOO/sd-scripts/tree/daam_sampling This branch for the sampling. 
https://github.com/rockerBOO/daam/pull/2 This PR for the DAAM library.

In the sample file: 
`a photo of a woman smiling, waving with her hands --d 42 --da woman,hands,smiling`

rockerBOO:
Thank you @kohya-ss . I appreciate your kind words. 

I was in the middle of working with that library to get this working and the changes I initially proposes with were just enough to get it working. After working with the library and augmenting it to work with CompVis like library I found it was assuming too many things to exist. I have reworked the library completely at this point to better support different types of libraries. So most of these changes I initially proposed won't be necessary to support.

If we look at the current sd-scripts version we are doing an attention processor in the forward. So my proposal has reduced significantly to just adding a `self.processor` and passing the values into that. This allows us to attach into the attention flow without having to resort to monkey patching. I added the additional names from HF Diffusers so it will translate appropriately. It would error if the names in the call were not there. 

The current proposal would be adequate for my needs. The most limited version of what I'd need is the naming in the forward needs to support the HF Diffusers names

- `hidden_states`
- `attention_mask`
- `encoder_hidden_states`

Thank you for taking the time to look at this proposal.

rockerBOO:
> However, I still wonder it might be advantageous to leave the existing code as is and only call the AttnProcessor if and when it is set, so that some compatibility testing with the existing code can be omitted.

I went through and simplified it to the approach you hinted at here. See #961 

I believe this is the simplest approach and captures all the properties for the forward arguments (for cases were people just pass `**kwargs` to the forward). Up to you which would be more impactful but I think #961 is better in every way. This reduces the amount of testing we will need to do as it will be segmented to only those who set the processor. 

No worries either way. I'm happy with this process we went through to reduce this down. 

Thank you again for your time.

rockerBOO:
> Thank you for updating, and sorry again for the delay. I agree that #961 seems to be the simplest approach.
> 
> Is it OK to merge #961 and close this without merging?

Yes, merging #961 and closing this one. I appreciate it and no worries. :)

PR #959, opened 2023-11-18
Add config args to Accelerator init_tracker

rockerBOO:
Related PR #792

Might need a "clean" for the config arguments but this may be better than nothing.

Screenshot 2023-11-18 at 16-25-24 Weights   Biases


rockerBOO:
Fixed tensorboard support. Only supports int, bool, str, float, and torch.Tensor.

PR #986, opened 2023-12-06
Fixed the path error in finetune/make_captions.py

CjangCjengh:
After executing
```python
os.chdir("finetune")
```
An additional `..` is needed in front of the path for caption_weights
```python
if not is_url(args.caption_weights):
    args.caption_weights = os.path.join("..", args.caption_weights)
```


CjangCjengh:
Thank you for your advice. And I have changed it to
```python
if not is_url(args.caption_weights) and not os.path.exists(args.caption_weights):
    args.caption_weights = os.path.join("..", args.caption_weights)
```

zadokov:
@CjangCjengh @bmaltais 
After merge - when trying BLIP Captioning I get the below log and captioning is aborted.

Current Working Directory is:  C:\SD\kohya_ss
  File "C:\SD\kohya_ss\finetune\make_captions.py", line 202, in <module>
    main(args)
  File "C:\SD\kohya_ss\finetune\make_captions.py", line 79, in main
    if not is_url(args.caption_weights) and not os.path.isfile(args.caption_weights):
NameError: name 'is_url' is not defined

PR #991, opened 2023-12-06
Added optional embeddings to train LoRA

Poiuytrezay1:


cyber-meow:
Hi [Poiuytrezay1](https://github.com/Poiuytrezay1),

Thank you for implementing this. I have been using pivotal tuning with HCP for a while. There, it is possible to set up some initialization words and token length, and it is also easy to have multiple embeddings configured in the configuration file.

I am wondering how you would achieve the same here. Do you have a separate script for embedding initialization (just like in the case of HCP)? For multiple embeddings I can probably just put a list in a toml file, but I am wondering if you see a better solution here.

madman404:
Apologies if this is a stupid question, but - when using this method, should the dataset be captioned with the name of the embedding? The words inside the embedding? the tokens? or not at all?

I saw this PR and was interested to give pivotal tuning a try now that it's in a suite I'm a little more familiar with, but this question doesn't really appear to be answered anywhere despite how important it is.

cyber-meow:
> Apologies if this is a stupid question, but - when using this method, should the dataset be captioned with the name of the embedding? The words inside the embedding? the tokens? or not at all?
> 
> I saw this PR and was interested to give pivotal tuning a try now that it's in a suite I'm a little more familiar with, but this question doesn't really appear to be answered anywhere despite how important it is.
> 
> Also, apologies if this is the wrong place to ask. I'm unfamiliar with github and would've reached out personally if I knew how.

Should be the word of the embedding. Any other alternative does not make sense.

Poiuytrezay1:
> Apologies if this is a stupid question, but - when using this method, should the dataset be captioned with the name of the embedding? The words inside the embedding? the tokens? or not at all?
> 
> I saw this PR and was interested to give pivotal tuning a try now that it's in a suite I'm a little more familiar with, but this question doesn't really appear to be answered anywhere despite how important it is.
> 
> Also, apologies if this is the wrong place to ask. I'm unfamiliar with github and would've reached out personally if I knew how.

Short answer: yes, you need to caption your images with this new token to train the LoRA with it.

Long answer: one token == one embedding. The tokenizer breaks down your tags into tokens, and the tokens are encoded into embeddings.  For example, "wyvern" is broken down into two tokens (wy + vern):
image

Using TI to create a new embedding for "wyvern" will add the token string "wyvern" at the end of the tokenizer and create an embedding associated with the new token that has nothing to do with either "wy" and "vern". When the tokenizer tries to tokenize your prompt, it will then return "wyvern" as a single token instead of two. This token is encoded into a single embedding (768 dimension vector) trained to get close to your concept.

Now, it is possible to have multiple vectors for a single embedding, but this is a hack from the SD community. Basically, it adds as many tokens to the tokenizer as there are vectors (e.g. for 3 vectors it would add "wyvern", "wyvern1", and "wyvern2"). Then, during generation, it replaces the initial token "wyvern" with  "wyvern wyvern1 wyvern2" so that the text encoder will produce 3 embeddings instead of one.

When using the script, it will take the name of your file and add this token string at the end of the tokenizer. Whenever it encounters the new token in your captions, it will use the added embedding instead of breaking it into multiple tokens.

JaredS215:
Is there any guidance on how many vectors the TI should have for use with Pivotal Tuning?

giggily:
> Is there any guidance on how many vectors the TI should have for use with Pivotal Tuning?

Ideally it should just be one. Using multiple vectors is effectively a workaround for a problem that PT solves.

toyssamurai:
Just to make sure I understand it correctly -- in order to use this feature, I need to train the embedding first, then pass the paths of the output files using the "--embeddings" when training for the LoRA? I assume that it also means I should use "--network_train_unet_only" for LoRA training? And, the same dataset should be used for both embedding and LoRA trainings?

yagemon:
I've had success using this to train 1.5 loras, but I can't get SDXL with SDXL embeddings to work:

```
load StableDiffusion checkpoint: \models\Stable-diffusion\sd_xl_base_1.0.safetensors
building U-Net
loading U-Net from checkpoint
U-Net:  <All keys matched successfully>
building text encoders
loading text encoders from checkpoint
text encoder 1: <All keys matched successfully>
text encoder 2: <All keys matched successfully>
building VAE
loading VAE from checkpoint
VAE: <All keys matched successfully>
Enable xformers for U-Net
Loaded token strings ['testertoken']
Textual Inversion embeddings `testertoken` loaded. Tokens are added: [49408]
  File "\sd-scripts\sdxl_train_network.py", line 185, in <module>
    trainer.train(args)
  File "\sd-scripts\train_network.py", line 256, in train
    text_encoder.get_input_embeddings().weight.data[token_id] = embed
TypeError: can't assign a str to a torch.FloatTensor
```

gesen2egee:
I suspect that when using multiple embeddings, there are misalignments. First, the length of the first embedding is incorrect, even exceeding 77 characters, making it unusable. Additionally, when training lora for multiple characters (>30), certain characters in fixed positions fail to learn properly and even unrelated characters appear, with other characters also inheriting features from different characters. After thoroughly checking the training and embeddings, and creating the embeddings in different ways, the issue still occurs with the same characters. Eventually, using non-Pivotal Tuning methods led to successful training. (for SD 1)

feffy380:
@Poiuytrezay1
> HCP diffusion seems to create empty embeddings and then fine-tune them at the same time as the unet. While this approach might yield good results, it's important to independently verify the embeddings, as they might not accurately capture the intended concepts on their own.

For what it's worth, the original [LoRA paper](https://arxiv.org/pdf/2106.09685.pdf) mentions exactly this approach (LoRA+PrefixEmbed under Appendix E). They only give a single set of training settings for the combined method, so I have to assume they train the LoRA and the embedding simultaneously. What the pivotal inversion paper refers to as an embedding seems to just be a point in latent space in our terminology, while LoRA+PrefixEmbed is a lot closer to LoRA+TI (where the prefix embedding/TI are both text embeddings)

PR #989, opened 2023-12-06
Fix gradients synchronization

Isotr0py:
- Related issue: #965 

- Fix #924

- Remove `train_util.transform_if_model_is_DDP`, this caused the gradients sync bug on DDP. Now model will be unwrapped by `accelerator.unwrap_model()` manually.

- Clean `accelerator.prepare()` codes in `train_network.py`

- Sync lora network gradients in `train_network.py` manually. 

For `train_network.py`, since we apply the lora through replacing the `forward()` method instead of the module, the DDP lora won't sync gradients automatically when calling `accelerator.backward(loss)`. Hence we need to use `accelerator.reduce()` to sync grad manually.

PR #998, opened 2023-12-11
Sync steps to prevent unexpected crash

jihnenglin:
I've found a bug that occurs while finetuning SD XL with multi-GPU configuration.

### Bug description
The steps across different accelerator processes are not synchronized. It happened when I configured `--sample_every_n_steps`. For example, the progress bar shows it's in the 950th step, but somehow one of the accelerator processes says it has already built a pipeline for generating the sample image for the 1000th step, yet another process says it's about to generate the sample image for the 900th step. If I were to let it run further, the discrepancy in steps across different processes would get even more significant and the script would eventually crash, throwing an NCCL DDP timeout error.

I've also tried to configure `--ddp_timeout=999999` in the training argument. It didn't throw a timeout error this time and finished training, but the script stalled just before it could save the final model checkpoint.

### What I did
I've added several `accelerator.wait_for_everyone()` to ensure that all accelerator processes synchronize step counts, without slowing down the overall training process. From my empirical findings, the output sample images I got from `--sample_every_n_steps` improved in quality after this fix (random seed controlled).

The above scenarios happened for `sdxl_train.py`, but I suspect it could happen to other training scripts. This pull request only tackled `sdxl_train.py`, so you might want to test other training scripts as well.

### DIFF
--- a/sdxl_train.py
+++ b/sdxl_train.py
@@ -591,6 +591,7 @@ def train(args):
                 lr_scheduler.step()
                 optimizer.zero_grad(set_to_none=True)
 
+            accelerator.wait_for_everyone()
             # Checks if the accelerator has performed an optimization step behind the scenes
             if accelerator.sync_gradients:
                 progress_bar.update(1)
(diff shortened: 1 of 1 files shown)


Isotr0py:
@jihnenglin

I think the crash is caused by the failed gradient sync fundamentally. And `accelerator.wait_for_everyone()` only make the processes look synchronous, while the gradients are still unsynchronized.

I'm afraid that this may cover up some serious bugs like the `train_util.transform_if_model_is_DDP` has caused...

- The PR #989 should has fixed this, it hasn't been merged to main branch, but in dev branch. 

Can you test the dev branch to see whether the crash is still existed? (training unet only should work, while there are some bugs in text_encoders haven't fixed yet)

jihnenglin:
@Isotr0py 

Thank you! I didn't realize that there was already a PR that addressed this issue, my bad. I'll test the dev branch and report later.

jihnenglin:
Since the issue was already solved in #1000, I'm closing this PR.

PR #1009, opened 2023-12-20
speed up latents nan replace

liubo0902:
use nan_to_num to speed up nan replace


liubo0902:
> Thank you for this! I didn't know `nan_to_num`. It is very nice.
> 
> Is it fine to remove `out=latents`, because the returned value is assigned to `latents`? I'm not sure which is better to omit `out` or omit the assignment.

with out, it is a inplace op, in this way, no new memory is allocated

PR #1032, opened 2024-01-03
Added cli argument for wandb session name

hopl1t:
Adds a command line argument that allows users to specify the wandb session name of their run.
Without this feature session names are not indicative.
While it is possible to fetch logger kwargs from a configuration toml this is cumbersome when sessions names are created in run time, which is very often the case.

This PR was first opened to the kohya_ss repo and now reopend in the sd_scripts repo as required.
https://github.com/bmaltais/kohya_ss/pull/1825


hopl1t:
@bmaltais thank you for creating and maintaining this beautiful repo :)

FurkanGozukara:
> @bmaltais thank you for creating and maintaining this beautiful repo :)

yes he is amazing

PR #1055, opened 2024-01-16
Check format by ruff

shirayu:
This is a replaced PR of #902.

- Add black format check every commit
- This helps keep code readability at all times

- Currently, the formatting has not been done. So the test will be failed. Please perform formatting at the appropriate time.

## Additional considerations after this PR
- Adding a [pre-commit](https://pre-commit.com/) configuration like https://github.com/kohya-ss/sd-scripts/pull/902#issuecomment-1893658672
- Use more rules of ruff for static analysis


akx:
Nice :smile: When actually applying the reformatting, we should note it will practically invalidate all current PRs (since there will be conflicts everywhere).

Regarding "Use more rules of ruff for static analysis", I actually applied some in https://github.com/kohya-ss/sd-scripts/compare/main...akx:sd-scripts:import-cleanup (it builds on the IPEX cleanup in #1054 so I'm not making a PR out of it yet).

shirayu:
Updated to `ruff==0.3.4` and performed `git rebase`

PR #1056, opened 2024-01-17
fix vram usage in LoRA training

kohya-ss:


FurkanGozukara:
Hello. will this reduce vram usage if only text encoder 1 trained and text encoder 2 is not?

PR #1060, opened 2024-01-18
Deduplicate ipex initialization code

akx:
This PR moves the repeated IPEX/XPU initialization code to a single place in `library/ipex/__init__.py`.

cc @Disty0


Disty0:
1) Don't change the init name to init_ipex or don't cahange the return values.  
I use my IPEX library with multiple projects. Unless you want to change it in every single project i use this library on, don't rename it or don't cahange the return values.

2) I am not sure if ipex will work with importing it in a submodule instead of the main file.  
This need testing when i get back to home.

akx:
Thanks for the comments!

> Don't change the init name to init_ipex and don't cahange the return values. I use my IPEX library with multiple projects. Unless you want to change them in every single project i use this library on, don't change them.

Fair, I can change those back – you could consider packaging your library as an actual Python package, though. If you need a hand with that, please let me know.

> I am not sure if ipex will work with importing it in a submodule instead of the main file.

It shouldn't have any technical difference. All this PR does is move the same code into a function and call it.

> This change makes it that if there is an issue with the ipex libs, entire script will error out.

I don't think so. The exact same exception handling is in the new initialization function, although in a slightly different form. Anyway, I'll change this a little more to avoid touching `library/ipex` at all :)

library/ipex_interop.py
```diff
@@ -0,0 +1,22 @@
...
+    try:
+        import intel_extension_for_pytorch as ipex  # noqa
+    except ImportError:
+        return
+
+    try:
+        from library.ipex import ipex_init
+
+        if torch.xpu.is_available():
+            ipex_init()
```
Disty0, line 20: You can add this too:
```python
is_initialized, error_message = ipex_init()
if not is_initialized:
    print("failed to initialize ipex:", error_message)
```

Disty0:
Other than the logging i mentioned, this pr is good to merge.

library/ipex_interop.py
akx: Done.

Disty0 (approved):

PR #1068, opened 2024-01-24
Soft min SNR gamma

rockerBOO:
In "Scalable High-Resolution Pixel-Space Image Synthesis with Hourglass Diffusion Transformers" they came up with Soft min SNR gamma which smooths out the transition area. 

Screenshot 2024-01-23 at 23-13-30 Scalable High-Resolution Pixel-Space Image Synthesis with Hourglass Diffusion Transformers - 2401 11605 pdf
Screenshot 2024-01-23 at 22-53-06 Scalable High-Resolution Pixel-Space Image Synthesis with Hourglass Diffusion Transformers - 2401 11605 pdf

31 = soft_min_snr_gamma = 5
32 = min_snr_gamma =  5
Screenshot 2024-01-24 at 00-01-36 Weights   Biases

Note I think the math is correct but I could be wrong so if anyone wants to correct I can update it.

### DIFF
--- a/library/custom_train_functions.py
+++ b/library/custom_train_functions.py
@@ -68,6 +68,25 @@ def apply_snr_weight(loss, timesteps, noise_scheduler, gamma, v_prediction=False
     return loss
 
 
+def apply_soft_snr_weight(loss, timesteps, noise_scheduler, gamma, v_prediction=False):
+    snr = torch.stack([noise_scheduler.all_snr[t] for t in timesteps])
+    # min_snr_gamma = torch.minimum(snr, torch.full_like(snr, gamma))
+    soft_min_snr_gamma_weight = 1 / (torch.pow(snr, 2) + (1 / float(gamma)))
+    with open("snr.txt", "a") as myfile:
+        myfile.write(f"{snr.item()},{gamma}\n")
+
+    # with open("snrmin.txt", "a") as myfile:
+    #     myfile.write(f"{min_snr_gamma.item()},{soft_min_snr_gamma.item()}\n")
+    # print("soft_min_snr_gamma", soft_min_snr_gamma, 1 / (snr + (1 / float(gamma))))
+    # print("min_snr_gamma", min_snr_gamma)
+    # if v_prediction:
+    #     snr_weight = torch.div(soft_min_snr_gamma, snr+1).float().to(loss.device)
+    # else:
+    #     snr_weight = torch.div(soft_min_snr_gamma, snr).float().to(loss.device)
+    loss = loss * soft_min_snr_gamma_weight
+    return loss
+
+
 def scale_v_prediction_loss_like_noise_prediction(loss, timesteps, noise_scheduler):
     scale = get_snr_scale(timesteps, noise_scheduler)
     loss = loss * scale
@@ -102,6 +121,12 @@ def apply_debiased_estimation(loss, timesteps, noise_scheduler):
 def add_custom_train_arguments(parser: argparse.ArgumentParser, support_weighted_captions: bool = True):
     parser.add_argument(
         "--min_snr_gamma",
+        type=float,
+        default=None,
+        help="gamma for reducing the weight of high loss timesteps. Lower numbers have stronger effect. 5 is recommended by paper. / 低いタイムステップでの高いlossに対して重みを減らすためのgamma値、低いほど効果が強く、論文では5が推奨",
+    )
+    parser.add_argument(
+        "--soft_min_snr_gamma",
         type=float,
         default=None,
         help="gamma for reducing the weight of high loss timesteps. Lower numbers have stronger effect. 5 is recommended by paper. / 低いタイムステップでの高いlossに対して重みを減らすためのgamma値、低いほど効果が強く、論文では5が推奨",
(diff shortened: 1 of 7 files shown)


library/custom_train_functions.py
```diff
@@ -68,6 +68,13 @@ def apply_snr_weight(loss, timesteps, noise_scheduler, gamma, v_prediction=False
     return loss
 
 
+def apply_soft_snr_weight(loss, timesteps, noise_scheduler, gamma, v_prediction=False):
+    snr = torch.stack([noise_scheduler.all_snr[t] for t in timesteps])
+    soft_min_snr_gamma_weight = 1 / (torch.pow(snr if v_prediction is False else snr + 1, 2) + (1 / float(gamma)))
```
feffy380, line 73: The math here is incorrect. SNR is equal to the whole expression `1/sigma**2`, not `sigma`.
The correct weight should be:
```
sigma2 = 1 / snr
1 / (sigma2 + 1/gamma)
1 / (1/snr + 1/gamma)
# simplified
weight = snr * gamma / (snr + gamma)
```
Finally, the given formulation for soft-min-snr is for `x_0` prediction. We use epsilon or v-prediction, which according to the original min-snr paper means we need to divide by SNR or SNR+1 respectively, so the final weight calculation should be:
```python
snr_weight = (snr * gamma / (snr + gamma)).float().to(loss.device)
if v_prediction:
    snr_weight /= snr + 1
else:
    snr_weight /= snr
```

rockerBOO: Tried this formula

```python
def apply_soft_snr_weight(loss, timesteps, noise_scheduler, gamma, v_prediction=False):
    snr = torch.stack([noise_scheduler.all_snr[t] for t in timesteps])
    snr_weight = (snr * gamma / (snr + gamma)).float().to(loss.device)
    if v_prediction:
        snr_weight /= snr + 1
    else:
        snr_weight /= snr
    loss = loss * snr_weight
    return loss
```
And produced the same loss curve as the current implementation. But the paper says it should match up except for the ones closer to the transition so the loss curves should be similar. Still seeing a difference with the Min SNR version though. 

38 = weight = snr * gamma / (snr + gamma)
37 = the current PR version
35 = Min SNR version. 

Screenshot 2024-01-24 at 11-58-51 Weights   Biases

Maybe there's something else that is missing in these calculations. 

Here is some example `snr,gamma` to use with the following test script
[snr.txt](https://github.com/kohya-ss/sd-scripts/files/14041456/snr.txt)

And using this code to test the formulas. 

```python
import math

with open("snr.txt", "r") as f:
    lines = f.readlines()

    print("snr min_snr soft soft2")
    for line in lines:
        snr, gamma = line.split(",")

        snr = float(snr)
        gamma = float(gamma)

        min_snr = min(1 / math.pow(snr, 2), gamma)
        soft_min_snr_gamma = 1 / (math.pow(snr, 2) + (1 / gamma))

        snr_weight = (snr * gamma / (snr + gamma))

        print(f"{snr:10.4f} {min_snr:4.4f}, {soft_min_snr_gamma:4.4f}, {snr_weight:4.4f}")
```

Birch-san:
here's how to formulate it as an EDM target:  
https://github.com/crowsonkb/k-diffusion/blob/6ab5146d4a5ef63901326489f31f1d8e7dd36b48/k_diffusion/layers.py#L65

here's how to formulate it as an x0 loss weighting:  
https://github.com/Birch-san/k-diffusion/blob/9bce54aec1e596548cf73f56f4842c11aa6271c6/k_diffusion/layers.py#L160

here's an alternative style for expressing it as an EDM target, where we use the x0 loss weighting and apply a correction to adapt it for EDM:  
https://github.com/Birch-san/k-diffusion/blob/9bce54aec1e596548cf73f56f4842c11aa6271c6/k_diffusion/layers.py#L250

library/custom_train_functions.py
```diff
@@ -106,6 +113,12 @@ def add_custom_train_arguments(parser: argparse.ArgumentParser, support_weighted
         default=None,
         help="gamma for reducing the weight of high loss timesteps. Lower numbers have stronger effect. 5 is recommended by paper. / 低いタイムステップでの高いlossに対して重みを減らすためのgamma値、低いほど効果が強く、論文では5が推奨",
     )
+    parser.add_argument(
+        "--soft_min_snr_gamma",
+        type=float,
+        default=None,
+        help="gamma for reducing the weight of high loss timesteps. Lower numbers have stronger effect. 5 is recommended by paper. / 低いタイムステップでの高いlossに対して重みを減らすためのgamma値、低いほど効果が強く、論文では5が推奨",
```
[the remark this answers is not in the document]
rockerBOO: Updated to recommend 1. Mistakenly I copied the help function from the min_snr_gamma option. Thank you!

rockerBOO:
> here's how to formulate it as an EDM target: https://github.com/crowsonkb/k-diffusion/blob/6ab5146d4a5ef63901326489f31f1d8e7dd36b48/k_diffusion/layers.py#L65
> 
> here's how to formulate it as an x0 loss weighting: https://github.com/Birch-san/k-diffusion/blob/9bce54aec1e596548cf73f56f4842c11aa6271c6/k_diffusion/layers.py#L160
> 
> here's an alternative style for expressing it as an EDM target, where we use the x0 loss weighting and apply a correction to adapt it for EDM: https://github.com/Birch-san/k-diffusion/blob/9bce54aec1e596548cf73f56f4842c11aa6271c6/k_diffusion/layers.py#L250

Thanks for sharing these. Not sure what EDM or x0 loss weighting are to know what makes sense to use here. I foolishly tried to implement it but do not understand the math underlying it. 

I will try these implementation details you shared to see if i can discern how to implement it properly into this code also integrating what feffy has suggested.

PR #1087, opened 2024-01-28
fix broken import in svd_merge_lora script

mgz-dev:
remove missing import, and remove unused imports so the script can be run.


mgz-dev:
You are correct, I was invoking the script from main dir with python -m networks.svd_merge_lora, apologies for the oversight.

PR #1100, opened 2024-02-03
Add Karras post-hoc EMA

vvern999:
This allows to generate different EMA profiles after training, without having to retrain.
Explanation of the method:  https://github.com/cloneofsimo/karras-power-ema-tutorial
Traditional EMA implementation is based on https://github.com/lucidrains/ema-pytorch

**Only train_network and sdxl_train_network are supported currently.**

### How to use:
Add `--enable_ema --ema_type="post-hoc"` to training options. `--ema_on_gpu` to use gpu.
It will save snapshots in `OUTPUT_DIR/OUTPUT_NAME_snapshots`
After training you can generate EMA checkpoints with `python .\tools\reconstruct_ema_weights.py PATH_TO_SNAPSHOT_FOLDER --target_gamma GAMMA`

Different values of gamma correspond to different "averaging length" sigma_rel. You can use `--target_sigma_rel` instead of gamma.
Screenshot 2024-02-01 151122

### Examples of different gamma:
SD1.5:
xyz_grid-resized
SDXL: 
resize-25

Different models may require different gamma for best results.

### DIFF
--- a/fine_tune.py
+++ b/fine_tune.py
@@ -32,7 +32,10 @@ from library.custom_train_functions import (
     scale_v_prediction_loss_like_noise_prediction,
     apply_debiased_estimation,
 )
-
+from library.train_util import (
+    EMA,
+    check_and_update_ema,
+)
 
 def train(args):
     train_util.verify_training_args(args)
@@ -240,6 +243,10 @@ def train(args):
         unet.to(weight_dtype)
         text_encoder.to(weight_dtype)
 
+    if args.enable_ema:      # u-net only
+        raise NotImplementedError
+        emas = train_util.setup_emas(args, unet)
+
     # acceleratorがなんかよろしくやってくれるらしい
     if args.train_text_encoder:
         unet, text_encoder, optimizer, train_dataloader, lr_scheduler = accelerator.prepare(
@@ -372,6 +379,7 @@ def train(args):
                 optimizer.step()
                 lr_scheduler.step()
                 optimizer.zero_grad(set_to_none=True)
+
 
             # Checks if the accelerator has performed an optimization step behind the scenes
             if accelerator.sync_gradients:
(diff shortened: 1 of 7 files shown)


feffy380:
100 is an extremely high value for gamma. Based on the explanation [here](https://github.com/cloneofsimo/karras-power-ema-tutorial), high values of gamma make the response time (averaging period) shorter. With gamma=100 you're essentially getting instant response, which completely negates whatever benefit EMA is supposed to bring.

You need to include a control test with no EMA. I suspect it will look very similar to the high gamma profiles.

vvern999:
@feffy380 
You are right, gamma=100 looks same as a model without ema, and with low gamma it looks same as no lora. 
As far as i understand, Lora is initialized with zeros, so with high averaging it just averages to 0. 

There's an option to start EMA updates only after x steps - `--ema_update_after_step`. I guess it may help, but it doesn't work right now
Also those lora are trained only for 750 steps, averaging is probably not practically useful with such low number of steps. People usually use EMA when training on thousands or millions of images for many steps

xyz_grid-3439-1953469724-best quality,_1girl, amane kanata, amane kanata 6th costume, ahoge, angel, angel wings, arm at side, hair ornament, black collar

FurkanGozukara:
from these experiment anime pictures i really can't understand quality

can't you test on a random guy face so we can see how much improvement it has?

dill-shower:
Any plans to update it for full finetune models?

gesen2egee:
Hello, I noticed that you've removed the args option for the traditional EMA. I'd like to know if the functionality of the traditional EMA is working fine now?

vvern999:
@gesen2egee  @dill-shower 
Sorry, I don't have the time or interest to test or finish this.
Traditional EMA worked in previous commits, but it didn't seem very useful.
I'm not sure if it's just not useful for a short training or if I made a mistake somewhere and it was not working correctly. 
Maybe someone else will implement it properly

CelesteHargreeves:
Will there be EMA for Flux Dev?

PR #1104, opened 2024-02-04
replace print with logger

kohya-ss:


fine_tune.py
```diff
@@ -461,12 +470,13 @@ def fn_recursive_set_mem_eff(module: torch.nn.Module):
...
             args, src_path, save_stable_diffusion_format, use_safetensors, save_dtype, epoch, global_step, text_encoder, unet, vae
         )
-        print("model saved.")
+        logger.info("model saved.")
 
 
 def setup_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser()
 
+    add_logging_arguments(parser)  # モジュール指定がないのがちょっと気持ち悪い / bit weird that this does not have module prefix
```
shirayu, line 479: Moving `add_logging_arguments()` and `setup_logging()` from `library/utils.py` to `library/logger_utils/__init__.py` is an option.

The call will be like this.
```python
logger_utils.add_logging_arguments(parser)
```

shirayu (changes requested):
Great!
Thank you for your updates!
Almost everything is fine, but I did make a few comments.

I am sure you already understand this, but just to be clear, I note that the call of `add_logging_arguments` is also needed in other files like `sdxl_train.py`.

shirayu:
I would like to add one more thing.
By default, ``rich`` logs to standard output, but I think it is better to change it as follows.

https://github.com/Textualize/rich/issues/2022

## Current 

```python
handler = RichHandler()
```

## Proposal

```python
from rich.console import Console    
from rich.logging import RichHandler
handler = RichHandler(console=Console(stderr=True))
```

library/utils.py
```diff
@@ -1,6 +1,72 @@
...
+    msg_init = None
+    if args is not None and args.console_log_file:
+        handler = logging.FileHandler(args.console_log_file, mode="w")
+    else:
+        handler = None
+        if not args or not args.console_log_simple:
+            try:
+                from rich.logging import RichHandler
+
+                handler = RichHandler()
```
shirayu, line 53: I would like to add one more thing.
By default, ``rich`` logs to standard output, but I think it is better to change it as follows.

https://github.com/Textualize/rich/issues/2022

## Current 

```python
handler = RichHandler()
```

## Proposal

```python
from rich.console import Console    
from rich.logging import RichHandler
handler = RichHandler(console=Console(stderr=True))
```

PR #1136, opened 2024-02-24
v0.8.4

kohya-ss:


FurkanGozukara:
Very nice work with highvram

How much speed up it leads and how much more VRAM it uses?

FurkanGozukara:
> `vae_batch_size` is a batch size used in caching latents. The default value is 1, so it will be quite faster if increasing `vae_batch_size` as much as possible.

haha i didnt know until all this time thank you so much for this amazing info

PR #1146, opened 2024-03-02
Faster training with token merging via tomesd

feffy380:
This PR adds support for training with token merging from the `tomesd` library (needs to be installed in the venv).
This significantly speeds up training at the cost of some quality loss. With 50% merge ratio I saw a 50% reduction in training time for SD1.x loras. The quality loss seems minor because the model learns to compensate during training.
I've added two new flags:
- `--tome_ratio` (float) token merging ratio. Float between 0 and 1. `tomesd` recommends 0.5 or lower and there are diminishing returns on speed improvement above this.
- `--stop_tome` (int or float) disable token merging after this many steps. Value less than 1.0 is treated as a fraction of total training time. The reason for this option is to minimize quality loss by disabling ToMe towards the end of training.

Currently, only train_network.py is supported. SDXL technically works too but I didn't see any speedup (might need different settings but it's also just more efficient overall so it benefits less from token merging).
Draft PR for now because I'm not sure if this is the best way to integrate the functionality. Patching could potentially be moved to `train_util.py` in the model loading function to avoid duplicate code. I am not sure how this might affect model saving for finetune or dreambooth. It only overrides the forward method of BasicTransformerBlock, so I think saving won't be affected?

Example:
| Name | Merge ratio | s/it | Speedup |
|---|---|---|---|
| feffy-v3.50 | 0 | 2 | 1x |
| feffy-tome30 | 0.3 | 1.36 | 1.47x |
| feffy-tome50 | 0.5 | 1 | 2x |

xyz_grid-5026-332293423-feffy female anthro latias, fluffy, fur tuft, cheek tuft, white fur, red hair, highlights _(coloring_), yellow eyes, black pupil

### DIFF
--- a/library/token_merging.py
+++ b/library/token_merging.py
@@ -18,21 +18,27 @@ def make_tome_block(block_class: Type[torch.nn.Module]) -> Type[torch.nn.Module]
         # Save for unpatching later
         _parent = block_class
 
-        def forward(self, hidden_states: torch.Tensor, context: torch.Tensor = None, timestep=None) -> torch.Tensor:
-            m_a, m_c, m_m, u_a, u_c, u_m = compute_merge(hidden_states, self._tome_info)
+    def forward(self, hidden_states: torch.Tensor, context: torch.Tensor = None, timestep=None) -> torch.Tensor:
+        m_a, m_c, m_m, u_a, u_c, u_m = compute_merge(hidden_states, self._tome_info)
 
-            # 1. Self-Attention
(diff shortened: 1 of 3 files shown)


feffy380:
I'm currently testing token downscaling from https://github.com/ethansmith2000/ImprovedTokenMerge
The implementation is simpler and quality is much better, so I will probably submit a new PR and close this one

feffy380:
I'm closing this PR in favor of the other one

PR #1151, opened 2024-03-03
Faster training with token downsampling

feffy380:
This PR adds support for training with [token downsampling](https://github.com/ethansmith2000/ImprovedTokenMerge) and replaces my token merge PR (#1146).

Token downsampling is a lossy optimization that significantly speeds up inference and training. It tries to avoid the quality loss of [token merging](https://github.com/dbolya/tomesd) by only downsampling K and V in the attention operation (Q is preserved) and replaces the expensive token similarity calculation with simple downsampling.
Applying the optimization during training seems to have less quality loss compared to inference, so I was able to increase the amount of downsampling a lot without negative effects.

With a downsampling factor of 2 and resolution of 768px I get a **2x speedup** for SD1.x LoRA training.
Downsampling at both levels with factors 4 and 2 gave me an even bigger **3.2x speedup** with no visible quality loss. I'm not sure how much this is affected by dataset, so YMMV.
SDXL benefits less because its architecture is already more efficient, but I still saw about 1.3x speedup at 1024px with downsample factor 2.

This PR adds two new flags:
- `--todo_factor`: (float) token downsampling factor. The inputs of the unet's self-attention layers are scaled down by this factor.
- `--todo_args`: (list[str]) pass "key=value" args to ToDo. See list below.

Advanced options:
- downsample_factor_depth_1: (float) amount to downsample in 1st down block of unet.
  - Default: `--todo_factor`
  - Note: Does nothing for SDXL because it doesn't have this block.
- downsample_factor_depth_2: (float) amount to downsample in 2nd down block of unet.
  - SD1/2 default: 1 (no downsampling)
  - SDXL default: `--todo_factor`
- downsample_method: (str) `torch.nn.functional.interpolate` mode argument for downsampling.
  - Default: `nearest-exact`

The unet is patched when the model is loaded, so the optimization should automatically work with all training scripts, but I only tested train_network.py and sdxl_train_network.py.
The downsampling operation is implemented with pytorch forward pre hooks, so model saving should be unaffected.

Example:
| Name | Downsample factors | s/it | Speedup |
|---|---|---|---|
| feffy-v3.50 | None | 2.0 | 1x |
| feffy-todo2 | 2/1 | 1.0 | 2x |
| feffy-todo4_2 | 4/2 | 0.63 | 3.2x |

image

Training details:
- 7900 XTX with `--mem_eff_attn`
- 768px resolution with bucketing
- Batch size 4

### DIFF
--- a/library/sdxl_train_util.py
+++ b/library/sdxl_train_util.py
@@ -12,6 +12,7 @@ from tqdm import tqdm
 from transformers import CLIPTokenizer
 from library import model_util, sdxl_model_util, train_util, sdxl_original_unet
 from library.sdxl_lpw_stable_diffusion import SdxlStableDiffusionLongPromptWeightingPipeline
+from library import token_merging
 from .utils import setup_logging
 setup_logging()
 import logging
@@ -56,6 +57,10 @@ def load_target_model(args, accelerator, model_version: str, weight_dtype):
 
             clean_memory_on_device(accelerator.device)
         accelerator.wait_for_everyone()
+
+    # apply token merging patch
+    if args.todo_factor:
+        token_merging.patch_attention(unet, args)
 
     return load_stable_diffusion_format, text_encoder1, text_encoder2, vae, unet, logit_scale, ckpt_info
 
--- a/library/token_merging.py
+++ b/library/token_merging.py
@@ -1,0 +1,106 @@
+# based on:
+#   https://github.com/ethansmith2000/ImprovedTokenMerge
+#   https://github.com/ethansmith2000/comfy-todo (MIT)
+
+import ast
+import math
+
+import torch
+import torch.nn.functional as F
+
+from library.sdxl_original_unet import SdxlUNet2DConditionModel
+
+from library.utils import setup_logging
+setup_logging()
+import logging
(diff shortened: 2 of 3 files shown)


gesen2egee:
Very impressive improvement in speed as well as maintaining training quality. 

Additionally, when I use --todo_args "downsample_method=nearest-exact", a parameter parsing error occurs, but it does not affect anything since it is the default value

feffy380:
@gesen2egee I see the problem and pushed a fix.

To be honest, `--todo_args` doesn't seem necessary. It's a leftover from the research code which had a lot more options. Here it's only needed for `downsample_method`, and the default method already works so well that there isn't much reason to use a more expensive interpolation mode. I could remove it and simplify the code

feffy380:
Ok I think that settles it. I'll go for simplicity

feffy-todo4_4-comparison

Bocchi-Chan2023:
このPRは私のRX 6600上で完璧に動作しています!
This PR works perfectly on my RX 6600!

kunibald413:
Hi, Very interesting, could you add that to the flux lora training aswell? or is it straight forward?

PR #1155, opened 2024-03-05
Support for multiple captions in one file

cheald:
This is a simple little change which adds support for using multiple captions in a single caption file, one per line. During training, each time the caption is needed, a random line is sampled.

I'm using this with a line of WD14 tags, a LLaVa natural language caption, and a LLaVa natural language caption which was prompted by including the WD14 tags. This seems to have helped quite a bit with training robustness.

It would be cleaner to assign an array of captions to `image_info.caption` rather than splitting the lines on each pass, but I wasn't sure if there was a dependency elsewhere on caption being a string, so this is just my first little quick and dirty pass.

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -1203,7 +1203,10 @@ class BaseDataset(torch.utils.data.Dataset):
             flippeds.append(flipped)
 
             # captionとtext encoder outputを処理する
-            caption = image_info.caption  # default
+            raw_caption = image_info.caption  # default
+            captions = [s.strip() for s in raw_caption.split("\n") if len(s.strip())>0]
+            caption = random.choice(captions)
+
             if image_info.text_encoder_outputs1 is not None:
                 text_encoder_outputs1_list.append(image_info.text_encoder_outputs1)
                 text_encoder_outputs2_list.append(image_info.text_encoder_outputs2)
@@ -1419,7 +1422,7 @@ class DreamBoothDataset(BaseDataset):
                             logger.error(f"illegal char in file (not UTF-8) / ファイルにUTF-8以外の文字があります: {cap_path}")
                             raise e
                         assert len(lines) > 0, f"caption file is empty / キャプションファイルが空です: {cap_path}"
-                        caption = lines[0].strip()
+                        caption = "\n".join([l.strip() for l in lines])
                     break
             return caption
 


FurkanGozukara:
this would be nice
i can implement this feature to my batch captioners so people can have all different captioners being used

FurkanGozukara:
> Thank you for this PR! I've implemented wildcard notation (like `{aaa|bbb|ccc}`) in dev branch. Is it fine to use it instead of this?

new line much easier
because to turn caption into this notation requires a lot of work
however appending new caption into a new line so easier

gesen2egee:
In the expansion of the web UI, using {A|B} is called Dynamic Prompts, while selecting randomly from different lines in a text file is called Wildcards. I think both can coexist. 

Additionally, I believe that combining a wildcard with "an image of XXX" should help in directly leveraging trigger words for effects.

cheald:
> Thank you for this PR! I've implemented wildcard notation (like `{aaa|bbb|ccc}`) in dev branch. Is it fine to use it instead of this?

I think they should be able to coexist; multiple captions, one per line, with wildcards within a line should be compatible with each other and should both serve to improve training robustness.

BootsofLagrangian:
If a user wishes to utilize multiple captions, derived from raw data, a tagger, or a Vision-Language Model (VLM), the script could handle this through an alternative format or file. This approach would address issues related to multiline captions.

Below are my suggestions and code modifications. I have altered the `caption` attribute in the `ImageInfo` class to a `property`. The `ImageInfo` class will now output the caption by managing a JSON file.

Change.
1. self.caption to self._caption
2. self.caption is a property
3. ImageInfo load JSON file which has same file name with image.
4. JSON contains "captions", "probabilities", "sampling method"
5. If script call ImageInfo.caption, ImageInfo will return sampled caption.

```
class ImageInfo:
    def __init__(self, image_key: str, num_repeats: int, caption: str, is_reg: bool, absolute_path: str) -> None:
        self.image_key: str = image_key
        self.num_repeats: int = num_repeats
        self._caption: str = caption
        self.is_reg: bool = is_reg
        self.absolute_path: str = absolute_path
        self.image_size: Tuple[int, int] = None
        self.resized_size: Tuple[int, int] = None
        self.bucket_reso: Tuple[int, int] = None
        self.latents: torch.Tensor = None
        self.latents_flipped: torch.Tensor = None
        self.latents_npz: str = None
        self.latents_original_size: Tuple[int, int] = None  # original image size, not latents size
        self.latents_crop_ltrb: Tuple[int, int] = None  # crop left top right bottom in original pixel size, not latents size
        self.cond_img_path: str = None
        self.image: Optional[Image.Image] = None  # optional, original PIL Image
        # SDXL, optional
        self.text_encoder_outputs_npz: Optional[str] = None
        self.text_encoder_outputs1: Optional[torch.Tensor] = None
        self.text_encoder_outputs2: Optional[torch.Tensor] = None
        self.text_encoder_pool2: Optional[torch.Tensor] = None

        # Sampling random caption
        self.json_caption = None
        """{
            "captions"  : List[str, ...],
            "p"         : List[float, ...],
            "sampling"  : choices=["weight", "softmax", "uniform"],
            ...
        }
        """
        self.setup_random_caption()

    @staticmethod
    def validate_json_caption_data(image_path: str, json_caption: dict) -> Tuple[List[str], np.ndarray]:
        captions: list = json_caption["captions"]
        p: list = json_caption["p"]
        assert all(isinstance(caption, str) for caption in captions), f"Type of caption must be string. Please check json file of image: {image_path}."
        assert isinstance(p, list), f"Type of probability must be list, but got {type(p)}"

        p: np.ndarray = np.asarray(p, dtype=np.float32)
        if len(p) != len(captions):
            print(f"Image: {image_path}. You should set same length for both probabilites and captions. Otherwise, it will use equal probability of each captions.")
            p = np.ones((len(captions)), dtype=np.float32) / float(len(captions))

        sampling: str = json_caption["sampling"]
        if sampling is None:
            return captions, p
        elif not isinstance(sampling, str):
            print(f"Image: {image_path}. sampling method is not a string. got type: {type(sampling)}, value: {sampling}")
            return captions, p

        if sampling.lower() == "weight":
            s_p = p.sum()
            if s_p != 1.:
                print(f"Image: {image_path}. Sum of probabilities is not equal to 1. It will automatically normalize probabilities.")
            p = p / s_p
        elif sampling.lower() == "softmax":
            def softmax(p: np.ndarray) -> np.ndarray:
                p = np.exp(p - p.max())
                p = p / p.sum()
                return p
            p = softmax(p)
        elif sampling.lower() == "uniform":
            p = np.ones((len(captions)), dtype=np.float32) / float(len(captions))
        else:
            assert False, f"Image: {image_path}. Not supported sampling method. got {sampling}, only support 'weight', 'softmax', 'uniform'"

        return captions, p

    @property
    def caption(self) -> str:
        if self.json_caption:
            _caption = np.random.choice(self._captions, p=self._captions_p)
            # print("[caption] load caption from json", _caption, "from ", self._captions, self._captions_p) # debug
            return _caption
        else:
            return self._caption

    def setup_random_caption(self) -> None:
        json_caption_path = os.path.splitext(self.absolute_path)[0] + ".json"
        if not os.path.exists(json_caption_path):
            self.json_caption = False
        else:
            with open(json_caption_path, "r", encoding="utf-8") as json_file:
                self.json_caption = json.load(json_file)

            captions, p = ImageInfo.validate_json_caption_data(self.absolute_path, self.json_caption)
            self._captions_p = p
            self._captions = captions

        assert self.json_caption is not None, f"Image: {self.image_key}. Loading json file failed."
```

FurkanGozukara:
i think you dont need to support all at once

for example for multi line caption people can provide a new parameter like

--multi_line_caption :)

cheald:
Right now, if the caption file contains multiple lines, all but the first are ignored; unless this is documented somewhere, I'd suggest that it's probably surprising behavior, and a random selection is better than none at all. My implementation does ignore blank lines, so there would be no chance of a blank line being randomly sampled.

That said, it wouldn't be hard to add a switch along the lines of --multiple_captions to turn on the multi-caption behavior, so that existing behavior can be fully preserved.  I like the idea of support for a more complex format, too, though for my purposes, just random sampling out of a multi-line file is sufficient :)

cheald:
Absolutely, implement however you'd like -- I'd just be happy to have it in the tool!

FurkanGozukara:
> I have updated dev branch to support multiline captions. I hope you could test it😀

awesome

so we just add multiple lines and they work right?

FurkanGozukara:
> > so we just add multiple lines and they work right?
> 
> Please enable it with `--enable_wildcard` option or `enable_wildcard = true` in .toml.
> 
> https://github.com/kohya-ss/sd-scripts/blob/dev/docs/config_README-en.md#multi-line-captions

--enable_wildcard got it
i never used toml so far :D

araleza:
I tried --enable_wildcard out for multiline captions. Seems to work when I tried adding a second caption to most of my face images.

And it works well for training on faces, as adding additional captions to each face pic makes it generalize the face better, making the learned face appear when inferencing new images with captions other than the ones in the training set.

The use of multiple captions also seems to make the training last for longer before the high-contrast appearance of overtraining sets in, although I'm not certain of this yet.

One thing is that I can't really see that it's enabled, other than the call to --enable_wildcard didn't get rejected. Maybe some debug line output like 'additional captions detected' would be useful if it's worked? It's not too important for this though, cause I'm pretty sure it's working.

FurkanGozukara:
hello again. what is the latest status of this? i want to today use this multi line feature. what do i need to do? write each line a different captions and it will randomly read? @kohya-ss @araleza @cheald

araleza:
> hello again. what is the latest status of this? i want to today use this multi line feature. what do i need to do? write each line a different captions and it will randomly read? @kohya-ss @araleza @cheald

It's available in the main branch. You just pass `--enable_wildcard` as a command line parameter.

Each line in your caption file then becomes its own caption, and captions will be chosen at random each time that image comes up in the training loop. Blank lines are ignored.

I got very good results from multi-line captions. Previously I thought my model had become overtrained/burned because I didn't have enough images, but adding extra captions has some of the same effect as adding entirely new images.

FurkanGozukara:
@araleza ty so much time to test it

recris:
@kohya-ss The `enable_wildcard` option seems to be ignored when caching text encoder outputs and training only UNet/DiT model.

I didn't find this mentioned anywhere in the documentation, but from looking at the code the method `process_caption` is never called if `ImageInfo` contains a previously cached `text_encoder_outputs`, which seems to happen in the above scenario.

Flux.1 is quite memory intensive, having cached TE outputs is critical for training in consumer hardware. I think it may be worth reworking the `ImageInfo` structure to accommodate multiple captions and their TE caches.

PR #1165, opened 2024-03-10
Validation loss again

gesen2egee:
Base on PR #914 Add validation loss by @rockerBOO 
使它可以在最新版上使用
而且減少code對原始功能的改變

修改了val_loss的算法

增加了兩個引數

--validation_every_n_step

### DIFF
--- a/library/config_util.py
+++ b/library/config_util.py
@@ -98,7 +98,8 @@ class BaseDatasetParams:
     resolution: Optional[Tuple[int, int]] = None
     network_multiplier: float = 1.0
     debug_dataset: bool = False
-
+    validation_seed: Optional[int] = None
+    validation_split: float = 0.0
 
 @dataclass
 class DreamBoothDatasetParams(BaseDatasetParams):
@@ -109,8 +110,7 @@ class DreamBoothDatasetParams(BaseDatasetParams):
     bucket_reso_steps: int = 64
     bucket_no_upscale: bool = False
     prior_loss_weight: float = 1.0
-
-
+    
 @dataclass
 class FineTuningDatasetParams(BaseDatasetParams):
     batch_size: int = 1
@@ -222,8 +222,11 @@ class ConfigSanitizer:
         "enable_bucket": bool,
         "max_bucket_reso": int,
         "min_bucket_reso": int,
+        "validation_seed": int,
+        "validation_split": float,        
         "resolution": functools.partial(__validate_and_convert_scalar_or_twodim.__func__, int),
         "network_multiplier": float,
+        
     }
 
     # options handled by argparse but not handled by user config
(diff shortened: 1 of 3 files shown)


DevArqSangoi:
Hey! I'd like to really thank you for this feature, it's absolutely incredible. However, I have a question that came up regarding the Text Encoder.

I understood the principle that, depending on the training, at some point it starts to overfit and the validation loss begins to increase. But what causes more problems in inference is the overfit in the Text Encoder, so much so that when I use my LoRAs, I reduce the weight during inference.

That said, I believe that the loss of the validation steps are also affected because the TE is frying/fried, while the Unet is not.

Is there a possibility to reduce the weight of the LoRA/Unet/Text-Encoder in these situations/steps/samples?

rockerBOO:
> That said, I believe that the loss of the validation steps are also affected because the TE is frying/fried, while the Unet is not.

Part of the issue is the TE is presented as cross attention to the UNet, so used in concert. Quantifying the TE would be a little more complex than just seeing the validation loss. Would need to have some comparison of the image CLIP value and the caption CLIP (TE) value and compare the similarity, possibly. I'm not sure how well this would ultimately line up or if it would be valuable to achieve though.

gesen2egee:
Adjusting is somewhat difficult for me because I'm unsure how to properly adjust the magnitude of TE and UNet for the different LoRA formats. Perhaps someone knowledgeable could provide a method.

gesen2egee:
> Not about validation but in relation to the TE overfit - the DB training method has the "Stop text encoder training" option which isn't available to TI and Lora. If it is technically possible to add that option to Lora training, may it be able to help with the TE overfit, perhaps? You could at least experiment by stopping TE training at different points.

It seems someone is working on this (PR #1011), but I'm not quite clear on the current progress.

DevArqSangoi:
I got it, I even realized some things in the trainings that indicate that an eventual adjustment in the LoRA weight wouldn't help with my 'problem'.

But now I'm going to bother you again, now with something more useful and related to the PR. 

I've come across a problem with the validation system automatically creating the dataset in question, because I can't use data augmentation along with the current validation system, as the same image, but with some subtle differences, could end up going to training and another to validation.

What would be the feasibility of, just like it's possible to define a 'reg' directory for regularization images, creating a 'val' for validation? That way, I would manually create the validation.

DevArqSangoi:
> > I got it, I even realized some things in the trainings that indicate that an eventual adjustment in the LoRA weight wouldn't help with my 'problem'.
> > But now I'm going to bother you again, now with something more useful and related to the PR.
> > I've come across a problem with the validation system automatically creating the dataset in question, because I can't use data augmentation along with the current validation system, as the same image, but with some subtle differences, could end up going to training and another to validation.
> > What would be the feasibility of, just like it's possible to define a 'reg' directory for regularization images, creating a 'val' for validation? That way, I would manually create the validation.
> 
> That's a good question. Previously, I hadn't considered the issue of data augmentation.
> 
> Perhaps you could place the data that will be augmented in the reg dataset, since the current val dataset will ignore the reg dataset.
> 
> Maybe I can create a specific exclusion condition for the val dataset dir name, to prevent it from entering the train dataset.
> 
> Otherwise, exclude the same image paths.

I think I'm gonna adopt a workaround with your suggestion of using the reg folder, but I'm gonna put all the training in the reg folder, and in the train folder I'll put the images I want to use for validation, then I'll set --validation_split=1. 🤔

gesen2egee:
> > > I got it, I even realized some things in the trainings that indicate that an eventual adjustment in the LoRA weight wouldn't help with my 'problem'.
> > > But now I'm going to bother you again, now with something more useful and related to the PR.
> > > I've come across a problem with the validation system automatically creating the dataset in question, because I can't use data augmentation along with the current validation system, as the same image, but with some subtle differences, could end up going to training and another to validation.
> > > What would be the feasibility of, just like it's possible to define a 'reg' directory for regularization images, creating a 'val' for validation? That way, I would manually create the validation.
> > 
> > 
> > That's a good question. Previously, I hadn't considered the issue of data augmentation.
> > Perhaps you could place the data that will be augmented in the reg dataset, since the current val dataset will ignore the reg dataset.
> > Maybe I can create a specific exclusion condition for the val dataset dir name, to prevent it from entering the train dataset.
> > Otherwise, exclude the same image paths.
> 
> I think I'm gonna adopt a workaround with your suggestion of using the reg folder, but I'm gonna put all the training in the reg folder, and in the train folder I'll put the images I want to use for validation, then I'll set --validation_split=1. 🤔

After re-reading the code, I realized that the same image paths being divided into training or validation shouldn't happen because the distribution of image paths is done before multiplying by num_repeats. However, I found that for the val_dataset, setting num_repeats=1 and canceling caption_dropout can make the val_loss calculation more consistent.

Additionally, one issue observed is that the val loss behaves abnormally after resuming training. I suspect this is due to the distribution of the val_dataset changing after resuming. (It's also possible that the resume function itself has issues. Who knows?)

rockerBOO:
> Additionally, one issue observed is that the val loss behaves abnormally after resuming training. I suspect this is due to the distribution of the val_dataset changing after resuming. (It's also possible that the resume function itself has issues. Who knows?)

Would make sure you're using the same seed. Validation also has it's own seed too which should keep it consistent (otherwise uses the seed for the run)

oceanus52:
> --validation_split=0.1 --validation_seed=1234 --validation_every_n_step=50 --max_validation_steps=5

image

the options seem to be:
 [--validation_seed VALIDATION_SEED] [--validation_split VALIDATION_SPLIT] [--validate_every_n_steps VALIDATE_EVERY_N_STEPS] [--validate_every_n_epochs VALIDATE_EVERY_N_EPOCHS] [--max_validation_steps MAX_VALIDATION_STEPS]
        
So it should be?                     
> --validation_split=0.1 --validation_seed=1234 --validate_every_n_step=50 --max_validation_steps=5

PR #1177, opened 2024-03-13
Random strength for Noise Offset and input perturbation noise

KohakuBlueleaf:
As title, I implemented a mechanism to apply noise offset/ip noise with random strength. (based on the strength value)

This is simple and could help model to learn more stably


masaint:
I'm interested in this concept, however is the random value truly 0-1? If so, have you tested it? Would be interested to see results.

Perhaps it could be improved if allowing a random range rather than truly random?

KohakuBlueleaf:
> I'm interested in this concept, however is the random value truly 0-1? If so, have you tested it? Would be interested to see results.
> 
> Perhaps it could be improved if allowing a random range rather than truly random?

image

And have you checked the code? it is random range. the upper bound is the strength set in the arguments.

PR #1178, opened 2024-03-13
[Experimental] Add cache mechanism for dataset groups to avoid long waiting time for initilization

KohakuBlueleaf:
For large scale dataset, sd-scripts will suffer from long waiting time to read the image size or other meta.

So I propose 2 improvements:
1. Cache the built dataset group object into disk so we don't need to calculate it multiple time
  * For ddp or some expriments across settings. This will be SUPER helpful
2. use `imagesize` library to read the image size, don't use PIL which is overkill.
  * This can provide 5~10 times speed up for reading image size on NVME ssd.

In my cache script, I successfully only use half hour to get the dataset groups. (Which will cost 4hour if I directly run 4card training)
And the loading for cached dataset groups also be fine. I have do a quick sanity check that first few images are same. But need more check from community

### DIFF
--- a/cache_dataset_meta.py
+++ b/cache_dataset_meta.py
@@ -1,0 +1,105 @@
+import argparse
+import random
+import pickle
+
+from accelerate.utils import set_seed
+
+import library.train_util as train_util
+import library.config_util as config_util
+from library.config_util import (
+    ConfigSanitizer,
+    BlueprintGenerator,
+)
+import library.custom_train_functions as custom_train_functions
+from library.utils import setup_logging, add_logging_arguments
+
+setup_logging()
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def make_dataset(args):
+    train_util.prepare_dataset_args(args, True)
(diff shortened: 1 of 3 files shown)


KohakuBlueleaf:
> Thank you for this! This seems to be really useful. I was thinking about how I can reduce the waiting time for preprocessing. I had not thought of the method for pickling of dataset group object.
> 
> However, I feel the pickling seems to be a bit aggressive. I wonder caching the sizes of images might be enough for reducing the waiting time...

I will implement a version which cache absolute path list and imagesize for each subset
So the loading procedure will be same, just ignore the listdir and imagesize part

KohakuBlueleaf:
@kohya-ss I have done the implementation
which only cache the image path/caption and image size.
With cached metadata, dataset with 32768 img only need 4sec from start the program to finish the dataset setup (include creating bucket)

I only implement it for DreamboothDataset at first.
But I think you can copy the implementation to another 2 dataset class easily.

PR #1189, opened 2024-03-16
Custom masked loss 

gesen2egee:
This PR is created from the masked-loss branch. This PR introduces enhancements to the masked loss functionality, 
inspired by the "Remove Near %" and "Remove Background %" features from the sd-webui-controlnet project. 

By applying thresholds to input depth maps, pixels below a certain threshold are binarized to 0 or 1, 
allowing the same depth map to have more variation and potentially improving the model's ability to focus on relevant features.

The remove_near functionality aims to turn values close to 1 into 1, while the remove_background functionality aims to turn values close to 0 into 0. It is anticipated that remove_near might allow the model to learn more about the background, whereas remove_background might facilitate learning more about the edges.

`--mask_remove_near`  Fixed threshold for near area removal.
`--mask_remove_background` Fixed threshold for background removal.
`--mask_random_remove_near`  Maximum random value for near area removal.
`--mask_random_remove_background`  Maximum random value for background removal.

Visual Illustration
image

Observations from Validation Loss: (purple is random 0.5 0.5)
image


recris:
This type of thing probably makes more sense as a separate preprocessing script to generate the mask files instead of something built into the training code.

There are a thousand different ways to create masks for a given image, depending on what we want to do, and we cannot possible handle all those use cases in the training scripts. Better to keep things simple and generic.

PR #1216, opened 2024-03-27
Rating support for WD Tagger

Disty0:
- Add rating support for WD Tagger.
   Inserts rating tag as the first tag.
 - Reorder character tags to be in front of the general tags.
   Characters can get lost into the token limit if it is at the end.
   AnimagineXL also uses this order.


sdbds:
Suggest inserting character tags in front as well

Disty0:
Added `--use_rating_tag`s and `--character_tags_first`.
Also did some formatting changes for arg parser part for consistency and did a few changes to make ruff happy that shouldn't affect the functionality.

Disty0:
I like the option `--character_tag_expand` and will use it on my dataset.

PR #1228, opened 2024-03-31
Add option to use Scheduled Huber Loss in all training pipelines to improve resilience to data corruption

kabachuha:
As heavily discussed in https://github.com/kohya-ss/sd-scripts/discussions/294, the presence even of a small number of outliers can heavily distort the image output.

While there have been proposed alternative losses, such as Huber loss, the problem was the loss of fine details when training the pictures. After researching the mathematical formulations of diffusion models, we came to the conclusion that it's possible to *schedule* the Huber loss, making it smoothly transition from Huber loss with L1 asymptotics on the first reverse-diffusion timesteps (when the image only begins to form and is most vulnerable to concept outliers) to the standard L2 MSE loss, for which diffusion models are originally formulated, on the last reverse-diffusion timesteps, when the fine-details of the image are forming.

Our method shows greater stability and resilience (similarity to clean pictures on corrupted runs - similarity to clean pictures on clean runs), than both pure Huber and L2 losses.

image

The experiments confirm that indeed this schedule improves the resilience greatly

image

Our paper: https://arxiv.org/abs/2403.16728

Diffusers discussion https://github.com/huggingface/diffusers/issues/7488

Most importantly, this approach has virtually no computational costs over the standard L2 computation. (and minimal code changes to the training scripts)

cc @cheald @kohya-ss

### DIFF
--- a/fine_tune.py
+++ b/fine_tune.py
@@ -354,7 +354,7 @@ def train(args):
 
                 if args.min_snr_gamma or args.scale_v_pred_loss_like_noise_pred or args.debiased_estimation_loss:
                     # do not mean over batch dimension for snr weight or scale v-pred loss
-                    loss = torch.nn.functional.mse_loss(noise_pred.float(), target.float(), reduction="none")
+                    loss = train_util.conditional_loss(noise_pred.float(), target.float(), reduction="none", loss_type=args.loss_type, huber_c=huber_c)
                     loss = loss.mean([1, 2, 3])
 
                     if args.min_snr_gamma:
@@ -366,7 +366,7 @@ def train(args):
 
                     loss = loss.mean()  # mean over batch dimension
                 else:
-                    loss = torch.nn.functional.mse_loss(noise_pred.float(), target.float(), reduction="mean")
+                    loss = train_util.conditional_loss(noise_pred.float(), target.float(), reduction="mean", loss_type=args.loss_type, huber_c=huber_c)
 
                 accelerator.backward(loss)
                 if accelerator.sync_gradients and args.max_grad_norm != 0.0:
--- a/fine_tune.py
+++ b/fine_tune.py
@@ -340,7 +340,7 @@ def train(args):
 
                 # Sample noise, sample a random timestep for each image, and add noise to the latents,
                 # with noise offset and/or multires noise if specified
-                noise, noisy_latents, timesteps = train_util.get_noise_noisy_latents_and_timesteps(args, noise_scheduler, latents)
+                noise, noisy_latents, timesteps, huber_c = train_util.get_noise_noisy_latents_and_timesteps(args, noise_scheduler, latents)
 
                 # Predict the noise residual
                 with accelerator.autocast():
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -3087,6 +3087,19 @@ def add_training_arguments(parser: argparse.ArgumentParser, support_dreambooth: 
         default=None,
         help="set maximum time step for U-Net training (1~1000, default is 1000) / U-Net学習時のtime stepの最大値を設定する（1~1000で指定、省略時はデフォルト値(1000)）",
     )
+    parser.add_argument(
+        "--loss_type",
+        type=str,
+        default="l2",
+        choices=["l2", "huber", "huber_scheduled"],
+        help="The type of loss to use and whether it's scheduled based on the timestep"
+    )
+    parser.add_argument(
+        "--huber_c",
+        type=float,
+        default=0.1,
+        help="The huber loss parameter. Only used if one of the huber loss modes is selected with loss_type.",
+    )
 
     parser.add_argument(
         "--lowram",
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -4591,6 +4591,35 @@ def save_sd_model_on_train_end_common(
         if args.huggingface_repo_id is not None:
             huggingface_util.upload(args, out_dir, "/" + model_name, force_sync_upload=True)
 
+def get_timesteps_and_huber_c(args, min_timestep, max_timestep, b_size, device):
+    timesteps = torch.randint(min_timestep, max_timestep, (b_size,), device=device)
+
+    #TODO: if a huber loss is selected, it will use constant timesteps for each batch
+    # as. In the future there may be a smarter way
+    if args.loss_type == 'huber_scheduled':
+        timesteps = torch.randint(
+            min_timestep, max_timestep, (1,), device='cpu'
+        )
+        timestep = timesteps.item()
+
+        alpha = - math.log(args.huber_c) / max_timestep
+        huber_c = math.exp(-alpha * timestep)
+        timesteps = timesteps.repeat(b_size).to(device)
+    elif args.loss_type == 'huber':
+        # for fairness in comparison 
+        timesteps = torch.randint(
+            min_timestep, max_timestep, (1,), device='cpu'
+        )
+        timesteps = timesteps.repeat(b_size).to(device)
+        huber_c = args.huber_c
+    elif args.loss_type == 'l2':
+        timesteps = torch.randint(min_timestep, max_timestep, (b_size,), device=device)
+        huber_c = 1 # may be anything, as it's not used
+    else:
+        raise NotImplementedError(f'Unknown loss type {args.loss_type}')
+    timesteps = timesteps.long()
+
+    return timesteps, huber_c
 
 def get_noise_noisy_latents_and_timesteps(args, noise_scheduler, latents):
     # Sample noise that we'll add to the latents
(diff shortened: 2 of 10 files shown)


kabachuha:
These are quite different approaches. @cheald makes adjustments to latents, then feeds them into the standard loss.

And as cheald is using mse_loss in the end, I believe these approaches can exist in synergy!

cheald:
This is very exciting! I've been experimenting with manipulating the mean/stddev of slices of the noise prior to forward noising, and have found that I can directly manipulate the level of detail trained with certain permutations of noise. However, it still suffers from outliers early in training having too large an impact, which I've had to manage through very careful tuning of noise. I'm very excited to try adding this into my experiments - if it performs how I imagine, then it could solve a number of problems which could lead to faster and more controlled training.

gesen2egee:
Tried a training, the results were impressive.

kabachuha:
@gesen2egee Thank you for giving a test run! Would you mind sharing some of them?

cheald:
Here's my quick training experiment. I just ran each for 9 epochs. Overall, results look very promising.

Ground truth: 

ComfyUI_00049_

General settings: adamw8bit, CosineAnnealingWarmRestarts w/ restart every 2 epochs, unet_lr 1e-4, and I _am_ using masks and masked loss here.

Each image pair is l2 on the left, huber_scheduled on the right. 

out1

Marginal improvement, I think!

----

Here's where it gets fun:

I'm experimenting with recentering and rescaling each noise channel individually, and then also experimenting with shifting and rescaling all noise channels together, as well as independently. This is conceptually similar to an expansion of the idea encoded in the `noise_offset` routine. It's yielding some very interesting results.

Dependent scale, dependent shift:

out2

Independent channel scaling, dependent channel shift:

out3

Dependent scaling, independent shift:

out4

and finally, independent scaling, independent shift:

out5

----

My general observation here is that the huber_scheduled loss definitely does improve detail retention, but isn't quite learning as well. However, I suspect that is likely due to nothing more than the lower loss values damping the rate of change in the learned weights. It might be that using huber_loss would permit a larger learning rate relative to l2 loss, which would be great if it can retain the same improvement in details.

cheald:
Here's another example. I'm working on extending the dynamic range of my trained samples, which I'm doing by pushing the mean of the first channel in the noise. I also cranked my LR up to 4e-4 to see if this would give the huber noise an edge, but it appears to actually not be working that way.

Same seed and training params for all 4 images, the left images are a "night time photo" and the right images are a "bright daytime photo". Upper is l2, lower is huber.

The huber examples have more detail, and it seems like a better dynamic range, but definitely feel more overtrained, which is interesting, since the nominal loss is higher for the l2 variants.

| Night L2 | Day L2 |
|-----|-----|
| Night Huber | Day Huber |

out

Loss curves for each:

feffy380:
Excited to try this out. Any idea how this interacts with Min-SNR-gamma, which weights the loss based on timestep?

drhead:
I'm looking forward to trying this out.  I do have a concern about part of the paper though, namely the function defining the schedule:
image
I don't think it makes much sense to define the schedule in terms of the indices of timesteps since they will vary between models, and even within SD1.5 there's zero terminal SNR which does modify the noise schedule.  If the reasoning is that loss should behave more like MAE at early timesteps/high noise levels and more like MSE at late timesteps/low noise levels, wouldn't it make more sense to define the schedule in terms of something like SNR directly?  From playing around in a notebook quickly, it seems that `1 - sqrt(1.0 - alphas_cumprod)` is fairly close (though I can't say with any certainty that it is the value that makes the most sense to use):
image
image

kabachuha:
@cheald @drhead we chose the exponential because of its simplicity, to test the claim that it should decrease with the (forward) diffusion timestep. It indeed quite likely may be suboptimal, and the idea about the snr-scheduling sounds very reasonable!

Adding some selectable huber_c schedules, I think, would be a great next addition

kabachuha:
There is another alternative for huber function: pseudo-Huber loss divided by delta:

$$\sqrt{\delta^2 + a^2} - \delta$$

whereas the "math" version is:

$$\delta^2*(\sqrt{1 + \frac{a^2}{\delta^2}} - 1)$$

The difference is math version converging to zero when the delta goes to zero, while the modified version transitions between L1 and L2

While the divided version gave us worse results for *resilience*, it may be better perceptually. (didn't analyze the perceptual part much, unfortunately) The former version, suggested by OpenAI, is used in Diffusers for LCM training.

Here's the gif comparison:

https://github.com/kohya-ss/sd-scripts/assets/14872007/38b95308-7765-42e8-87a9-40e1c3ab00ab

I think it may be worth adding it here too and making some more experiments

kabachuha:
@drhead the second main guy agreed that `1 / (1 + sigmas)**2` makes sense, so I think `delta = (1-delta_0) / (1 + sigmas)**2 + delta_0`, accounting for the zero-snr padding, may do a good gob. Will make some tests once I'll get home

kabachuha:
Good suggestion! Btw I'm making the experiments with different losses/schedules right now. Will post it here soon

kabachuha:
@drhead (on David Revoy dataset)
image

drhead:
Well, I've completed a small-ish ablation test of my own on a larger finetune of data that is *in theory* clean, and the results look pretty promising!

Here's regular MSE loss with my adaptive loss weights: 
huber_ablation_control

And here's with scheduled pseudo-huber loss:
huber_ablation_test

Backgrounds are much sharper, and most significantly the character's form is more consistent.  Very happy with how this turned out and I think I will continue to use this.

Interestingly, my loss weight model chose a different weighting (ignore the fact that it looks weird on the tails, I don't think this is a true reflection of the ideal timestep weighting and is instead an artifact of the model used to train timestep weights which I am still trying to nail down the correct hyperparameters for).  
loss_weights_20099_a2cade8c70c640af9afd
This effectively reflects the relative difficulty of each timestep (in form of the weighting required to equalize the model's capability at each timestep).  So, these changes indicate that scheduled pseudo huber loss made the high-noise timesteps relatively easier, and later timesteps relatively harder.  Again, my model is v-prediction, so the high noise timesteps are closer to x0 prediction and lower noise timesteps are closer to epsilon prediction -- you would probably see different results on epsilon prediction.   But I do find this to be an interesting result.

kabachuha:
> The default is exponential

I think it will be better to use 'snr' by default as it have been shown to have better quality

dill-shower:
Is there any information about when different scheduling method and loss function type should be applied? For example, for a large dataset one, for a small dataset another. For training on photos one, for training on anime artwork another

kabachuha:
@kohya-ss It's still "exponential" in the code itself

https://github.com/kohya-ss/sd-scripts/blob/dfa30790a99754f2f20e4980e91a000cce51d8a4/library/train_util.py#L3249

drhead:
I did want to follow up on this since some further experiments led me to notice some tradeoffs with the outcomes of training on this loss function.

While the benefits to image coherency still seem very clear, it seems that on my training setup, one of the side effects of this loss function is a loss of control over image contrast -- manifesting as things like "dark" in a prompt being less powerful on zero terminal SNR models.  Incidentally, [I have a very good idea of what the root cause might be](https://old.reddit.com/r/StableDiffusion/comments/1ag5h5s/the_vae_used_for_stable_diffusion_1x2x_and_other/), and it makes perfect sense given what this loss function does.

TL;DR of the above post is that the VAE used by SD 1.5 (and SD 2.1 and Midjourney and the Pixart series of models and DALL-E 3 and EDM2 and probably many others), tends to produce a high magnitude artifact, especially on large images and desaturated images (noting that desaturated in latent space is closer to sepia).  My theory is that the purpose of this artifact is to abuse the normalization layers of the model by placing an artifact that serves to desaturate the image by having a high magnitude that will cause other values to decrease significantly as they pass through the model normalization layers.  @madebyollin made some very helpful graphics demonstrating some of these effects: https://gist.github.com/madebyollin/ff6aeadf27b2edbc51d05d5f97a595d9

With this in mind, it does make perfect sense why this would happen, especially on my training setup.  I'm training a high resolution model (actually with several different resolution groups) with v-prediction.  With v-prediction, higher noise timesteps are closer to x_0 prediction, and the terminal timestep is outright x_0 prediction.  This means that the prediction target has something that is both high magnitude and extremely important for image reconstruction.  Errors are generally likely to be proportional to the magnitude of the target, which is not good when our loss function is more relaxed on large errors at this point.  The loss function can't take into account the fact that this will cause a massive error in pixel space -- for MSE loss, this would be less of a problem, since larger errors would be pulled in much harder.  While the loss objective does effectively ignore large outliers from the data in this way, the VAE artifact is sadly not separable from this.  It is possible that it could be mitigated by finding some metric for a latent that would indicate consistency in its saturation level, but the solution that makes more sense is to just move to a model that doesn't use the SD1.5 VAE.

As a final note, I have found the effects of this to be very similar to the effects of [FreeU](https://arxiv.org/abs/2309.11497), both in its improvements of image coherence and in having a similar issue with weakening of saturation (and they do combine together well, I was using it with my samples above).  FreeU's underlying theory is that high-frequency information in the skip connections converges too fast in the decoder and therefore the authors choose to modulate backbone features and skip connections to place more emphasis on low frequency feature maps.  That might be related to why this improves coherency -- mitigating the impact of outliers allowing for low frequency features to be better represented.  I'm sure there's a lot more research to be done on optimizing hierarchical denoising with things like this.

All of that being said, I do think that these issues are not likely to be a problem for epsilon prediction since the VAE artifact is not directly part of the prediction target, and it is not likely to be much of an issue on SDXL or for several other models using a different VAE or for pixel-space models.  I'll be trying it out on SD3 when I can, hopefully their VAE doesn't have the same issues.

Deathawaits4:
Can i ask which function im supposed to use to actually use scheduled huber loss in kohya? there is no explanation.. Do i need to set smooth_l1? or huber? What are the settings im supposed to use?

rockerBOO:
> Can i ask which function im supposed to use to actually use scheduled huber loss in kohya? there is no explanation.. Do i need to set smooth_l1? or huber? What are the settings im supposed to use?

There is some details explained on the main README [about-scheduled-huber-loss](https://github.com/kohya-ss/sd-scripts?tab=readme-ov-file#about-scheduled-huber-loss)

araleza:
Hey, I just tried out this new Huber loss parameter... and it's amazing.

Thanks for all the hard work by the people here!

DevArqSangoi:
What’s the practical effect of increasing or decreasing the value of huber_c?

DarkAlchy:
> What’s the practical effect of increasing or decreasing the value of huber_c?

My experience is total destruction of the training almost instantly.  Even a 0.01 in either direction resulted in total devastation/noise.

Issue #1123, opened 2024-02-16
wandb token exposed to wandb users


bghira:
hello, there exists a parameter to pass a wandb token into the training utility. unfortunately, this shows on the "Overview" tab of the training session, as a part of the complete commandline that is used to execute the trainer.

This exposes users to potentially nefarious activity, or at the very least, a sense of unease and the possibility of snooping/leaking internal data.

bghira:
@bmaltais and @kohya-ss should probably do something about this :D does this project not care about security?

bmaltais:
This is not good. I guess the Wanda token need to be exempted from making it in the metadata. This is something @kohya-ss need to filter in the training scripts. There is not much I can do about the metadata…

bghira:
need to recommend users to go with `wandb login` command instead of a cmdline argument. just `--report_to=wandb,tensorboard` would work.

bghira:
fwiw in Diffusers, there is simply a check whether wandb is enabled and the commandline args contain the Huggingface Hub token, because any secrets on cmdline will be exposed via wandb.

but in this case, its the wandb token itself on the cmdline. i am not sure that a documentation hint is enough to stop a user from doing this. the fact that it's on the cmdline at all means it will be exposed 100% of the time this is used.

bghira:
none that i have found. i resorted in simpletuner to using json config files to hide cmdline args

rafstahelin:
Would be great bowever to send parameters to Wanda. At the moment in kohya_ss, we don't see  any training parameters in Wandb, only the project name (tracker name), and the run name set from kohya gui. Is it a problem to send all parameters through config?

PR #1231, opened 2024-04-01
Expose training parameters to wandb.ai

VelocityRa:
This change sends the the training parameters as a payload to [wandb.ai](https://wandb.ai/) when that is used instead of TensorBoard.

They show up in the Runs table so it helps to get a better understanding of what the difference is between runs. It also uses the model's name as run name to give more context.

263689196-57506e88-1eda-477e-8b16-54adcb80f7e4

---

This is a re-implementation of #792.

### DIFF
--- a/fine_tune.py
+++ b/fine_tune.py
@@ -296,6 +296,9 @@ def train(args):
             init_kwargs["wandb"] = {"name": args.wandb_run_name}
         if args.log_tracker_config is not None:
             init_kwargs = toml.load(args.log_tracker_config)
+        else:
+            run_name = args.wandb_run_name if args.wandb_run_name else args.output_name
+            init_kwargs["wandb"] = {"name": run_name, "config": args}
         accelerator.init_trackers("finetuning" if args.log_tracker_name is None else args.log_tracker_name, init_kwargs=init_kwargs)
 
     # For --sample_at_first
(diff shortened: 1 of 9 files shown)


ccharest93:
This wouldn't affect normal workflow if not using wandb, that is because the extra configs are passed within the ['wandb'] key of the configs. This also is not a breaking change for wandb users since the extra arguments can just be ignored on the wandb side. \
However i agree that some filtering is necessary, can i revise this PR to be in line with #1123 ?

rafstahelin:
> This change sends all the training parameters to [wandb.ai](https://wandb.ai/) when that is used instead of TensorBoard (by default ie. if `log_tracker_config` is not specified).
> 
> Now they show up in the Runs table, so it helps to get a better understanding of what the difference is between runs. It also uses the model's name as run name to give more context.
> 
> 263689196-57506e88-1eda-477e-8b16-54adcb80f7e4
> This is a re-implementation of #792.

HI
Seems like SDXL_Train.py has changed. Could you please advise how to change this code so it works again? 

This is what I am seeing in Kohya_ss >> sdxl_train.py
image

```
if accelerator.is_main_process:
        init_kwargs = {}
        if args.wandb_run_name:
            init_kwargs["wandb"] = {"name": args.wandb_run_name, "config": args}
        if args.log_tracker_config is not None:
            init_kwargs = toml.load(args.log_tracker_config)
        accelerator.init_trackers("finetuning" if args.log_tracker_name is None else args.log_tracker_name, init_kwargs=init_kwargs)
```

Unfortunately, that code `"config": args` isn't sending args to wandb

thank you

ccharest93:
@rafstahelin  check out #1285

PR #1233, opened 2024-04-01
Add LoRA+ support

rockerBOO:
LoRA+ adds a hyperparameter to the optimizer for B or up in the LoRA to increase the ratio of that LR.
Screenshot 2024-03-25 at 13-55-27 2402 12354 pdf

Screenshot 2024-03-25 at 16-20-08 2402 12354 pdf

The hyperparameter should be adjusted depending on your dataset and  how difficult it is to train. They suggest starting at 2^4 or 16 ratio to start.

The Text Encoder and UNet are currently separate ratios as with recommended starting values as follows. 

`--loraplus_unet_lr_ratio=16`
`--loraplus_text_encoder_lr_ratio=16`

I am willing to change how the code is implemented, name of arguments and so on as necessary. 

https://arxiv.org/abs/2402.12354
https://github.com/nikhil-ghosh-berkeley/loraplus

### DIFF
--- a/networks/lora.py
+++ b/networks/lora.py
@@ -1035,21 +1035,43 @@ class LoRANetwork(torch.nn.Module):
         return lr_weight
 
     # 二つのText Encoderに別々の学習率を設定できるようにするといいかも
-    def prepare_optimizer_params(self, text_encoder_lr, unet_lr, default_lr):
+    def prepare_optimizer_params(self, text_encoder_lr, unet_lr, default_lr, unet_lora_plus_ratio=None, text_encoder_lora_plus_ratio=None):
         self.requires_grad_(True)
         all_params = []
 
-        def enumerate_params(loras):
-            params = []
+        def assemble_params(loras, lr, lora_plus_ratio):
+            param_groups = {"lora": {}, "plus": {}}
             for lora in loras:
-                params.extend(lora.parameters())
+                for name, param in lora.named_parameters():
+                    if lora_plus_ratio is not None and "lora_up" in name:
+                        param_groups["plus"][f"{lora.lora_name}.{name}"] = param
+                    else:
+                        param_groups["lora"][f"{lora.lora_name}.{name}"] = param
+
+            # assigned_param_groups = ""
+            # for group in param_groups:
+            #     assigned_param_groups += f"{group}\n {list(param_groups[group].keys())}\n\n"
+            # logger.info(assigned_param_groups)
+
+            params = []
+            for key in param_groups.keys():
+                param_data = {"params": param_groups[key].values()}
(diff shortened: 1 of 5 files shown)


feffy380:
> LoRA+ adds a hyperparameter to the optimizer for B or up in the LoRA to increase the ratio of that LR.

B is `lora_down`, not up. See the paper where B is defined as $B \in \mathbb{R}^{n_1 \times r}$ for $W \in \mathbb{R}^{n_1 \times n_2}$, which corresponds to `lora_down = Linear(in_dim, lora_dim)` in the codebase.

feffy380:
LoRA+ doesn't make sense for LoRA-FA where one of the matrices is already frozen

rockerBOO:
> LoRA+ doesn't make sense for LoRA-FA where one of the matrices is already frozen

My inclination was to replicate the behavior for all the networks to make it consistent. Since it leaves B to be trainable I think it makes sense to still be able to utilize this same behavior for it. I think it will be less valuable for LoRA-FA but I don't think it's entirely wrong to support. 

Do you think this assessment is ok? Ideally all the networks use the same prepare_optimizer_params to reduce code duplication in the future so I think this is ok?

networks/lora.py
```diff
@@ -1063,21 +1085,15 @@ def enumerate_params(loras):
...
+                        params = assemble_params(block_loras, default_lr * self.get_lr_weight(block_loras[0]), unet_lora_plus_ratio)
+                    all_params.extend(params)
 
             else:
-                param_data = {"params": enumerate_params(self.unet_loras)}
-                if unet_lr is not None:
-                    param_data["lr"] = unet_lr
-                all_params.append(param_data)
+                params = assemble_params(self.unet_loras, unet_lr, unet_lora_plus_ratio)
+                all_params.extend(params)
```
feffy380, line 1096: Found a bug: If you set the learning rate using `--learning_rate`, `assemble_params` will receive `None` from `unet_lr` and LoRA+ won't be used. Need to pass `default_lr` if `unet_lr == None`. The same applies to L1073 for the text encoder

rockerBOO:
Bugs:
- Default LR not applied 
- Logging LR shows incorrect labels (group 1 and group 2 will get labeled like lr/text_encoder and lr/unet)

rockerBOO:
Made some fixes, including Default LR and invalid LR now properly working. Should now properly support single param group, only TE and only UNet and all LoRA+ and LR permutations. 

The following test script I made to make sure it is working as expected. Willing to fix anything you find incorrect.

https://gist.github.com/rockerBOO/7e58b268d266c6f55a7f90085a1d1dfe

Issue #1269, opened 2024-04-17
Some RGBA images cause broken backgrounds due to incorrect RGB conversion


city96:
Currently, the `load_image` function in `train_utils` uses the PIL `image.convert("RGB")` method for converting the loaded 4 RGBA channel images to the expected 3 channel RGB format.

This is incorrect for some edge cases, and causes the background to become distorted.

As an example, loading [this RGBA image](https://img3.gelbooru.com/images/b6/63/b663ee70d6ef873c1c97fee64079dda6.png) results in [heavy background artifacts.](https://github.com/kohya-ss/sd-scripts/assets/125218114/0518ec51-c26e-4122-8ca9-a9f8e71eac46) when used with the test code below:

```py
from PIL import Image
from library.train_util import load_image

img = load_image("test.png")
Image.fromarray(img, 'RGB').save("test_convert.png")
```

To verify this issue, a LoRA was created using that single image as the dataset, using a caption that includes the words "white background". Seed and everything else was fixed between the two tests.

The top row is the latest code using a fresh VENV. The bottom row is the proposed fix, which I will create a PR for shortly. It uses the `Image.alpha_composite` function with a new blank white image to handle the alpha channel.

![image](https://github.com/kohya-ss/sd-scripts/assets/125218114/9a8b3114-a004-4fc0-ae01-ed513d51a009)

feffy380:
They're not artifacts. Those colors really exist in the png, they're just fully transparent. Dropping the alpha channel makes them visible again. Most image editors automatically set transparent pixels to black but for some reason the artist disabled that.
![image](https://github.com/kohya-ss/sd-scripts/assets/114889020/0010b216-0d71-433a-95df-3dcac0775e2a)

city96:
@feffy380 Interesting, seems like you're right. I kind-of just assumed it was some funky conversion logic but I guess that explains it.

I've [created a quick script](https://gist.github.com/city96/2db92a207f3f93b1170dbd412688f812) to check for hidden/corrupted backgrounds.

Currently testing it against my smaller 300K dataset to see what comes up. The current hit rate is 101 out of ~10000 images, only considering PNGs, so seems pretty rare.

city96:
@kohya-ss Thank you. That solution is indeed more optimal, I'll close my PR and will follow the one you linked instead.
Leaving the issue open until that PR gets merged in case someone else runs into this same behavior.

PR #1270, opened 2024-04-17
Fix RGBA Image loading

city96:
Proposed fix for issue #1269 

Using `Image.convert("RGB")` results in broken backgrounds for some images, this PR changes the logic for RGBA images to use `Image.alpha_composite` with a new empty image instead. See issue above for details.

The color chosen for the alpha fill in this PR is pure white, which is different from what the old implementation defaulted to (pure black), since it seemed like a better fit for most images. This can be changed if required.


city96:
Closing if favor of #1223

PR #1278, opened 2024-04-19
Display name of error latent file

Cauldrath:
When trying to load stored latents, if an error occurs, this change will tell you what file failed to load
Currently it will just tell you that something failed without telling you which file

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -2135,8 +2135,10 @@ def is_disk_cached_latents_is_expected(reso, npz_path: str, flip_aug: bool):
                 return False
             if npz["latents_flipped"].shape[1:3] != expected_latents_size:
                 return False
-    except:
-        raise RuntimeError(f"Error loading file: {npz_path}")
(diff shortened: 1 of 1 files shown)


feffy380:
IMO instead of raising an exception and crashing, it should log the error and return False so the latent is immediately regenerated.

Cauldrath:
> IMO instead of raising an exception and crashing, it should log the error and return False so the latent is immediately regenerated. Otherwise the user has to manually delete the corrupted .npz file and re-run the script each time.

Good idea. It will now give the file name and the error message, then return false so the file is regenerated without the user needing to do anything.

Cauldrath:
The error will still be displayed, but it is likely going to be an issue of a corrupt npz file (I encountered this issue when I ran out of hard drive space while caching latents, so the file wasn't completed), which will be resolved by regenerating it. If it isn't a case where the error is resolved by regenerating, the user can still look through the logs for the error.

PR #1279, opened 2024-04-19
Catch model saving error

Cauldrath:
If an error is thrown when trying to save a checkpoint, it will give the user the error and ask if they would like to save again.
This helps with cases where something goes wrong at the end of an hours-long training (like if you didn't have enough disk space)
The code was also rearranged so that the epochs and steps parameters do not get altered and new_ckpt is only created if it is used

### DIFF
--- a/library/sdxl_model_util.py
+++ b/library/sdxl_model_util.py
@@ -469,6 +469,29 @@ def convert_text_encoder_2_state_dict_to_sdxl(checkpoint, logit_scale):
 
     return new_sd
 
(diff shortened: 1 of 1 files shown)


Cauldrath:
I updated all of the model-saving logic to funnel into a single function that has the try/catch and consolidated the logic for them as much as possible while keeping all the functionality the same. There is probably some potential to consolidate more if we change some of the differing logic to match.

Includes a merge of this PR: https://github.com/kohya-ss/sd-scripts/pull/1371

PR #1284, opened 2024-04-20
Fix train controlnet

sdbds:
Now we can train controlnet for SD 1.5.
Sample image part not change.

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -1982,8 +1982,8 @@ class ControlNetDataset(BaseDataset):
         self.bucket_manager = self.dreambooth_dataset_delegate.bucket_manager
         self.buckets_indices = self.dreambooth_dataset_delegate.buckets_indices
 
-    def cache_latents(self, vae, vae_batch_size=1, cache_to_disk=False, is_main_process=True):
-        return self.dreambooth_dataset_delegate.cache_latents(vae, vae_batch_size, cache_to_disk, is_main_process)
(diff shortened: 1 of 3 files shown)


sdbds:
> Thank you for this! Is `omegaconf` needed? I'd like to minimize the dependencies...

I think just having a method that wraps the config would be fine, but diffusers don't support the old SimpleNamespace anymore because SimpleNamespace doesn't support loops.

sdbds:
> > I think just having a method that wraps the config would be fine, but diffusers don't support the old SimpleNamespace anymore because SimpleNamespace doesn't support loops.
> 
> Thank you for clarification. It's annoying...
> 
> The following code may be working. Is it fine to use this? (I'm testing this PR on my env, but I get another error currently.)
> 
> ```python
>     # make unet.config iterable and accessible by attribute
>     class CustomConfig:
>         def __init__(self, **kwargs):
>             self.__dict__.update(kwargs)
> 
>         def __getattr__(self, name):
>             if name in self.__dict__:
>                 return self.__dict__[name]
>             else:
>                 raise AttributeError(f"'{self.__class__.__name__}' object has no attribute '{name}'")
> 
>         def __contains__(self, name):
>             return name in self.__dict__    
>         
>     unet.config = CustomConfig(**unet.config)
> ```
it seems works in my env.
2024-05-14 20-13-16 的屏幕截图

PR #1285, opened 2024-04-20
Hyperparameter tracking

ccharest93:
Sensitive parameters are not really DRY since they are repeated in verify_command_line_training_args, but for now only two repeats is not too bad. 

The setting of hyperparameters through config is handled by Accelerate differently for each tracker so the changes are not breaking. 

Tested on Wandb and tensorboard, they both work. More testing on tensorboard is welcomed as i mostly use wandb.

Related PR
#792 
#959 @rockerBOO 
#1231 

@kohya-ss  Lets push this through hyperparameter tracking is very useful when designing large scale experiments.

### DIFF
--- a/fine_tune.py
+++ b/fine_tune.py
@@ -310,7 +310,7 @@ def train(args):
             init_kwargs["wandb"] = {"name": args.wandb_run_name}
         if args.log_tracker_config is not None:
             init_kwargs = toml.load(args.log_tracker_config)
-        accelerator.init_trackers("finetuning" if args.log_tracker_name is None else args.log_tracker_name, init_kwargs=init_kwargs)
+        accelerator.init_trackers("finetuning" if args.log_tracker_name is None else args.log_tracker_name, config=train_util.filter_sensitive_args(args), init_kwargs=init_kwargs)
 
     # For --sample_at_first
     train_util.sample_images(accelerator, args, 0, global_step, accelerator.device, vae, tokenizer, text_encoder, unet)
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -3378,6 +3378,20 @@ def add_masked_loss_arguments(parser: argparse.ArgumentParser):
         help="apply mask for calculating loss. conditioning_data_dir is required for dataset. / 損失計算時にマスクを適用する。datasetにはconditioning_data_dirが必要",
     )
 
+def filter_sensitive_args(args: argparse.Namespace):
+    sensitive_args = ["wandb_api_key", "huggingface_token"]
+    sensitive_path_args = [
+        "pretrained_model_name_or_path",
+        "vae",
+        "tokenizer_cache_dir",
+        "train_data_dir",
+        "conditioning_data_dir",
+        "reg_data_dir",
+        "output_dir",
+        "logging_dir",
+    ]
+    filtered_args = {k: v for k, v in vars(args).items() if k not in sensitive_args + sensitive_path_args}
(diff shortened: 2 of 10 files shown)


gshawn3:
Looks great. Thank you for implementing this! I was having to take care of it through an additional script.

rafstahelin:
> Looks great. Thank you for implementing this! I was having to take care of it through an additional script.

@gshawn3 
Hey does this mean by applying the code changes above passing args will work?

gshawn3:
@rafstahelin yep, the hyperparameters will be passed as wandb config parameters for the training run and can then be used in wandb to filter, compare runs, etc. IMHO the current wandb integration is not very useful without tracking these values, which is why this PR is sorely needed. Hope this helps.

rafstahelin:
> @rafstahelin yep, the hyperparameters will be passed as wandb config parameters for the training run and can then be used in wandb to filter, compare runs, etc. IMHO the current wandb integration is not very useful without tracking the run hyperparameters, which is why this PR is sorely needed. Hope this helps.

yup can confirm it works. 
Until a merge, if there is an update, what's the best way to maintain this code when i git pull kohya_ss? Do i have to keep these bits of code to repurpose if updating the scripts is necessary? Sorry, I am by no means very ml literate

gshawn3:
Until this PR gets merged by @kohya-ss, when there are changes in the main branch you would just need to merge those changes back into your local branch.

Might be easier to understand if I just list the git commands:


```
# initially, check out this PR into a local branch
git fetch origin pull/1285/head:hyperparams
git checkout hyperparams
```


```
# whenever kohya pushes updates to the main branch, merge those changes back into your branch
git switch main
git pull
git switch hyperparams
git merge main
```

That should do it!

rafstahelin:
@gshawn3 thanks so much, much appreciated. This will help me

rafstahelin:
@gshawn3 
i use kohya_ss gui. Can i run these git commandsd in the sd-scripts of the kohya_ss parent local install?

gshawn3:
@rafstahelin it sounds like your issue might require some more specific assistance. I’m happy to help out, but could you please open a separate Discussion in bmaltais' repo with details, so we can follow up there? This way we can avoid derailing the PR conversation and keep this thread focused on reviewing the proposed changes.

In the meantime, it looks like this should help: https://github.com/bmaltais/kohya_ss/blob/master/examples/pull%20kohya_ss%20sd-scripts%20updates%20in.md

rafstahelin:
@gshawn3 thanks for the helpful link. I'll give this a try. And need be will open a thread at kohya_ss.. thanks again

rafstahelin:
> @rafstahelin it sounds like your issue might require some more specific assistance. I’m happy to help out, but could you please open a separate Discussion in bmaltais' repo with details, so we can follow up there? This way we can avoid derailing the PR conversation and keep this thread focused on reviewing the proposed changes.
> 
> In the meantime, it looks like this should help: https://github.com/bmaltais/kohya_ss/blob/master/examples/pull%20kohya_ss%20sd-scripts%20updates%20in.md

hi @gshawn3 
I tried your instructions. Please see 
https://github.com/bmaltais/kohya_ss/issues/2482

thanks

ccharest93:
No worries and thanks for all the work you do on sd-scripts! Yes this replaces each of the PR in my initial comment, #959 was much better written than #1231 so i pulled more heavily on that code and implement a sensitive information filter as per your request in #1231.

Issue #1290, opened 2024-04-23
UnboundLocalError occurred when not using onnx and with --max_data_loader_n_workers


frodo821:
I have been experiencing an issue with `tag_images_by_wd14_tagger.py`.
I ran `finetune/tag_images_by_wd14_tagger.py` by using the options shown below:

```bash
$ accelerate launch "sd-scripts/finetune/tag_images_by_wd14_tagger.py" \
>   --batch_size=8 --general_threshold=0.35 --character_threshold=0.35 \
>   --caption_extension=".txt" --caption_separator="," \
>   --model="SmilingWolf/wd-v1-4-convnextv2-tagger-v2" \
>   --max_data_loader_n_workers="2" --debug --remove_underscore \
>   --frequency_tags "data/images"
```

Then I got this UnboundLocalError:

```
Traceback (most recent call last):
  File "/workspace/sd-scripts/finetune/tag_images_by_wd14_tagger.py", line 514, in <module>
    main(args)
  File "/workspace/sd-scripts/finetune/tag_images_by_wd14_tagger.py", line 333, in main
    data = torch.utils.data.DataLoader(
UnboundLocalError: local variable 'torch' referenced before assignment
Traceback (most recent call last):
  File "/opt/conda/bin/accelerate", line 8, in <module>
    sys.exit(main())
  File "/opt/conda/lib/python3.10/site-packages/accelerate/commands/accelerate_cli.py", line 46, in main
    args.func(args)
  File "/opt/conda/lib/python3.10/site-packages/accelerate/commands/launch.py", line 1075, in launch_command
    simple_launcher(args)
  File "/opt/conda/lib/python3.10/site-packages/accelerate/commands/launch.py", line 681, in simple_launcher
    raise subprocess.CalledProcessError(returncode=process.returncode, cmd=cmd)
```

I have investigated why the error occurs, and I noticed an unnecessary extra import of torch is made here:

https://github.com/kohya-ss/sd-scripts/blob/71e2c91330a9d866ec05cdd10584bbb962896a99/finetune/tag_images_by_wd14_tagger.py#L113-L116

I suspect this extra import statement makes `torch` a local, unbound variable, and it is only available when run with the `--onnx` option. 

`torch` is already imported this line:

https://github.com/kohya-ss/sd-scripts/blob/71e2c91330a9d866ec05cdd10584bbb962896a99/finetune/tag_images_by_wd14_tagger.py#L8

And the import statement on line 115 must be removed.

PR #1291, opened 2024-04-23
removed unnecessary `torch` import on line 115

frodo821:
as per #1290


frodo821:
@sdbds You may not need to import `torch` there because `torch` is already imported at line 8.
If you do `import torch` inside a function, the Python interpreter regards `torch` as a local yet unbound variable until the import statement actually runs.

frodo821:
Or we do have another choice to add a `global torch` (or a `nonlocal torch` for this case) statement at the beginning of the function definition.

frodo821:
@kohya-ss thank you. since this PR has been merged, I will close #1290.

PR #1353, opened 2024-05-31
Resume correct step for "resume from state" feature.

KohakuBlueleaf:
The corrected version of `resume-step-and-epoch` branch.
This one have been used on Kohaku-XL Epsilon so I think it is working.
I also ensured the batch order are exactly same before/after resume.

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -657,8 +657,15 @@ class BaseDataset(torch.utils.data.Dataset):
 
     def set_current_epoch(self, epoch):
         if not self.current_epoch == epoch:  # epochが切り替わったらバケツをシャッフルする
-            self.shuffle_buckets()
-        self.current_epoch = epoch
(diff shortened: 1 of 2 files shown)


KohakuBlueleaf:
@kohya-ss 
This one is only for train_network scripts.
I think you may want to copy this to other scripts too.

KohakuBlueleaf:
Some detail:
1. Since the first batch of the dataset is super weird (also shown in non-resumed training, I think it could be kind of bug of dataloader or dataset implementation in sd-scripts) So I skip to initial-1 step, than load first batch but drop it, to ensure the batch sent to training loop is what we expected.
2. @kohya-ss you have some improvement in that branch but is not related to this topic so I didn't copy it. (use "with" statement for image loading) I guess you may want to push that into dev branch directly?

PR #1359, opened 2024-06-04
Train resume step

kohya-ss:
#1353 and following updates:

- Set global step to stop the training when step >= global step, and track the step in the state.
- Fix LossRecorder to work with skipping.
- Set the epoch to start to make `initial_step` less than `len(train_dataloader)`.
- Fix to work resuming with `--skip_until_initial_step`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,12 @@ The majority of scripts is licensed under ASL 2.0 (including codes from Diffuser
 
 - The ControlNet training script `train_controlnet.py` for SD1.5/2.x was not working, but it has been fixed. PR [#1284](https://github.com/kohya-ss/sd-scripts/pull/1284) Thanks to sdbds!
 
+- `train_network.py` and `sdxl_train_network.py` now restore the order/position of data loading from DataSet when resuming training. PR [#1353](https://github.com/kohya-ss/sd-scripts/pull/1353) [#1359](https://github.com/kohya-ss/sd-scripts/pull/1359) Thanks to KohakuBlueleaf!
+  - This resolves the issue where the order of data loading from DataSet changes when resuming training.
+  - Specify the `--skip_until_initial_step` option to skip data loading until the specified step. If not specified, data loading starts from the beginning of the DataSet (same as before).
+  - If `--resume` is specified, the step saved in the state is used.
+  - Specify the `--initial_step` or `--initial_epoch` option to skip data loading until the specified step or epoch. Use these options in conjunction with `--skip_until_initial_step`. These options can be used without `--resume` (use them when resuming training with `--network_weights`).
+
 - An option `--disable_mmap_load_safetensors` is added to disable memory mapping when loading the model's .safetensors in SDXL. PR [#1266](https://github.com/kohya-ss/sd-scripts/pull/1266) Thanks to Zovjsra!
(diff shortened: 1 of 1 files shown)


KohakuBlueleaf (approved):
All LGTM

KohakuBlueleaf:
Tiny question:
Is there any reason that the default behaviour is "skip to state's epoch" instead of "skip to state's step"?

KohakuBlueleaf:
> > Tiny question: Is there any reason that the default behaviour is "skip to state's epoch" instead of "skip to state's step"?
> 
> This is for backward compatibility. If `--skip_until_initial_step` is not set, only epoch no is skipped. This doesn't change the actual training, so the training is same as before (the dataloader is reset). I know this is not good behavior, but there is the option `skip_until_initial_step` now, so I chose backward compatibility over the spec changes.

No problem UwU

KohakuBlueleaf:
I think this PR is ready to Merge
(And lot of users are waiting it XD)

Jannchie:
Is there a chance to implement this in sdxl_train or other tasks?

Pevernow:
@kohya-ss 
image
I use the code and resume function on the master branch. As shown in the picture, resume should have been loaded normally, but each training starts from the beginning. How can I solve this problem?

KohakuBlueleaf:
> @kohya-ss image I use the resume function on the master branch. As shown in the picture, resume should have been loaded normally, but each training starts from the beginning. How can I solve this problem?

Have you used `--skip_until_initial_step` ?

Pevernow:
> > @kohya-ss image I use the resume function on the master branch. As shown in the picture, resume should have been loaded normally, but each training starts from the beginning. How can I solve this problem?
> 
> Have you used `--skip_until_initial_step` ?

No.
Is this option connected with the resume function?

KohakuBlueleaf:
> > > @kohya-ss image I use the resume function on the master branch. As shown in the picture, resume should have been loaded normally, but each training starts from the beginning. How can I solve this problem?
> > 
> > 
> > Have you used `--skip_until_initial_step` ?
> 
> No. Is this option connected with the resume function?

 Check the conversation above……
You need to enable this argument to continue the train loader as well

6DammK9:
> Is there a chance to implement this in sdxl_train or other tasks?

Bump. I see this option should be available in `sdxl_train` also.
In theory it is ` accelerator.skip_first_batches` [ref](https://huggingface.co/docs/accelerate/usage_guides/checkpoint), [ref](https://huggingface.co/docs/accelerate/v1.3.0/en/package_reference/accelerator#accelerate.Accelerator.skip_first_batches) and calculate the `initial_step` along with a static seed `--seed` to make it effective (it wlil reflected in log metric also).

With [this major "push to main"](https://github.com/kohya-ss/sd-scripts/pull/1879) I'll look for the adaption. 

However it may not work with multi epoch, see [set_current_epoch](https://github.com/kohya-ss/sd-scripts/blob/sd3/library/train_util.py#L761). If the dataset is not so large, maybe start from a fresh epoch, [and then edit the log metric](https://www.tensorflow.org/tensorboard/dataframe_api). My case is "1EP under weeks" ~~danbooru+e621~~ so I need this for any process failure (e.g. NCCL error / Linux kernel / PSU blown etc.)

PR #1363, opened 2024-06-05
Add DDP for controlnetLLLite

sdbds:
Prevent multi-card training from reporting errors.

### DIFF
--- a/sdxl_train_control_net_lllite.py
+++ b/sdxl_train_control_net_lllite.py
@@ -289,6 +289,8 @@ def train(args):
     # acceleratorがなんかよろしくやってくれるらしい
     unet, optimizer, train_dataloader, lr_scheduler = accelerator.prepare(unet, optimizer, train_dataloader, lr_scheduler)
 
+    unet = unet.module if isinstance(unet, DDP) else unet
+
     if args.gradient_checkpointing:
         unet.train()  # according to TI example in Diffusers, train is required -> これオリジナルのU-Netしたので本当は外せる
     else:


sdbds:
> Thank you for this. However, because this unwraps the model from DDP wrapper, so I wonder the gradient might not be accumulated from distributed GPUs. In addition, this also unwraps Accelerate's wrapper, so I also wonder it may break other distribution strategies and the mixed precision wrapper.
> 
> Without this update, what kind of the error occur?

I find this solve from train_controlnet.py and it works.

Original error:
`RuntimeError: Expected to have finished reduction in the prior iteration before starting a new one. This error indicates that your module has parameters that were not used in producing loss. You can enable unused parameter detection by (1) passing the keyword argument find_unused_parameters=True to torch.nn.parallel.DistributedDataParallel; (2) making sure all forward function outputs participate in calculating loss. If you already have done the above two steps, then the distributed data parallel module wasn't able to locate the output tensors in the return value of your module's forward function. Please include the loss function and the structure of the return value of forward of your module when reporting this issue (e.g. list, dict, iterable).`

zixuzhuang:
Hi, I test the newest dev version, but the lllite still can not train with both ddp and gradient checkpointing. 
I miss this issue when  accelerator.backward(loss):
 RuntimeError: expect_autograd_hooks_ INTERNAL ASSERT FAILED at "../torch/csrc/distributed/c10d/reducer.cpp":1591, please report a bug to PyTorch.

PR #1371, opened 2024-06-12
Use model_util.is_safetensors everywhere

Cauldrath:
Instead of having a copy of the same code everywhere, use the same model_util function across the project

Function in question: https://github.com/kohya-ss/sd-scripts/blob/56bb81c9e6483b8b4d5b83639548855b8359f4b4/library/model_util.py#L963

In preparation for consolidating the model saving code so that this PR can be applied across all model types without copying the code to every one of them individually: https://github.com/kohya-ss/sd-scripts/pull/1279


Cauldrath:
Because the original PR that uses this code change (https://github.com/kohya-ss/sd-scripts/pull/1279) imports other functions from that file, I updated that one to remove any cases where it only imports is_safetensors and am closing this one.

Issue #1288, opened 2024-04-22
cache_text_encoder_outputs.py raises AttributeError: 'Namespace' object has no attribute 'deepspeed'


deepdelirious:
Seeing this on the main branch:

Traceback (most recent call last):
  File "/home/deli/images/sd-scripts/tools/cache_text_encoder_outputs.py", line 194, in <module>
    cache_to_disk(args)
  File "/home/deli/images/sd-scripts/tools/cache_text_encoder_outputs.py", line 102, in cache_to_disk
    accelerator = train_util.prepare_accelerator(args)
  File "/home/deli/images/sd-scripts/library/train_util.py", line 4303, in prepare_accelerator
    deepspeed_plugin = deepspeed_utils.prepare_deepspeed_plugin(args)
  File "/home/deli/images/sd-scripts/library/deepspeed_utils.py", line 71, in prepare_deepspeed_plugin
    if not args.deepspeed:
AttributeError: 'Namespace' object has no attribute 'deepspeed'

deepdelirious:
FYI @BootsofLagrangian

BootsofLagrangian:
It is because `cache_text_encoder_outputs.py` does not prepare deepspeed config not like `train_.py`

you can add ad-hoc for this

1) from library import deepspeed_utils

2) in line between 174-178,

```
      train_util.add_sd_models_arguments(parser)
      train_util.add_training_arguments(parser, True)
      train_util.add_dataset_arguments(parser, True, True, True)
      config_util.add_config_arguments(parser)
      sdxl_train_util.add_sdxl_training_arguments(parser)
```
to

```
      train_util.add_sd_models_arguments(parser)
      train_util.add_training_arguments(parser, True)
      train_util.add_dataset_arguments(parser, True, True, True)
+    deepspeed_utils.prepare_deepspeed_args(args)
      config_util.add_config_arguments(parser)
      sdxl_train_util.add_sdxl_training_arguments(parser)
```

noisefloordev:
Same problem in cache_latents.py.  (I thought this "Deepspeed" stuff was going to be in a separate branch since it was hard to maintain...)

PR #1372, opened 2024-06-14
Validate text encoder cache + add deepspeed arg parsing

heyalexchoi:
Changes:
- Validates text encoder cache, similarly to how image latents are validated. Previously only checked existence of file, but I had issue with incomplete or corrupted file breaking my training 600 steps in.
- Fixes deepspeed arg parsing in cache latents scripts, related to https://github.com/kohya-ss/sd-scripts/issues/1288 
- Adds `--skip_to_step` arg to sdxl_train. I am somewhat new to this project so not sure if this is correct. It is much simpler than https://github.com/kohya-ss/sd-scripts/pull/1359. I will remove this change if not correct.


heyalexchoi:
> Thank you for this! For `--skip_to_step`, this implementation calls DataLoader for all steps, it will take a lot of time if the dataset is large. For example, if one epoch has 10,000 steps, skipping 10 epochs needs 100,000 calls for DataLoader. So we needs 'skipping epoch' feature.
> 
> I will review other updates in this PR.


Would something like this work?:
```
# before dataloader loop
steps_skipped = args.skip_to_step - 1
 accelerator.skip_first_batches(train_dataloader, steps_skipped)
 progress_bar.update(steps_skipped)
 global_step += steps_skipped
 
 # then
 for step, batch in enumerate(train_dataloader):
 ```
 
 I don't completely follow the changes in the other PR. I can remove the skipping part from the PR if I'm not getting it right.
 
 Also, I added another small change, details in original post.

heyalexchoi:
Question:
Do you know if there is a reason `batch_size=1` in cache latents and text encoder outputs?
https://github.com/kohya-ss/sd-scripts/blob/main/tools/cache_text_encoder_outputs.py#L129
https://github.com/kohya-ss/sd-scripts/blob/main/tools/cache_latents.py#L124

Would it be ok to make this adjustable w/ args?

PR #1381, opened 2024-06-24
add Optimi(more fused_background_optimizer and new function)

sdbds:
https://optimi.benjaminwarner.dev/

New optimizer:
adamw、lion、ranger and stableadamw from Optimi

New function:
Low Precision Training with Kahan Summation(auto use when use above optimizers)
image

Gradient Release(same as fused background pass)
Auto use when use above optimizers.

Fully Decoupled Weight Decay(looks like decoupled lr)

Optimizer Accumulation
[Gradient accumulation](https://pytorch.org/docs/stable/notes/amp_examples.html#gradient-accumulation) reduces training memory by splitting a batch into micro-batches and accumulating micro-batch gradients into the larger batch. [Gradient release](https://optimi.benjaminwarner.dev/gradient_release/) reduces training memory by limiting gradients to one layer at any given time. Optimizer accumulation unifies these two disparate approaches by accumulating gradients directly into optimizer states while performing gradient release.

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -4017,7 +4017,7 @@ def get_optimizer(args, trainable_params):
 
     if args.fused_backward_pass:
         assert (
-            optimizer_type == "Adafactor".lower()
+            optimizer_type in ["Adafactor".lower(),"lion", "adan", "adamw","ranger","stableadamw"]
         ), "fused_backward_pass currently only works with optimizer_type Adafactor / fused_backward_passは現在optimizer_type Adafactorでのみ機能します"
         assert (
             args.gradient_accumulation_steps == 1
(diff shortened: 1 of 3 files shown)


sdbds:
The problem now is that the VRAM usage keeps going up after using prepare_for_gradient_release() for hook, not sure about the problem at the moment...

feffy380:
Some of optimi's features do not support fp16. They should not replace the original optimizers.

sdbds:
> Some of optimi's features do not support fp16. They should not replace the original optimizers.

Sure, i will add fp16 checker, thanks

feffy380:
I'm saying it should not be considered a substitute for the original optimizers. Some features do work with fp16 (like low precision training), but others don't.

You don't need an fp16 check, you need to treat the optimi versions as completely separate optimizers, the same way Adamw8bit is considered separate from Adamw

PR #1383, opened 2024-06-24
Generate samples together with the dataset.

gesen2egee:
If your sample's prompt includes __caption__, it will select an image from the batch (preferring those without caption dropout). 
It then combines the original latent and caption to produce a sample, which is displayed at the top. 
Each occurrence of __caption__ generates once, and can be combined with the original sample args.

image
image

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -5106,6 +5106,7 @@ def sample_images_common(
     tokenizer,
     text_encoder,
     unet,
+    example_tuple=None,
     prompt_replacement=None,
     controlnet=None,
 ):
@@ -5202,7 +5203,18 @@ def sample_images_common(
     if distributed_state.num_processes <= 1:
         # If only one device is available, just use the original prompt list. We don't need to care about the distribution of prompts.
         with torch.no_grad():
+            idx = 0
             for prompt_dict in prompts:
+                if prompt_dict.get("prompt") == '__caption__' and example_tuple:
+                    while example_tuple[1][idx] == '':
+                        idx = (idx + 1) % len(example_tuple[1])
+                        if idx == 0:
+                            break
+                    prompt_dict["prompt"] = example_tuple[1][idx] 
+                    prompt_dict["height"] = example_tuple[0].shape[2] * 8
+                    prompt_dict["width"] = example_tuple[0].shape[3] * 8
+                    prompt_dict["original_lantent"] = example_tuple[0][idx].unsqueeze(0)
+                    idx = (idx + 1) % len(example_tuple[1])
                 sample_image_inference(
                     accelerator, args, pipeline, save_dir, prompt_dict, epoch, steps, prompt_replacement, controlnet=controlnet
                 )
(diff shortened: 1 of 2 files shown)


library/train_util.py
```diff
@@ -5202,7 +5203,18 @@ def sample_images_common(
...
         # If only one device is available, just use the original prompt list. We don't need to care about the distribution of prompts.
         with torch.no_grad():
+            idx = 0
             for prompt_dict in prompts:
+                if '__caption__' in prompt_dict.get("prompt") and example_tuple:
+                    while example_tuple[1][idx] == '':
+                        idx = (idx + 1) % len(example_tuple[1])
+                        if idx == 0:
+                            break
+                    prompt_dict["prompt"] = prompt_dict.get("prompt").replace('__caption__', 'example_tuple[1][idx]')
```
v0xie, line 5213: Should this be:
 `prompt_dict["prompt"] = prompt_dict.get("prompt").replace('__caption__', example_tuple[1][idx])`?

Current behavior replacing with the string 'example_tuple[1][idx]' instead of the value of the tuple

library/train_util.py
```diff
@@ -5202,7 +5203,18 @@ def sample_images_common(
...
+                if prompt_dict.get("prompt") == '__caption__' and example_tuple:
+                    while example_tuple[1][idx] == '':
+                        idx = (idx + 1) % len(example_tuple[1])
+                        if idx == 0:
+                            break
+                    prompt_dict["prompt"] = example_tuple[1][idx] 
+                    prompt_dict["height"] = example_tuple[0].shape[2] * 8
+                    prompt_dict["width"] = example_tuple[0].shape[3] * 8
+                    prompt_dict["original_lantent"] = example_tuple[0][idx].unsqueeze(0)
+                    idx = (idx + 1) % len(example_tuple[1])
```
DKnight54, line 5217: Placing the replacement here would likely cause it to not work properly when running distributed training (ie, training on more than one device) as this section of code will only run when there is only one device as a result of this line:
https://github.com/gesen2egee/sd-scripts/blob/fa5fafcb5eecbe7c573590854d62b9df7baaecff/library/train_util.py#L5203

I would instead suggest that you move it to the line_to_prompt_dict() function, where you can directly overwrite the prompt and the prompt arguements as it's being loaded.
https://github.com/gesen2egee/sd-scripts/blob/fa5fafcb5eecbe7c573590854d62b9df7baaecff/library/train_util.py#L5043-L5095

library/train_util.py
```diff
@@ -5202,7 +5203,18 @@ def sample_images_common(
...
                 sample_image_inference(
```
DKnight54, line 5218: Having tried running the code, I realized that `--sample_at_first` doesn't work well with it, understandably so as the latents and captions have not been loaded at this point.

Perhaps adding a check to skip generation if the caption is still '__caption__' and provide a warning message that generation was skipped due to no captions loaded yet would make the experience more user friendly than the user being confused why the samples generated before training are all random images.

DKnight54: Hrm. Having done more testing, this code breaks if it encounters a training batch size of 1, either due to train_batch_size set to one, or if coincidentally the batch only has one item due to bucketing.

PR #1391, opened 2024-06-27
Translation Update for Multiple Documents

shiertier:


shiertier:
Your code's awesome, and my translation gig is just adding to the coolness. I've tweaked the doc names like you asked, dumping the '-en' from the English ones. Whenever I've got some free time in the future, I'll keep plugging away at translating those docs that are only in Japanese, making sure the English versions are all set.

PR #1394, opened 2024-06-28
fix mmdit.model_type for multi-GPU training with torch

liaoliaojun:
code: 
python -m torch.distributed.run --nproc_per_node=4 "./sd3_train.py" --config_file="xxxx.toml"


liaoliaojun:
Yes, my test is ok, it's great!

PR #1396, opened 2024-06-28
Add VRAM free after sample images

v0xie:
This PR addresses an issue with VRAM usage after generating sample images.

Without using this PR, when training with SDXL and batch size > 1, the first time during training when sample images are generated, I would run out of VRAM even with 24GB (no --xformers or --mem_eff_attn).

Notably, the `sample_images_common` method in train_util calls `clean_memory_on_device` near the end of the function, but I hypothesize that some of the memory is still used after the operation at the end. So calling it again after exiting the function clears the remaining.

I'm not sure if anyone else has had this issue or if it's an issue with my configuration only. Please feel free to close this PR if it's unnecessary. Thank you for your consideration.

### DIFF
--- a/train_network.py
+++ b/train_network.py
@@ -1028,6 +1028,8 @@ class NetworkTrainer:
 
                     self.sample_images(accelerator, args, None, global_step, accelerator.device, vae, tokenizer, text_encoder, unet)
 
+                    clean_memory_on_device(accelerator.device)
+
                     # 指定ステップごとにモデルを保存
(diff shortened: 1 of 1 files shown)


araleza:
This seems good, I think that at times in the past, I've seen training work up until the sample images were generated, and then not be able to restart afterwards.

PurplefinNeptuna:
using this makes my vram usage moving alot during training steps. did it also clean memory every training steps?

PR #1406, opened 2024-07-01
Adjustments to resuming training

Cauldrath:
Currently skips the resumed epoch if partway through and calculates global steps as only the number of steps into the current epoch.

These changes make it resume mid epoch on the appropriate step, with the right global step count, so max steps will be honored.

This also includes a change to just do a multiplication instead of a for loop over every elapsed epoch.

### DIFF
--- a/train_network.py
+++ b/train_network.py
@@ -882,10 +882,9 @@ class NetworkTrainer:
 
         # training loop
         if initial_step > 0:  # only if skip_until_initial_step is specified
-            for skip_epoch in range(epoch_to_start):  # skip epochs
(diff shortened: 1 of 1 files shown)


slashedstar:
This fixed the problems I was having, when resuming from 200, 400 to 1000 steps (when it outputted "epoch is incremented. current_epoch: 0, epoch: 1") it worked as intended, but resuming from the 1200 steps onwards (when it outputted "epoch is incremented. current_epoch: 0, epoch: 2") the training continued after the maximum amount of steps (it also didn't even save a model when reaching the max steps)
image

Cauldrath:
Yes, `--max_train_steps` combined with resuming or setting `--initial_steps` is the main problem if it isn't starting in the first epoch.

PR #1417, opened 2024-07-08
Sd3 freeze x_block

sdbds:
use
`--num_last_block_to_freeze`
to set freeze blocks num,reverse freeze.
(default none,you can set 0~240)

SD3 has 240 x_blocks, i test if freeze half of it will get better effect.

Comparison：
Before↓
image

After freeze 120↓
image

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -3061,6 +3061,12 @@ def add_sd_models_arguments(parser: argparse.ArgumentParser):
         default=None,
         help="directory for caching Tokenizer (for offline training) / Tokenizerをキャッシュするディレクトリ（ネット接続なしでの学習のため）",
     )
+    parser.add_argument(
+        "--num_last_layers_to_freeze",
+        type=int,
+        default=None,
+        help="num_last_layers_to_freeze",
+    )
 
 
 def add_optimizer_arguments(parser: argparse.ArgumentParser):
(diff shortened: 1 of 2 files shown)


sdbds:
> Thank you for this! This seems to work like `block_lr` for SDXL. I will merge when I can confirm the SD3 training is fine :)

Yes, the functionality is very similar with ``block_lr`.
After freezing the blocks, I only input the parameters of the blocks that need to be trained to the optimizer. 
This can reduce additional overhead.

FurkanGozukara:
how do you know which layers you should freeze or it is purely experimental?

sdbds:
> how do you know which layers you should freeze or it is purely experimental?

I received advice from SD3 trainer.

FurkanGozukara:
> > how do you know which layers you should freeze or it is purely experimental?
> 
> I received advice from SD3 trainer.

makes sense. so what parameter they suggest ? like `--num_last_block_to_freeze 80` ?

sdbds:
> > > how do you know which layers you should freeze or it is purely experimental?
> > 
> > 
> > I received advice from SD3 trainer.
> 
> makes sense. so what parameter they suggest ? like `--num_last_block_to_freeze 80` ?

120~240，i think magic numbers between these.

PR #1426, opened 2024-07-12
Judge image size for using diff interpolation

sdbds:
cv2.INTER_AREA only use small size in resizing.


sdbds:
update for replacing CV2 resize to Pil resize

PR #1429, opened 2024-07-17
Update train_util.py to allow for using SGD without momentum

ddpasa:
Using momentum increases VRAM requirement by quite a bit, since you have to keep track of past gradients.  momentum=0 might not be optimal, but is nice to have if you're using low-memory GPUs.

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -3965,7 +3965,9 @@ def get_optimizer(args, trainable_params):
             optimizer_kwargs["momentum"] = 0.9
 
         optimizer_class = torch.optim.SGD
-        optimizer = optimizer_class(trainable_params, lr=lr, nesterov=True, **optimizer_kwargs)
+        uses_momentum = False if optimizer_kwargs["momentum"]==0 else True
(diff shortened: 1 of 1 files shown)


FurkanGozukara:
can you give some information regarding what is SGD  and momentum?

ddpasa:
SGD is stocastic gradient descent optimizer here: 
https://en.wikipedia.org/wiki/Stochastic_gradient_descent
https://pytorch.org/docs/stable/generated/torch.optim.SGD.html

Momentum uses past gradients to improve convergence.  It works nice, but also increases VRAM usage quite a bit.  momentum=0 uses less VRAM, allowing training on older GPUs.

FurkanGozukara:
@ddpasa thanks a lot for explanation

ddpasa:
@kohya-ss what do you think?  Is this okay to merge?

ddpasa:
It's kind of funny to call it SGDNesterov because the underlying optimizer used is just pytorch SGD, which supports SGD with or without Nesterov momentum.  The most ideal solution is to rename it to just SGD, but that would probably break too many things.

PR #1409, opened 2024-07-01
Add CAME Optimizer

xzuyn:
Adds [CAME: Confidence-guided Adaptive Memory Efficient Optimization](https://arxiv.org/abs/2307.02047). Requires you to `pip install pytorch-came`.

Github: https://github.com/yangluo7/CAME

PR #1468, opened 2024-08-17
Adam-mini

sdbds:
Tested batch size 1 can use about 16GB of VRAM.

### DIFF
--- a/library/adam_mini.py
+++ b/library/adam_mini.py
@@ -1,0 +1,259 @@
+import math
+from typing import Iterable, Tuple, Union, Optional
+
+import torch
+import torch.nn as nn
+import torch.distributed as dist
+from torch.distributed._tensor import Replicate
+
+
+class Adam_mini(torch.optim.Optimizer):
+    def __init__(
+            self,
+            named_parameters: Iterable[Tuple[str, nn.Parameter]],
+            lr: Union[float, torch.Tensor] = 1e-3,
+            betas: Tuple[float, float] = (0.9, 0.999),
+            eps: float = 1e-8,
+            weight_decay: float = 0,
+            *,
+            model_sharding: bool = True,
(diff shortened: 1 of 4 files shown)


sdbds:
> Thank you, that looks good!
> 
> Is the Adam_mini code from some repository? If so, please let me know so I can add the reference source and the license.

code from here and i think we should test for different blocks.
I changed block checked for lora part.
https://github.com/zyushun/Adam-mini

FurkanGozukara:
> > Thank you, that looks good!
> > Is the Adam_mini code from some repository? If so, please let me know so I can add the reference source and the license.
> 
> code from here and i think we should test for different blocks. Officail says reduce 50% VRAM but i just found reduce 20%... I changed block checked for lora part. https://github.com/zyushun/Adam-mini

this sounds amazing half VRAM of Adam for same performance

ah only 20% vram usage :/ did you got good results with it? par with Adam for real?

sdbds:
> > > Thank you, that looks good!
> > > Is the Adam_mini code from some repository? If so, please let me know so I can add the reference source and the license.
> > 
> > 
> > code from here and i think we should test for different blocks. Officail says reduce 50% VRAM but i just found reduce 20%... I changed block checked for lora part. https://github.com/zyushun/Adam-mini
> 
> this sounds amazing half VRAM of Adam for same performance
> 
> ah only 20% vram usage :/ did you got good results with it? par with Adam for real?

At least on the SDXL I tested it works about the same.

FurkanGozukara:
@sdbds thanks a lot for answer

do you think adafactor or mini adam would save more vram have you compared?

Bocchi-Chan2023:
I'm curious about the difference in performance between CAME and adam-mini
https://github.com/kohya-ss/sd-scripts/pull/1409

PR #1508, opened 2024-08-24
Fix issue: "RuntimeError: only Tensors of floating point dtype can require gradients" and RuntimeError: "index_select_cuda" not implemented for 'Float8_e4m3fn'

fireicewolf:
As Issue #1509


fireicewolf:
Fix bug in issue #1512

fireicewolf:
As [kohya-ss](https://github.com/kohya-ss) will modify fp8 codes with own code structure. This PR closed.

PR #1529, opened 2024-08-29
fix: 

Akegarasu:
fix `text_encoder_conds` referenced before assignment

And due to [train_network.py#L1118](https://github.com/kohya-ss/sd-scripts/blob/sd3/train_network.py#L1118), `text_encoder_conds` can't be `None` so i make it an empty list


Akegarasu:
This part not only affects the training of the SD3 series, but SDXL also seems to be affected when cache_text_encoder_outputs is turned off

PR #1564, opened 2024-09-05
Flux shift method update

sdbds:
Mainly the math part is updated.

```
def get_lin_function(x1: float = 256, y1: float = 0.5, x2: float = 4096, y2: float = 1.15) -> Callable[[float], float]:
    m = (y2 - y1) / (x2 - x1)
    b = y1 - m * x1
    return lambda x: m * x + b
```

The original y1, y2 is based on the minimum resolution x1=256,maximum resolution x2=4096 for muti-reso calculation.

However, we usually use 256~2048 for training, so we directly use the original multiplier for linear scaling.

For example if x2=2048 then the original function result should be 0.8. 

Since it is scaled down twice, the final result is scaled up to 1.6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,9 +77,9 @@ There are many unknown points in FLUX.1 training, so some settings can be specif
 
 The existing `--loss_type` option may be useful for FLUX.1 training. The default is `l2`.
 
-~~In our experiments, `--timestep_sampling sigma --model_prediction_type raw --discrete_flow_shift 1.0` with `--loss_type l2` seems to work better than the default (SD3) settings. The multiplier of LoRA should be adjusted. ~~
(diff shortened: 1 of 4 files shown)


sdbds:
> I'm not sure, but I think that x1 and x2 are the hyperparameters ​​when the model is trained. Therefore, no matter what resolution we train at, I think that x2 and x2 should remain their original values.
> 
> If we train it on a large dataset with x1=256 and x2=2048, the model will be able to generate images with a resolution of 256-2048 very well. But I think that would require some really extensive fine tuning.

Further experiments are being conducted to see if the parameters can be modified.Theoretically these should be fixed parameters after the model has been trained, but some comments on Twitter say that using 1.5 to 1.6 works better, suspecting a linear relationship.The results so far have been both fuzzy and good.

PR #1576, opened 2024-09-07
Improve wandb logging

p1atdev:
This PR does that:

   - Fix sample image logging that causes incorrect step skips logged.
   - Log sample image prompts as their captions to wandb.


p1atdev:
FYI, the sample image log with the caption looks like this:

image

PR #1605, opened 2024-09-16
sd3 schedule free opt

kohya-ss:
Some refactoring #1600

- Add ScheduleFreeWrapper for the optimizers other than torch.optim.
- Add another wrapper called `OptimizerProxy` which is a subclass or Optimizer to make Accelerate can wrap ScheduleFreeWrapper. Because ScheduleFreeWrapper is not a subclass of Optimizer, Accelerate ignores it. 
- Returns dummy lr scheduler if a schedule free optimizer is used. This can remove checking for `lr_scheduler.step()` and `accelerator.prepare()`. Also logging LR may work.
- Remove `optimizer.eval()` and `optimizer.train()`. Because sd-scripts doesn't have evaluation mode (validation), so optimizer always can be train mode.

However, this branch doesn't work with `--optimizer_schedulefree_wrapper` at this time. The error is like this (this sample is with AdamW, another optimizer may raise with `KeyError: 'step'`):
```
    exp_avgs.append(state["exp_avg"])
KeyError: 'exp_avg'
```

Any help would be appreciated.

@sdbds Please let me know if you have any ideas.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,14 @@ The command to install PyTorch is as follows:
 
 ### Recent Updates
 
+Sep 18, 2024:
+
+- Schedule-free optimizer is added. Thanks to sdbds! See PR [#1600](https://github.com/kohya-ss/sd-scripts/pull/1600) for details.
+  - `schedulefree` is added to the dependencies. Please update the library if necessary.
+  - AdamWScheduleFree or SGDScheduleFree can be used. Specify `adamwschedulefree` or `sgdschedulefree` in `--optimizer_type`.
+  - Wrapper classes are not available for now.
+  - These can be used not only for FLUX.1 training but also for other training scripts after merging to the dev/main branch.
+
 Sep 16, 2024:
 
  Added `train_double_block_indices` and `train_double_block_indices` to the LoRA training script to specify the indices of the blocks to train. See [Specify blocks to train in FLUX.1 LoRA training](#specify-blocks-to-train-in-flux1-lora-training) for details.


sdbds:
> Some refactoring #1600
> 
> * Add ScheduleFreeWrapper for the optimizers other than torch.optim.
> * Add another wrapper called `OptimizerProxy` which is a subclass or Optimizer to make Accelerate can wrap ScheduleFreeWrapper. Because ScheduleFreeWrapper is not a subclass of Optimizer, Accelerate ignores it.
> * Returns dummy lr scheduler if a schedule free optimizer is used. This can remove checking for `lr_scheduler.step()` and `accelerator.prepare()`. Also logging LR may work.
> * Remove `optimizer.eval()` and `optimizer.train()`. Because sd-scripts doesn't have evaluation mode (validation), so optimizer always can be train mode.
> 
> However, this branch doesn't work with `--optimizer_schedulefree_wrapper` at this time. The error is like this (this sample is with AdamW, another optimizer may raise with `KeyError: 'step'`):
> 
> ```
>     exp_avgs.append(state["exp_avg"])
> KeyError: 'exp_avg'
> ```
> 
> Any help would be appreciated.
> 
> @sdbds Please let me know if you have any ideas.

The need to additionally store dictionary state in some optimizers, such as exp_avg, requires modifying the original schedulefreewarper class to do so.Obviously this requires assistance from the official Facebook repository, so I thought I could start by implementing adamwschedulefree and SGDschedulefree

madman404:
Terribly sorry to reopen this, but I do need to note that the schedule free optimizers require the eval call to function properly. Even if there is no actual evaluation, as the schedule free GitHub page says: "The optimizer should also be placed in eval mode when storing checkpoints." https://github.com/facebookresearch/schedule_free?tab=readme-ov-file#how-to-use

This is because (as far as I understand), the weighted averaging math that makes the method actually work is done with the eval call, and is not used for training gradient calls only because it causes slow / poor convergence if trained on.

In line with that, I believe the eval call should be done before saving checkpoints or generating sample images, at minimum.

PR #1640, opened 2024-09-25
New optimizer:AdEMAMix8bit and PagedAdEMAMix8bit

sdbds:
image

Tested to work better than AdamW.

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -2994,7 +2994,7 @@ def add_optimizer_arguments(parser: argparse.ArgumentParser):
         "--optimizer_type",
         type=str,
         default="",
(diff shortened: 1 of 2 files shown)


sdbds:
> Thank you! But I think, by updating the dependent library, we can use optimizer_type arg like `--optimizer_type bnb.optim.AdEMAMix8bit` instead of adding new optimizer types. I don't like to increase the number of optimizer types infinitely😅

It's true that you can do this for non-invasive optimizers, but it's a bit of a gateway for non-developers, and the 8bit class of optimizers is sort of more commonly used.
I've actually  tested about 76 optimizers using pytorch_optimizer and have no plans to add them in.

FurkanGozukara:
@sdbds which optimizer you find best at the moment for sd 1.5 sdxl and flux?

rockerBOO:
> The extra code section is now removed and only the version of the dependency libraries and the comments section are modified.Additional release notes are needed regarding the use of `--optimizer_type bnb.optim.AdEMAMix8bit`

Maybe it can be fully qualified path in the help text. Regular users would then easily know what to enter. 

```
AdEMAMix8bit -> bnb.optim.AdEMAMix8bit
```

gesen2egee:
Is it possible to combine with schedule free?

sdbds:
> Is it possible to combine with schedule free?

It has been submitted to schedulfree officials.

sdbds:
> @sdbds which optimizer you find best at the moment for sd 1.5 sdxl and flux?
> 
> Individually for each

I can tell you if you keep my credits in your videos or whatever:
SD1.5 sdxl prodigy
flux adamWschedulefree and this for now

PR #1646, opened 2024-09-27
Fix CNet training for sd1 (and maybe 2)

elthariel:
This PR fixes:
- An incompatible datastructure passed to diffusers as the `unet.config`, which resulted in an error about SimpleNamespace not being iterable. (It doesn't have the `__contains__` method).
- A missing configuration item `num_attention_heads`. The value was taken from the yml model config from the original cnet repo
- A missing configuration item : `"mid_block_type": "UNetMidBlock2DCrossAttn"` which was guessed based on a vague interpretation of the original cnet's code. It seems to work. I haven't added this value for sd2, maybe we should ? In all the cases please correct me if I'm wrong on this side

### DIFF
--- a/train_controlnet.py
+++ b/train_controlnet.py
@@ -164,6 +164,7 @@ def train(args):
             "num_class_embeds": None,
             "resnet_time_scale_shift": "default",
             "projection_class_embeddings_input_dim": None,
+            "num_attention_heads": 8,
         }
     else:
         unet.config = {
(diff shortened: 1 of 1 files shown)


elthariel:
I'm really sorry, I wasn't aware of the `dev` branch existence. It is indeed fixed in that branch, thanks !

That being said, my implementation of the SimpleNamespace fix is simpler and shorter, in case you want to limit the amount of code to maintain.

I'll close it anyway

PR #1655, opened 2024-09-29
ControlNet training for SDXL

kohya-ss:
WIP Sample image generation is not implemented yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,16 @@ The command to install PyTorch is as follows:
 
 ### Recent Updates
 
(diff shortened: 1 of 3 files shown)


sdbds:
Is it necessary to set strategy_controlnet for different architectures later?
For example, xinsir trained controlnet++, 
others like controlnext and controlnetllite.

blacklig:
Hi, @kohya-ss why did  you merged sdxl_train_contol_net.py into SD3 branch? Took mi a while to found it :))

Issue #1375, opened 2024-06-16
[Feature Request] Add new way to manage timestep distribution and new loss offset feature


Anzhc:
We were working some time on an way to create timestep distribution specific to any particular dataset used in training.

We tested it for last couple months using Derrian easy scripts trainer with hardcoded changes, but we're not familiar with sd-scripts structure enough to fully integrate that as toggleable feature on our own at the current point in time, hence feature request.

https://github.com/Anzhc/Timestep-Attention-and-other-shenanigans

Here we give code of functions structured roughly correctly for usage with variables in sd-scripts. Also examples how we implemented it in hardcoded way are included in repo (look into "Kohya file examples" folder).
It builds distribution based on received loss values, similar to this:
![322636402-c5368e3c-c074-45a9-a929-e49808f87618](https://github.com/kohya-ss/sd-scripts/assets/133806049/a75b8150-360b-4ea0-919b-9bd934047c37)
But unique to each dataset(as it's based on particular loss values received in scope of training).

In majority of cases it improved convergence speed and quality of outcome in our tests.

We also provide code for Loss Offset Curve feature that works with said distribution, can be scaled and scheduled. It also supposed to improve convergence speed.

Some experimental loss functions are included too, but those are really just experiments.

cheald:
This looks like it could pair really well with an approach that I'm working on and about to publish (the short version is "synchronoise noise std/mean to the underlying model"), which also dramatically improves convergence speed. I'm going to see if I can adapt this approach into my approach, because I have a theory that my approach will flatten the curves you're seeing there, which could potentially mean that the two approaches could cause faster convergence together.

Thanks!

Anzhc:
> This looks like it could pair really well with an approach that I'm working on and about to publish (the short version is "synchronoise noise std/mean to the underlying model"), which also dramatically improves convergence speed. I'm going to see if I can adapt this approach into my approach, because I have a theory that my approach will flatten the curves you're seeing there, which could potentially mean that the two approaches could cause faster convergence together.
> 
> Thanks!

We actually were following your development on new loss approach and it's weighting for last couple days, would love to see that incorporated with your really fancy tech! :D

cheald:
Can you elaborate a bit on the underlying theory of operation of your approach? A quick read through seemed to suggest that you're just tracking losses per-timestep, then weighting timestep selection towards higher-loss timesteps over the course of the run. Do I have that roughly correct?

Anzhc:
[...]

PR #1671, opened 2024-10-04
Add Timestep Sampling Function from SD3 Branch to SD (dev base)

gesen2egee:
This PR introduces the `timestep_sampling` feature from SD3 Branch into the original SD model. The new timestep sampling options offer a more concentrated probability distribution compared to the default uniform sampling, which helps the model focus on specific aspects of learning. The new options can be used via the `--timestep_sampling` argument.

- **Default (`uniform`)**: Keeps the original uniform timestep sampling, which evenly distributes the learning steps.
- **Sigmoid**: Behaves similarly to `shift` by default --discrete_flow_shift = 1.
- **Shift**: Skews the timestep distribution, helping the model concentrate on style learning or specific objects.
- **Flux Shift**: This option shift with the image size, similar to how FLUX behaves in larger images

## Additional Parameters:
- **`--discrete_flow_shift`**: By default set to `1`, uses random normal distribution to sample timesteps. A rightward shift helps the model focus more on style, while a leftward shift aids in learning specific objects. The default value is `1`, providing a balanced form.
- **`--sigmoid_scale`**: Adjusts the shape of the sigmoid function. 
image

While **flux_shift** maintains the distortion effects of FLUX, its application may vary in SD due to differences in model nature and training at fixed resolutions.

It is recommended to use the `--timestep_sampling sigmoid` option, combined with `--soft_min_snr_gamma = 1` 
 By [rockerBOO](https://github.com/rockerBOO) #1068
 https://github.com/kohya-ss/sd-scripts/pull/1068
 for better results, as these settings seem to significantly improve model performance.
 
 Suggested to merge this PR along with the soft min snr gamma PR.

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -19,6 +19,7 @@ from typing import (
     Sequence,
     Tuple,
     Union,
+    Callable,
 )
 from accelerate import Accelerator, InitProcessGroupKwargs, DistributedDataParallelKwargs, PartialState
 import glob
@@ -5239,7 +5240,6 @@ def get_timesteps_and_huber_c(args, min_timestep, max_timestep, noise_scheduler,
             timesteps = time_shift(mu, 1.0, timesteps)
         else:
             timesteps = (timesteps * shift) / (1 + (shift - 1) * timesteps)
-        t = timesteps.view(-1, 1, 1, 1)
         timesteps = min_timestep + (timesteps * (max_timestep - min_timestep))
     else:
         timesteps = torch.randint(min_timestep, max_timestep, (b_size,), device="cpu")
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -3493,7 +3493,7 @@ def add_training_arguments(parser: argparse.ArgumentParser, support_dreambooth: 
         "--discrete_flow_shift",
         type=float,
         default=1.0,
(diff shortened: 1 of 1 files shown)


recris:
Great job! But I wonder if we could improve the approach to adding additional sampling functions.

The space of sampling strategies is vast and there are other valid functions, for example:
- https://github.com/kohya-ss/sd-scripts/discussions/294#discussioncomment-9954382
- https://github.com/kohya-ss/sd-scripts/issues/1375

If we keep adding more parameters every time a new approach shows up it will make maintenance more difficult. I've been thinking on this problem for some time and I'd like to propose a more flexible mechanism.

The idea is to make timestep sampling pluggable in the same way the choice of LR schedule or optimizer are. We could have a generic interface (class), say `TimestepSampler`, and different functions would be implemented as subclasses. Then we chose the class to use via some configuration parameter (like we do for optimizers using `optimizer_type` and `optimizer_args`). 

This way we can add other functions in the future without touching the existing code, and would also simplify the maintenance of private forks by reducing the need to deal with rebase conflicts (which is the main reason I am interested in this).

I leave it to @kohya-ss to comment on this as long term approach.

FurkanGozukara:
@gesen2egee thank you so much

for flux dev i found that Model Prediction Type = raw and Timestep Sampling = sigmoid

What do you think about this?

PR #1677, opened 2024-10-07
No token padding for train_network

gesen2egee:
When training with Illustrious-xl 0.1, it appears that token padding was disabled during the training process. 
https://[arxiv.org/pdf/2409.19946](https://arxiv.org/pdf/2409.19946)
Illustrious: an Open Advanced Illustration Model A.6

Testing with token padding enabled seems to result in worse outcomes. 
To address this issue, I've added the `--no_token_padding option` to the training network, 
allowing users to disable token padding during training for potentially better results.

image

### DIFF
--- a/train_network.py
+++ b/train_network.py
@@ -205,6 +205,9 @@ class NetworkTrainer:
         ds_for_collator = train_dataset_group if args.max_data_loader_n_workers == 0 else None
         collator = train_util.collator_class(current_epoch, current_step, ds_for_collator)
 
+        if args.no_token_padding:
+            train_dataset_group.disable_token_padding()
+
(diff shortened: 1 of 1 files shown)


feffy380:
As far as I can tell, the Illustrious report is saying the padding tokens unintentionally learned information about the images, which causes adverse effects when they're included during inference especially since most UIs will optimize CFG by padding so cond and uncond can be batched together.

As mentioned above, everything must be padded to the same length for batching to work, so as a fix they suggest masking out the padding tokens during training.
@kohya-ss when masking is implemented for SD3, could this also be extended to SDXL since they use the same CLIP encoders?

Issue #1627, opened 2024-09-21
When I train with multiple GPUs, this error will be reported. This has been bothering me for a long time. How can I solve it


jinwei1660:
env：
pytorch 2.6 
2gpus

W0921 17:28:08.135000 56224 Lib\site-packages\torch\distributed\elastic\multiprocessing\redirects.py:29] NOTE: Redirects are currently not supported in Windows or MacOs.
Traceback (most recent call last):
  File "e:\pinokio\bin\miniconda\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "e:\pinokio\bin\miniconda\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "E:\pinokio\api\fluxgym.git\env\Scripts\accelerate.exe\__main__.py", line 7, in <module>
  File "e:\pinokio\api\fluxgym.git\env\lib\site-packages\accelerate\commands\accelerate_cli.py", line 48, in main
    args.func(args)
  File "e:\pinokio\api\fluxgym.git\env\lib\site-packages\accelerate\commands\launch.py", line 1159, in launch_command
    multi_gpu_launcher(args)
  File "e:\pinokio\api\fluxgym.git\env\lib\site-packages\accelerate\commands\launch.py", line 793, in multi_gpu_launcher
    distrib_run.run(args)
  File "e:\pinokio\api\fluxgym.git\env\lib\site-packages\torch\distributed\run.py", line 910, in run
    elastic_launch(
  File "e:\pinokio\api\fluxgym.git\env\lib\site-packages\torch\distributed\launcher\api.py", line 138, in __call__
    return launch_agent(self._config, self._entrypoint, list(args))
  File "e:\pinokio\api\fluxgym.git\env\lib\site-packages\torch\distributed\launcher\api.py", line 260, in launch_agent
    result = agent.run()
  File "e:\pinokio\api\fluxgym.git\env\lib\site-packages\torch\distributed\elastic\metrics\api.py", line 137, in wrapper
    result = f(*args, **kwargs)
  File "e:\pinokio\api\fluxgym.git\env\lib\site-packages\torch\distributed\elastic\agent\server\api.py", line 696, in run
    result = self._invoke_run(role)
  File "e:\pinokio\api\fluxgym.git\env\lib\site-packages\torch\distributed\elastic\agent\server\api.py", line 849, in _invoke_run
    self._initialize_workers(self._worker_group)
[...]

PR #1686, opened 2024-10-10
fix: fix some distributed training error in windows

Akegarasu:
- use gloo backend for windows
- disable libuv in windows ( torch >= 2.4 )

fix: 
- Distributed package doesn't have NCCL built in
- RuntimeError: use_libuv was requested but PyTorch was build without libuv support (#1627)

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -5045,17 +5045,18 @@ def prepare_accelerator(args: argparse.Namespace):
     if args.torch_compile:
         dynamo_backend = args.dynamo_backend
 
-    kwargs_handlers = (
-        InitProcessGroupKwargs(timeout=datetime.timedelta(minutes=args.ddp_timeout)) if args.ddp_timeout else None,
-        (
-            DistributedDataParallelKwargs(
-                gradient_as_bucket_view=args.ddp_gradient_as_bucket_view, static_graph=args.ddp_static_graph
-            )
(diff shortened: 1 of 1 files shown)


FurkanGozukara:
awesome so after this it will work multi gpu on windows?

Akegarasu:
I think its ok. However, theoretically, it's only necessary to set use_libuv=False. There were no libuv-related errors in Torch 2.3 and earlier versions, so further testing can be done.

6DammK9:
Currently trying hard to use Win10 + X299 + 4x3090 for this case.
Best approach is downgrade to `torch==2.3.1`. I have found that `accelerate launch` throw this error without ever visiting codes in this repo. In worse, the stack trace shows that `pytorch` is calling `TCPStore` directly: [pytorch/issues/139990](https://github.com/pytorch/pytorch/issues/139990)
**It requries code change already.** After that [there is a Windows exclusive IP parsing error.](https://github.com/pytorch/pytorch/issues/118378).
Along with `gloo` in thie PR, and pairing `bitsandbytes==0.45.0` to `torch==2.5.1+cu124`, I can get it running, until hitting strange error like `RuntimeError: Trying to create tensor with negative dimension -1727503612: [-1727503612]`. 
~~It's 2am now and I will update this later~~

PR #1690, opened 2024-10-12
Caching latents and Text Encoder outputs with multiple GPUs

kohya-ss:


FurkanGozukara:
awesome. this will work automatically when multi gpu used?

FurkanGozukara:
@kohya-ss those --highvram and --lowvram made 0 impact on my testings previously what they do actually? I tested both for FLUX Fine tuning and FLUX LoRA training

I can test FLUX LoRA multi gpu caching - fine tuning still requires 80 gb GPUs, fused backward pass not working

FurkanGozukara:
@kohya-ss you are awesome thank you so much

SoonNOON:
I hope the full version of GPT 01 will help find the right solution very soon.

sdbds:
I hope it is possible to push the cached data directly to HF, so that pulling latents directly in the cloud platform will not take up cache time, and more hard disk space can be saved on large datasets.

PR #1710, opened 2024-10-18
Differential Output Preservation loss for LoRA

kohya-ss:
A loss that brings the output when LoRA is applied without a trigger word closer to the output when LoRA is not applied.

### DIFF
--- a/sdxl_train_control_net_lllite.py
+++ b/sdxl_train_control_net_lllite.py
@@ -181,7 +181,7 @@ def train(args):
         vae.requires_grad_(False)
         vae.eval()
 
-        train_dataset_group.new_cache_latents(vae, accelerator.is_main_process)
+        train_dataset_group.new_cache_latents(vae, accelerator)
 
         vae.to("cpu")
         clean_memory_on_device(accelerator.device)


FurkanGozukara:
how do we use this can you provide some more info? @kohya-ss

FurkanGozukara:
@bmaltais can we add this to gui so many people want to test


Oct 19, 2024:

- Added an implementation of Differential Output Preservation (temporary name) for SDXL/FLUX.1 LoRA training.
  - A method to make the output of LoRA closer to the output when LoRA is not applied, with captions that do not contain trigger words.
  - Define a Dataset subset for the regularization image (`is_reg = true`) with `.toml`. Add `custom_attributes.diff_output_preservation = true`.
    - See [dataset configuration](docs/config_README-en.md) for the regularization dataset.
  - Specify "number of training images x number of epochs >= number of regularization images x number of epochs".
  - Specify a large value for `--prior_loss_weight` option (not dataset config). We recommend 10-1000.
  - Set the loss in the training without using the regularization image to be close to the loss in the training using DOP.
```
[[datasets.subsets]]
image_dir = "path/to/image/dir"
num_repeats = 1
is_reg = true
custom_attributes.diff_output_preservation = true # Add this

**Comment by recris** [comment]
> A loss that brings the output when LoRA is applied without a trigger word closer to the output when LoRA is not applied.

Do you have a link to research paper (or anything) explaining the concept behind this feature? I'd like to understand this a bit more.

**Comment by dxqbYD** [comment]
> > A loss that brings the output when LoRA is applied without a trigger word closer to the output when LoRA is not applied.
> 
> Do you have a link to research paper (or anything) explaining the concept behind this feature? I'd like to understand this a bit more.

I have proposed this method and am not aware of any paper. People have noted that the idea is close to the regularization proposed by the original Dreambooth paper, but it's not the same.

Here you can find a proof-of-concept implementation that is easier to understand than the full code:
https://pastebin.com/3eRwcAJD

And the SimpleTuner implementation with samples here: https://www.reddit.com/r/StableDiffusion/comments/1g2i13s/simpletuner_v112_now_with_masked_loss_training/

**Comment by dxqbYD** [comment]
> > all my proof-of-concept samples were at weight 1.
> 
> Thank you for your valuable insight. This copy is old and the latest is `10-100`.
> 
> Have you tested it with Flux? I tested it at 100 on SDXL and it seemed to work better than 1, but I tested only once. It seems that FLUX needs a smaller value, 10 seems to work.

only tested (successfully) with Flux, so everything I said above applies to Flux.

tried with SDXL a few months ago before FLUX, but did not get any useful results - maybe this is the reason.

**Comment by recris** [comment]
Interesting approach. I will give it a go, using training data with adjusted captions. Though, having to double the forward passes makes this a questionable trade-off for _my_ use case.

**Comment by FurkanGozukara** [comment]
> > all my proof-of-concept samples were at weight 1.
> 
> Thank you for your valuable insight. This copy is old and the latest is `10-100`.
> 
> Have you tested it with Flux? I tested it at 100 on SDXL and it seemed to work better than 1, but I tested only once. It seems that FLUX needs a smaller value, 10 seems to work.

hopefully i will make a full research once @bmaltais  adds

**Comment by recris** [comment]
@kohya-ss  I am running into trouble when I set `is_reg` and `conditioning_data_dir` (loss masks) in the same `[dataset]`:
```
voluptuous.error.MultipleInvalid: extra keys not allowed @ data['datasets'][0]['subsets'][1]['is_reg']
```

It doesn't have to be the same subset, for example this gives me the error:
```
[[datasets]]
resolution = 512
batch_size = 4
enable_bucket = true

[[datasets.subsets]]
num_repeats = 1
image_dir = "/home/recris/sd_train/test/export/samples_default"
conditioning_data_dir = "/home/recris/sd_train/test/export/samples_default/mask"

[[datasets.subsets]]
is_reg = true
num_repeats = 1
image_dir = "/home/recris/sd_train/test/export/samples_reg"
custom_attributes.diff_output_preservation = true
```

**Comment by aria1th** [comment]
Hello, the initial idea seems to be very similar to our paper : https://arxiv.org/abs/2404.07554
and I totally agree with the idea.
The key idea is, "Adapter applied + but no token" has to be same or similar as "no adapter, no token".

Some obvious limitation of this method is, it is not applicable when vanilla model is not capable or 'poisoned', 'overfitted', which causes loss diverging phenomenon (or, OFT paper described this as 'hyperspheric energy diverge'), so it might have to be used with caution.

**Comment by dxqbYD** [comment]
> Hello, the initial idea seems to be very similar to our paper : https://arxiv.org/abs/2404.07554 and I totally agree with the idea. The key idea is, "Adapter applied + but no token" has to be same or similar as "no adapter, no token".
> 
> Some obvious limitation of this method is, it is not applicable when vanilla model is not capable or 'poisoned', 'overfitted', which causes loss diverging phenomenon (or, OFT paper described this as 'hyperspheric energy diverge'), so it might have to be used with caution.

This is the same idea. Thanks for pointing this out, I didn't know about your paper.
I think this method has even more potential than just preserving prior knowledge on a separate concept ("no token" in your paper). I've written a few sentences about this here: https://github.com/Nerogar/OneTrainer/pull/505#issuecomment-2474763609 but it's mostly just unpublished experiments at this point.

**Comment by FurkanGozukara** [comment]
@bmaltais currently we have to manually edit the toml file to be able to use this?

did you add it to the gui? i want to try : custom_attributes.diff_output_preservation = true # Add this

**Comment by DarkViewAI** [comment]
> @bmaltais currently we have to manually edit the toml file to be able to use this?
> 
> did you add it to the gui? i want to try : custom_attributes.diff_output_preservation = true # Add this
> 
> @kohya-ss this is as usual as previous dreambooth training with reg images but we only add custom_attributes.diff_output_preservation = true to the toml or where?
> 
> @aria1th any way you could make a pull request to kohya for your paper implementation ?

you need to make a dataset.toml file and add it in there, i tried it, it doesn't work too good with flux, i think because of the distillation . works good on sdxl

**Comment by FurkanGozukara** [comment]
> > @kohya-ss this is as usual as previous dreambooth training with reg images but we only add custom_attributes.diff_output_preservation = true to the toml or where?
> 
> Just add `custom_attributes.diff_output_preservation = true` and it works. From my understanding it should be based on the same theory as in that paper.

started trainings for testing  ty

**Comment by aria1th** [comment]
@FurkanGozukara I think current implementation is more stable than our original implementation - since it takes 2x VRAM at original (which effectively forces batch size 2+, always one with condition, one without condition).
Current implementation in kohya-ss allows you to dynamically use batch size - however it we observed that asynchronous / separated updates does not ensure knowledge preservation behavior.

But yeah, the **token removing function** seems to be missing in the implementation, I guess that should be added to correctly specify what has to be preserved as contrastive data.

We are preparing other paper for this phenomenon.
I'll prepare to upload results * push pr within few weeks with some intensive experiment results

**Comment by aria1th** [comment]
@dxqbYD Yes, our paper aims to implement continual learning (for some more specific personalization) - so concept has to be controllable with `class token`, but in general, flux/large DiT tuning with current implementation means `not losing original knowledge generally`  so its the difference.
and yes, as tuning Illustrious model and with large batch experiments, we observed that batch size 1x32 update behaving worse than 2x32 updates (same LR), or 2x16 update (double LR).(We experimented with 256/512/4096 tho) but, gradient accumulation is okay, that is why batch is more stable & better way to handle this.

Having them in same batch (without class token condition separation), could lead to worse result, 'hyperspheric energy diverge' like thing, it gets really unstable, however. 

With EMA, theoritically it should be okay in either way.

**Comment by dxqbYD** [comment]
Thanks for your reply.
To translate this for users: If you are using any optimizer with beta1 > 0, which is basically everything except Adafactor, it shouldn't matter much how the batches are organized.

I'm having good experimental results with Adafactor on Flux too though, but I'm not doubting that there might be cases it works less well with SGD only when regularization is in a separate batch.

**Comment by FurkanGozukara** [comment]
my results failed completely for some reason - used fast branch but i think it has this merged code already

i used below dataset config and 5200 real man images as regularization images (collected from unsplash) , did 1 repeat 

i used 28 training images, set repeating 200 

so did total 10800 steps, i assume 5200 made for reg images and 5600 made for training images

i didnt use any other captions tokens etc only 

ohwx
and
man

image


1 : reduced likeliness

image


2 : 0 improvement of bleeding

image

3 : all dataset still turned into photos of me

image

**Comment by FurkanGozukara** [comment]
> @FurkanGozukara Have you set `--prior_loss_weight` to a large value? The default of 1 has little effect on DOP. Please try 100 or so.

i tried 1 , it was FLUX

so you still say 100? or 10?

**Comment by dxqbYD** [comment]
you can try higher values, but my guess is that something else must be wrong.
I've always used 1 on Flux, and the effect is so strong it is undeniable. The SimpleTuner author has published one of my samples here. This was with weight 1:
https://www.reddit.com/r/StableDiffusion/comments/1g2i13s/simpletuner_v112_now_with_masked_loss_training

look at the second column and compare with the rest

**Comment by FurkanGozukara** [comment]
I have yesterday started training with  prior_loss_weight  1 10 100

prior_loss_weight  10 is definitely preventing bleeding issue but my character, myself, is not anymore learnt properly - still training though at 100 epoch right now 

and prior_loss_weight 100 is like model didnt learn anything at all

I am testing LoKR as well but it sacrifices quality - didnt test reg images method with LoKR

download (21)

**Comment by heinrichI** [comment]
> The following should work now:
> 
> ```
> [[datasets]]
> resolution = 512
> batch_size = 4
> enable_bucket = true
> 
> [[datasets.subsets]]
> num_repeats = 1
> image_dir = "/home/recris/sd_train/test/export/samples_default"
> conditioning_data_dir = "/home/recris/sd_train/test/export/samples_default/mask"
> 
> [[datasets]]
> resolution = 512
> batch_size = 4
> enable_bucket = true
> 
> [[datasets.subsets]]
> is_reg = true
> num_repeats = 1
> image_dir = "/home/recris/sd_train/test/export/samples_reg"
> custom_attributes.diff_output_preservation = true
> ```

This don't work for me.
datasetMeMaskReg.toml:
```
[general]
shuffle_caption = false
caption_extension = '.txt'
keep_tokens = 1

[[datasets]]
batch_size = 2
enable_bucket = true
resolution = [1024, 1024]

  [[datasets.subsets]]
  image_dir = 'j:\Train'
  num_repeats = 1
  conditioning_data_dir = 'j:\TrainMask'

[[datasets]]
batch_size = 2
enable_bucket = true
resolution = [1024, 1024]

  [[datasets.subsets]]
  is_reg = true
  image_dir = 'j:\RegulizationPeople'
  num_repeats = 2
  custom_attributes.diff_output_preservation = true
```
And the output in the process:
```
num train images * repeats / 学習画像の数×繰り返し回数: 102
num reg images / 正則化画像の数: 178
num batches per epoch / 1epochのバッチ数: 55
num epochs / epoch数: 100
batch size per device / バッチサイズ: 2, 2
gradient accumulation steps / 勾配を合計するステップ数 = 1
total optimization steps / 学習ステップ数: 5500
```
As you can see, the number of batches is equal to half of the images, and the regulation images are completely ignored.

dxqbYD:
@aria1th 
> (We're still experimenting and checking more stuff but kinda sure about this)

More experiments: https://github.com/Nerogar/OneTrainer/pull/505#issuecomment-2563860623

PR #1713, opened 2024-10-20
Support image2image mode for Flux

gritukan:


flux_minimal_inference.py
```diff
@@ -196,17 +194,23 @@ def generate_image(
...
     negative_prompt: Optional[str],
     cfg_scale: float,
+    strength: float,
 ):
     seed = seed if seed is not None else random.randint(0, 2**32 - 1)
     logger.info(f"Seed: {seed}")
 
+    if steps is None:
+        steps = 4 if is_schnell else 50
+    timesteps = get_schedule(steps, image_width * image_height // 64 // 4, shift=not is_schnell)
```
gritukan, line 212: Using the schedule parameters from original Flux inference: https://github.com/black-forest-labs/flux/blob/main/demo_st.py#L193

PR #1715, opened 2024-10-21
Fix training for V-pred and ztSNR

catboxanon:
This PR:
1) Updates debiased estimation loss function for V-pred. The previous function was intended only for epsilon. For reference: https://github.com/kohya-ss/sd-scripts/issues/1058#issuecomment-1916893278
2) Prevents now-deprecated scaling of loss if ztSNR is enabled. For reference: https://github.com/kohya-ss/sd-scripts/pull/934

cc @feffy380 @sdbds, let me know if I'm missing something.

### DIFF
--- a/fine_tune.py
+++ b/fine_tune.py
@@ -383,10 +383,10 @@ def train(args):
 
                     if args.min_snr_gamma:
                         loss = apply_snr_weight(loss, timesteps, noise_scheduler, args.min_snr_gamma, args.v_parameterization)
-                    if args.scale_v_pred_loss_like_noise_pred:
+                    if args.scale_v_pred_loss_like_noise_pred and not args.zero_terminal_snr:
                         loss = scale_v_prediction_loss_like_noise_prediction(loss, timesteps, noise_scheduler)
                     if args.debiased_estimation_loss:
-                        loss = apply_debiased_estimation(loss, timesteps, noise_scheduler)
+                        loss = apply_debiased_estimation(loss, timesteps, noise_scheduler, args.v_parameterization)
 
                     loss = loss.mean()  # mean over batch dimension
                 else:
--- a/library/custom_train_functions.py
+++ b/library/custom_train_functions.py
@@ -96,10 +96,13 @@ def add_v_prediction_like_loss(loss, timesteps, noise_scheduler, v_pred_like_los
     return loss
 
 
-def apply_debiased_estimation(loss, timesteps, noise_scheduler):
+def apply_debiased_estimation(loss, timesteps, noise_scheduler, v_prediction=False):
     snr_t = torch.stack([noise_scheduler.all_snr[t] for t in timesteps])  # batch_size
     snr_t = torch.minimum(snr_t, torch.ones_like(snr_t) * 1000)  # if timestep is 0, snr_t is inf, so limit it to 1000
-    weight = 1 / torch.sqrt(snr_t)
+    if v_prediction:
+        weight = 1 / (snr_t + 1)
+    else:
+        weight = 1 / torch.sqrt(snr_t)
     loss = weight * loss
     return loss
 
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -3731,6 +3731,11 @@ def verify_training_args(args: argparse.Namespace):
         raise ValueError(
             "scale_v_pred_loss_like_noise_pred can be enabled only with v_parameterization / scale_v_pred_loss_like_noise_predはv_parameterizationが有効なときのみ有効にできます"
         )
+    
+    if args.scale_v_pred_loss_like_noise_pred and args.zero_terminal_snr:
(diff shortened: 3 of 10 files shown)


liesened:
At the current time I believe you never actually want to use scale_v_prediction_loss_like_noise_prediction, as it was originally a "fix" for an already outdated and wrongly implemented version of MinSNR. Tying it to ZSNR doesn't really make any sense.

catboxanon:
You're right. I guess a warning only will have to suffice.

catboxanon:
> is this correct?

Difficult for me to say affirmatively. I'd prefer waiting for a second or third opinion on it.

> Based on this chart, I think there may be some use for `scale_v_pred_loss_like_noise_pred` as well.

I can remove the deprecation notice if you would like.

liesened:
Your plots are a bit confusing, I assume "Debiased estimation" is the one in this PR with v_prediction=False, "Debiased estimation vpred" is with v_prediction=True, so far good, but with "SNR" it gets unclear. If you use "Scale vpred like noise vpred" with "SNR weighted loss vpred", then "Scale vpred like noise vpred" definitely doesn't look right to me. The weighting shouldn't decrease with timesteps increasing.

The purpose of this change is to clamp debiased weighting between (0,1) to accomodate for v-prediction loss.

liesened:
Some A/B/C tests from bluvoll trying to convert eps-based checkpoint to vpred:
MinSNR | broken Debiased | fixed Debiased
xyz_grid-0007-4018816163
xyz_grid-0010-2740262790
xyz_grid-0011-4062610619
xyz_grid-0012-1817032482

liesened:
Tried to do some math and plot this as well. 
image



I think it's important to see how each weighting strategy changes the effective SNR. However, in practice no one trains v-pred without ZSNR, it would make more sense to plot using ZSNR schedule. Notice that `Debiased estimation vpred weighting * SNR(t)` overlaps with `vpred-like loss`:
image

- Debiased estimation vpred weighting: `apply_debiased_estimation` , v_pred=True
- vpred-like loss: `scale_v_prediction_loss_like_noise_prediction`
- MinSNR(5) vpred: `apply_snr_weight`, min_snr_gamma=5, vpred=True
- True Debiased estimation weighting: discussed in #1058, basically it's `1/SNR(t)`

If there's anything of note here, it's that **Debiased+vpred** AND **vpred-like loss** look suspiciously alike to **MinSNR+vpred** at higher timesteps, which can suggest that `scale_v_prediction_loss_like_noise_prediction` may (or may not) be useful for **Debiased+vpred**. Didn't include some options since the readability is already really poor. Either way, discarding `scale_v_prediction_loss_like_noise_prediction` right away was a poor choice.

liesened:
Additionally, here's a plot comparing variants of debiased estimation.
image

Non-vpred variant is kinda useless with that spike at the beginning.

What I'd suggest is to leave `scale_v_pred_loss_like_noise_pred` as is. V-prediction variant of Debiased estimation seems useful.

sdbds:
> Some A/B/C tests from bluvoll trying to convert eps-based checkpoint to vpred on a tiny dataset: MinSNR | broken Debiased | fixed Debiased xyz_grid-0007-4018816163 xyz_grid-0010-2740262790 xyz_grid-0011-4062610619 xyz_grid-0012-1817032482

It looks good from the results.
For some reason, it seems that debias estimation loss function late in training cause color contamination.

sdbds:
Considering that the original paper was published a long time ago, I thought it would be a good idea to refer to the cited paper for updates.
https://scholar.google.com/scholar?cites=6450976606823846518&as_sdt=2005&sciodt=0,5&hl=zh-CN&oi=gsb

liesened:
@sdbds 
> For some reason, it seems that debias estimation loss function late in training cause color contamination.

I'm not exactly sure what do you mean by "color contamination", but if it's about weird color splotches, I do think this is rather strange. My theory is that they appear because of the new prediction target and schedule being wonky at first, and this will eventually go away with sufficient training. In fact, [this model](https://civitai.com/models/833294?modelVersionId=962003) was trained on 300k samples using `1/(snr+1)` and I don't see any splotches there. It doesn't happen with MinSNR on the test examples likely because MinSNR doesn't rescale neither the mid nor high timesteps, and mid timesteps recieve less training with debiased. Notice how "grey" MinSNR results look. This is only a speculation though, and it may not be true.

>I thought it would be a good idea to refer to the cited paper for updates.

I only found these loosely related papers, but looks like they don't focus on what this PR attempts to do at all. 
- https://www.ecva.net/papers/eccv_2024/papers_ECCV/papers/00328.pdf
- https://arxiv.org/pdf/2312.00210

O-J1:
Did we ever get to the bottom of why min-SNR causes the colors to become washed out? It completely defeats the purpose of ZTSNR (which is deeper blacks and bright whites) on SDXL

PR #1741, opened 2024-10-30
Compute aspect ratios in log-space

catboxanon:
Updates aspect ratio bucketing to align with NovelAI's implementation. See their GitHub repo and paper (specifically footnote [2]) for reference.

https://github.com/NovelAI/novelai-aspect-ratio-bucketing/commits/main/?since=2024-09-25&until=2024-09-25
https://arxiv.org/abs/2409.15997

> [2] aspect ratios compared in log-space have the desirable property that a 1:1 aspect ratio can be equidistant
from buckets 2:1 and 1:2, each of which fit an equal proportion of its area.

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -245,7 +245,7 @@ class BucketManager:
             if reso in self.predefined_resos_set:
                 pass
             else:
-                ar_errors = self.predefined_aspect_ratios - aspect_ratio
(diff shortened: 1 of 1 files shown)


StableLlama:
The full chapter "4.1 Aspect-Ratio Bucketing" is an interesting read.

How much of it (especially the criteria for selection during training sounds interesting as well) is already implemented in the kohya-ss/sd-scripts?

PR #1768, opened 2024-11-07
Dango/timesteps fix

Dango233:
* Remove diffusers dependency in ts & sigma calc
* Support Shift Setting 
* Support timesteps range setting
* Add uniform distribution
* Default to Uniform distribution and shift 1

### DIFF
--- a/library/sd3_train_utils.py
+++ b/library/sd3_train_utils.py
@@ -253,12 +253,12 @@ def add_sd3_training_arguments(parser: argparse.ArgumentParser):
         " / 複数解像度学習時に解像度ごとに位置埋め込みをスケーリングする。SD3.5M以外では予期しない動作になります",
     )
 
-    # copy from Diffusers
+    # Dependencies of Diffusers noise sampler has been removed for clearity.
     parser.add_argument(
         "--weighting_scheme",
         type=str,
-        default="logit_normal",
-        choices=["sigma_sqrt", "logit_normal", "mode", "cosmap"],
+        default="uniform",
+        choices=["sigma_sqrt", "logit_normal", "mode", "cosmap", "uniform"],
         help="weighting scheme for timestep distribution and loss / タイムステップ分布と損失のための重み付けスキーム",
     )
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


Dango233:
It's dataset and training purpose dependent. Sigmoid/logit normal works really well on initial construction of denoising capabilities but not necessarily best for small scale finetuning - uniform distribution is more universal for downstream usecases.
If a training focus heavily on learning overall structures and can ignore details (details/objects/patterns already in model base weight), logit normal (with a shift > 1) still performs great;
but for dataset that needs a lot of attention into details (like anime characters) - uniform distribution should work better.
s
Some extrem cases even need to have shift < 1 to emphasis details (like if you are training a detailed pattern).

So it really depends on what you want to achieve.

________________________________
发件人: Bagheera ***@***.***>
发送时间: 2024年11月7日 20:54
收件人: kohya-ss/sd-scripts ***@***.***>
抄送: Dango233 ***@***.***>; Mention ***@***.***>
主题: Re: [kohya-ss/sd-scripts] Dango/timesteps fix (PR #1768)


@Dango233<https://github.com/Dango233> you guys are seeing better results than the normal flux schedule of sigmoid sampling?

―

bghira:
does this scale with batch size such that around 2048 we really want to weight sampling or is uniform alright then as well?

the explanation about early structure during pretraining does make sense.

Dango233:
I would have to say that's task dependent - but in general if your dataset is very large and they share similar details... logit_normal still make sense sometimes.

________________________________
发件人: Bagheera ***@***.***>
发送时间: 2024年11月7日 21:14
收件人: kohya-ss/sd-scripts ***@***.***>
抄送: Dango233 ***@***.***>; Mention ***@***.***>
主题: Re: [kohya-ss/sd-scripts] Dango/timesteps fix (PR #1768)


does this scale with batch size such that around 2048 we really want to weight sampling or is uniform alright then as well?

the explanation about early structure during pretraining does make sense.

―

dsienra:
I'm having problems to set the LR for the TE and the unet independently, it trains the TE with the same LR I set to the unet.

That's is in my config file

  "learning_rate": 1e-06,
  "learning_rate_te": 4e-06,
  "learning_rate_te1": 4e-06,
  "learning_rate_te2": 4e-06,

Additional parameters: --fused_backward_pass --use_t5xxl_cache_only --train_text_encoder

I think the problem started after the last update but I'm not sure,

FurkanGozukara:
> It's dataset and training purpose dependent. Sigmoid/logit normal works really well on initial construction of denoising capabilities but not necessarily best for small scale finetuning - uniform distribution is more universal for downstream usecases. If a training focus heavily on learning overall structures and can ignore details (details/objects/patterns already in model base weight), logit normal (with a shift > 1) still performs great; but for dataset that needs a lot of attention into details (like anime characters) - uniform distribution should work better. s Some extrem cases even need to have shift < 1 to emphasis details (like if you are training a detailed pattern). So it really depends on what you want to achieve.

Thank you so much

would you recommend any different values for these configs? SD 3.5 Large

image

PR #1779, opened 2024-11-12
Improve block swap speed and apply to LoRA

kohya-ss:
Copy from the README:

- Improved the implementation of block swap and made it available for both FLUX.1 and SD3 LoRA training. See FLUX.1 LoRA training etc. for how to use the new options. Training is possible with about 8-10GB of VRAM.
- During fine-tuning, the memory usage when specifying the same number of blocks has increased slightly, but the training speed when specifying block swap has been significantly improved.
- There may be bugs due to the significant changes. Feedback is welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ The command to install PyTorch is as follows:
 
 ### Recent Updates
 
-Nov 12, 2024:
+Nov 14, 2024:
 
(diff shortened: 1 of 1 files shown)


FurkanGozukara:
@bmaltais i hope you can update asap thank you so much this is a great news for community

bmaltais:
This new feature appear to be mostly on the internal of sd-scripts... So pulling the most recent sd3 branch should bring this in. I just need to make sure the number of block to swap can be set to 35 in the GUI... And that should be it.

FurkanGozukara:
@bmaltais i am testing new commit but there is no block swap feature for LoRA yet?

PR #1808, opened 2024-11-27
Implement pseudo Huber loss for Flux and SD3

recris:
This change set implements pseudo Huber loss functionality that was missing from Flux.1 and SD3 model training.

This also introduces a new parameter `huber_scale` for controlling base threshold of the Huber function. The previous logic had a fixed threshold that worked very poorly with Flux.1 models; This is something that needs to be tuned to the loss "profile" of each model (mean abs latent error, variance, etc.). The existing `huber_c` could only be used for adjusting the decay of the exponential schedule, so a new parameter was required.

Refactoring notes:
- Moved most of the loss estimation logic inside the `conditional_loss` function, `huber_c` is no longer passed through the rest of the code; most of the file changes were due to this.
- Added an explicit error message when "snr" schedule is used with a model currently incompatible with it.

Tested with:
 - SDXL Lora
 - Flux.1 Dev Lora
 - SD3.5 Medium Lora (only checked that it doesn't crash)

For use with Flux.1 I recommend using `huber_scale = 1.8` (or a bit higher) and `huber_c = 0.25`. YMMV.

### DIFF
--- a/fine_tune.py
+++ b/fine_tune.py
@@ -380,7 +380,7 @@ def train(args):
 
                 # Sample noise, sample a random timestep for each image, and add noise to the latents,
                 # with noise offset and/or multires noise if specified
-                noise, noisy_latents, timesteps, huber_c = train_util.get_noise_noisy_latents_and_timesteps(
+                noise, noisy_latents, timesteps = train_util.get_noise_noisy_latents_and_timesteps(
                     args, noise_scheduler, latents
                 )
 
@@ -397,7 +397,7 @@ def train(args):
                 if args.min_snr_gamma or args.scale_v_pred_loss_like_noise_pred or args.debiased_estimation_loss:
                     # do not mean over batch dimension for snr weight or scale v-pred loss
                     loss = train_util.conditional_loss(
-                        noise_pred.float(), target.float(), reduction="none", loss_type=args.loss_type, huber_c=huber_c
+                        args, noise_pred.float(), target.float(), timesteps, "none", noise_scheduler
                     )
                     loss = loss.mean([1, 2, 3])
 
@@ -411,7 +411,7 @@ def train(args):
                     loss = loss.mean()  # mean over batch dimension
                 else:
                     loss = train_util.conditional_loss(
-                        noise_pred.float(), target.float(), reduction="mean", loss_type=args.loss_type, huber_c=huber_c
+                        args, noise_pred.float(), target.float(), timesteps, "none", noise_scheduler
                     )
 
                 accelerator.backward(loss)
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -3905,7 +3905,14 @@ def add_training_arguments(parser: argparse.ArgumentParser, support_dreambooth: 
         "--huber_c",
         type=float,
         default=0.1,
-        help="The huber loss parameter. Only used if one of the huber loss modes (huber or smooth l1) is selected with loss_type. default is 0.1 / Huber損失のパラメータ。loss_typeがhuberまたはsmooth l1の場合に有効。デフォルトは0.1",
+        help="The Huber loss decay parameter. Only used if one of the huber loss modes (huber or smooth l1) is selected with loss_type. default is 0.1 / Huber損失のパラメータ。loss_typeがhuberまたはsmooth l1の場合に有効。デフォルトは0.1",
+    )
+
+    parser.add_argument(
+        "--huber_scale",
+        type=float,
+        default=1.0,
+        help="The Huber loss scale parameter. Only used if one of the huber loss modes (huber or smooth l1) is selected with loss_type. default is 1.0 / Huber損失のパラメータ。loss_typeがhuberまたはsmooth l1の場合に有効。デフォルトは0.1",
     )
 
     parser.add_argument(
@@ -5821,29 +5828,10 @@ def save_sd_model_on_train_end_common(
             huggingface_util.upload(args, out_dir, "/" + model_name, force_sync_upload=True)
 
 
-def get_timesteps_and_huber_c(args, min_timestep, max_timestep, noise_scheduler, b_size, device):
-    timesteps = torch.randint(min_timestep, max_timestep, (b_size,), device="cpu")
-
-    if args.loss_type == "huber" or args.loss_type == "smooth_l1":
-        if args.huber_schedule == "exponential":
-            alpha = -math.log(args.huber_c) / noise_scheduler.config.num_train_timesteps
-            huber_c = torch.exp(-alpha * timesteps)
-        elif args.huber_schedule == "snr":
-            alphas_cumprod = torch.index_select(noise_scheduler.alphas_cumprod, 0, timesteps)
-            sigmas = ((1.0 - alphas_cumprod) / alphas_cumprod) ** 0.5
-            huber_c = (1 - args.huber_c) / (1 + sigmas) ** 2 + args.huber_c
-        elif args.huber_schedule == "constant":
-            huber_c = torch.full((b_size,), args.huber_c)
-        else:
-            raise NotImplementedError(f"Unknown Huber loss schedule {args.huber_schedule}!")
-        huber_c = huber_c.to(device)
-    elif args.loss_type == "l2":
-        huber_c = None  # may be anything, as it's not used
-    else:
-        raise NotImplementedError(f"Unknown loss type {args.loss_type}")
-
-    timesteps = timesteps.long().to(device)
-    return timesteps, huber_c
+def get_timesteps(min_timestep, max_timestep, b_size, device):
+    timesteps = torch.randint(min_timestep, max_timestep, (b_size,), device=device)
+    timesteps = timesteps.long()
+    return timesteps
 
 
 def get_noise_noisy_latents_and_timesteps(args, noise_scheduler, latents):
@@ -5865,7 +5853,7 @@ def get_noise_noisy_latents_and_timesteps(args, noise_scheduler, latents):
     min_timestep = 0 if args.min_timestep is None else args.min_timestep
     max_timestep = noise_scheduler.config.num_train_timesteps if args.max_timestep is None else args.max_timestep
 
-    timesteps, huber_c = get_timesteps_and_huber_c(args, min_timestep, max_timestep, noise_scheduler, b_size, latents.device)
+    timesteps = get_timesteps(min_timestep, max_timestep, b_size, latents.device)
 
     # Add noise to the latents according to the noise magnitude at each timestep
     # (this is the forward diffusion process)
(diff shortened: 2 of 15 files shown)


araleza:
This is great, I was literally just considering asking for Huber loss to be implemented for kohya Flux somehow. It was one of the biggest steps forward for my SDXL training. I'll try to give it a go soon.

fine_tune.py
```diff
@@ -411,7 +411,7 @@ def fn_recursive_set_mem_eff(module: torch.nn.Module):
                     loss = loss.mean()  # mean over batch dimension
                 else:
                     loss = train_util.conditional_loss(
-                        noise_pred.float(), target.float(), reduction="mean", loss_type=args.loss_type, huber_c=huber_c
+                        args, noise_pred.float(), target.float(), timesteps, "none", noise_scheduler
```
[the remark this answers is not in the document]
recris: oops! nice catch.

library/train_util.py
```diff
@@ -5821,29 +5828,10 @@ def save_sd_model_on_train_end_common(
...
-    elif args.loss_type == "l2":
-        huber_c = None  # may be anything, as it's not used
-    else:
-        raise NotImplementedError(f"Unknown loss type {args.loss_type}")
-
-    timesteps = timesteps.long().to(device)
-    return timesteps, huber_c
+def get_timesteps(min_timestep, max_timestep, b_size, device):
+    timesteps = torch.randint(min_timestep, max_timestep, (b_size,), device=device)
+    timesteps = timesteps.long()
```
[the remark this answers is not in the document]
recris: I also do not like the current approach, there should be a "step context" object that collects information about the training context, then it would get passed to all the various methods/strategy/utility functions instead of having to pass around a handful of separate parameters everywhere. But that is a bigger refactoring that I'd like to make here.

recris:
> Thank you, this is great. I think it makes sense to split `get_timesteps_and_huber_c`.
> 
> However, regarding the responsibilities of the methods, I personally don't want to pass `args` to `conditional_loss`. We may refactor it after merging to call `get_timesteps`, `get_huber_threshold`, and `conditional_loss` in order from each script. We appreciate your understanding.

I also do not like the current approach, there should be a "step context" object that collects information about the training context, then it would get passed to all the various methods/strategy/utility functions instead of having to pass around a handful of separate parameters everywhere. But that is a bigger refactoring that I'd like to make here.

sd3_train.py
```diff
@@ -845,7 +845,7 @@ def grad_hook(parameter: torch.Tensor):
                 # )
                 # calculate loss
                 loss = train_util.conditional_loss(
-                    model_pred.float(), target.float(), reduction="none", loss_type=args.loss_type, huber_c=None
+                    args, model_pred.float(), target.float(), timesteps, "none", noise_scheduler
```
OmerMachluf, line 848: where is this noise_scheduler initialized?

library/train_util.py
```diff
@@ -5821,29 +5828,10 @@ def save_sd_model_on_train_end_common(
             huggingface_util.upload(args, out_dir, "/" + model_name, force_sync_upload=True)
 
 
-def get_timesteps_and_huber_c(args, min_timestep, max_timestep, noise_scheduler, b_size, device):
+def get_timesteps(min_timestep, max_timestep, b_size, device):
     timesteps = torch.randint(min_timestep, max_timestep, (b_size,), device="cpu")
```
OmerMachluf, line 5832: would this work under the current implementation of sd3 in this repo? haven't looked if it was updated for a while.
but as far as i can recall, the scheduling was implemented in this repo, and in different way and didnt have the 'add_noise' like other implementations from diffusers.
so essentially, it was randomizing the index for timestep, but then taking the timestep itself from the noise_scheduler and calaculating the noise using sigmas.
maybe this changed, haven't looked in the repo. but worth a double check

recris: If that is an issue, then it is separate from the scope of this PR.

sd3_train.py
recris: Fixed, thanks.

araleza:
I gave it a try... I'm not seeing good results so far. :(  I went for:
```
--loss_type huber --huber_c 0.25 --huber_scale 1.8 --huber_schedule="exponential" 
```
Learning with Huber loss enabled seemed a lot slower for the same LR, and I didn't see any big image quality improvements either.  Maybe it's dataset-specific, and it'll work better for someone else? My training dataset is around 200 images in size, mostly high quality, but with some poorer quality ones mixed in, so I hoped Huber would help with that.

recris:
Also, from past experience with this type of loss function in other models I don't expect this to improve image quality much, this usually helps more with getting _consistency_ in the outputs.

araleza:
I have some better news to report now. Without changing anything from the settings I gave above, I've let the training continue longer. And the image quality has been rising steadily for a while now, and I'm getting some nice output. Maybe Huber just has different learning characteristics.

araleza:
I'm seeing a little bit of grid artifact creeping back in though. I don't know why it is, that when my image quality gets to be very good, that DiT artifact appears. Maybe we need block-specific learning rates? I don't fully understand the problem though.

This is the grid artifact I'm referring to (although this one seems mostly vertical):
image

It seems to be particularly bad at the right hand edge of the images for some reason, although I can see it faintly across the whole image sometimes.

recris:
I usually observe this kind of thing when pushing LoRA strength way above 1.0 during inference, or when applying multiple LoRAs simultaneously without lowering the strength of each. I suspect this is related to having very large weight magnitude due to stacking, or over-training.

Training at 5e-4 seems super high to me, I usually don't go above 2e-4.

What are your `average_key_norm` and `max_key_norm` values (in tensorboard)?  These give an indication on how much the trained model has "deviated" from the base model.

araleza:
I usually have a lower LR, but once the gradient accumulation value goes up, I find I usually have to push the LR higher, as I think the gradient update is divided by the number of accumulations, unlike batch size.

And higher LRs actually work amazingly well with Flux. The images I've been getting have rich colors and interesting camera angles (i.e. the Flux DPO seems to still be active), but also include the trained objects. The pictures I get out from flux_minimal_inference.py (which is what I use for sample images) are honestly the kind of thing I'd put up on my art page, if it wasn't for the grid artifact appearing on them.

I really should get tensorboard up and running with my runs. I do wonder if maybe what we need for Flux is the ability to set the LR individually for each of the 19 double and the 38 single blocks. Maybe the average_key_norm and max_key_norm sizes would guide me towards which ones.

recris:
I've been told that high gradient accumulation (with 16-bit floats) can cause numerical instability and/or loss of precision, not sure if that applies here.

From tests I've made usually these types of visual artifacts tend to appear when `average_key_norm` starts to go above ~1.2 (in datasets with around 50 to 100 images). Usually I try to reduce the LR and see if it converges to optimal results in the same number of steps.  `average_key_norm` tends to grow proportional to the LR used, but this is affected by the size and diversity of the dataset, with larger ones it seems to grow more slowly. You could also try to use max norm regularization to penalize weights growing above a certain threshold, by setting `scale_weight_norms` to something like 4.0 or 5.0.

I suspect it would be even more effective to set distinct `scale_weight_norms` at block level but that is currently not implemented.

araleza:
Okay, so I have so more good news: my training run at 1e-4 with @recris's Huber loss went very well. Now I have some very high quality images, with the concept in the training images learned close to perfectly. (It's still training).

The big thing to note is how long it took. Training with gradient accumulation 16 at 1e-4 took pretty much a whole day of training to get it to work. But it did make continuous steady progress throughout this time. Without Huber loss, I either found the learning of the training concept would get stuck and sample images would continually mis-draw the objects being learned, or that the sample images would lose their high image quality and draw boring scenes. That doesn't seem to be the case with Huber loss.

@recris, I've heard things about high gradient accumulation being bad too, but it hasn't seem to have affected me here. Before Huber loss was implemented, I pushed up from accumulating 8 to accumulating 16 to try to reach better image quality, and that did seem to work. But yeah, gradient accumulation feels a bit suspect in general compared to increasing batch size. It does use less memory though, and another avantage to gradient accumulation that people often miss is that batch size only works on images that are in the same bucket for image dimensions. And often buckets only have single images in them, silently preventing batch size from working for them. Gradient accumulation is immune to that.

Anyway, I'm glad that this PR has made it into the SD3 branch, as it seems worthwhile to me.

StableLlama:
All insights about how to use it are great - but having them in this PR makes it very hard to use them in future.
So please put it also in a documentation file so that it can be looked up anytime in the future!

FurkanGozukara:
@bmaltais can we have this i would like to test this ty

araleza:
> You could also try to use max norm regularization to penalize weights growing above a certain threshold, by setting scale_weight_norms to something like 4.0 or 5.0.

Hey there @recris, I gave this a try (with 4.0), and yeah that lets me get to the higher LRs without those vertical lines; great suggestion.

I hope we can get individual LRs for individual blocks at some point, like we had for SDXL with `--block_lr=...`. That would probably mean that whatever keys are growing very fast wouldn't get so big and need clipped in length. But in the meantime, renormalizing them seems to do the job.

araleza:
Okay, I found something I think is pretty cool, and I thought I'd put it here cause it relates to what's been said so far.

I found that using `--scale_weight_norms` did get rid of the vertical lines on the image when I used higher LRs... but it also seemed to limit the amount of learning that occurred. The object I was trying to learn didn't contain all the fine details that I wanted.

I tried pushing up the value for `--scale_weight_norms`, but then my images became 'constrasty', with faces often being bright and washed out.

I suspected that the issue was that the later single blocks were getting overtrained, and they can't deal with the higher key lengths - but the earlier double blocks were not overtrained and needed the longer key lengths available to them to be able to learn the trained object.

So I added this code to `lora_flux.py`: (round about line 1145)
```
            # Attempt a different max scaling for single blocks
            if downkeys[i].find('single') != -1:
                this_max_norm_value = 1
            else:
                this_max_norm_value = max_norm_value
```
and changed the next couple of lines to use the new `this_max_norm_value` instead of `max_norm_value`:
```
            norm = updown.norm().clamp(min=this_max_norm_value / 2)

            desired = torch.clamp(norm, max=this_max_norm_value)
```
This seemed to immediately get rid of the constrasty look, and let me increase my value for `--scale_weight_norms`, allowing the object I was learning to be learned far more accurately.

Maybe we need a `--scale_weight_norms_single 1.0` option?

recris:
Like you said before, we need a way to control the learning rate at block level. I find it too easy to create a burned Lora before it properly learns the target concept(s).

Alternatively, there is a chance this could be mitigated with a better loss function - maybe the work in https://github.com/kohya-ss/sd-scripts/discussions/294 should be revisited and translated to Flux training.

recris:
@araleza If you're training a style you may wish to have a look at https://github.com/kohya-ss/sd-scripts/pull/1838 , it could help improving learning issues.

recris:
After playing a bit more with the parameters I think some of the image artifacts from training with higher learning rates can be mitigated with a more aggressive decay in the exponential schedule.

 Something like `huber_c = 1.0`  and `huber_scale = 2.5`.

YMMV, this are from limited tests of mine.

araleza:
> After playing a bit more with the parameters I think some of the image artifacts from training with higher learning rates can be mitigated with a more aggressive decay in the exponential schedule.
> 
> Something like `huber_c = 1.0` and `huber_scale = 2.5`.
> 
> After training with these I don't get as much "white band" artifacts on the image when I scale the LoRA strength past 100%, or use multiple LoRAs simultaneously.
> 
> YMMV, this is from limited tests of mine.

These numbers are working great for me, @recris . And as an added bonus, it turns out that even though I've fully trained some LoRAs with L2 loss, switching to Huber pretty much immeditately boosts the quality of the output. It seems that the LoRAs don't have to be trained with Huber from the start.

recris:
When using multi-resolution training I find it useful to scale the loss weight by the batch size. If you're using different batch sizes per resolution (to fit available VRAM) it may lead to "unbalanced" learning because learning scales differently with the size of the batch.

There seems to be some empirical evidence to support that learning rate should scale with `sqrt(batch_size)`.

I've been testing this change to the code (before `post_process_loss` is called in `train_network.py`):
```
loss = loss * math.sqrt(loss.shape[0])
```
It seems to have helped a lot with the artifacting.

araleza:
Nice one @recris, good tip. I might get a change to try that soon when (if) I get a 5090 and can push my batch size up from  1. (I chose my LoRA rank to just fit in VRAM, so I can't use bigger batch sizes than 1 right now)

I did find another way to avoid those vertical stripes. If you start with a weight_decay that's significantly larger than usual (e.g. 2.0 to 3.0) then the artifacts show up less. And once the network has learned what it can at that high weight_decay, it's possible to then reduce the weight_decay back to something more useful for detailed learning (e.g. 1.0, or 0.1) and continue training from there without the line artifacts appearing, even though they would have done so if you'd started training from that lower value.

I've started seeing weight_decay as effectively a softer form of max_key_norm.

I'm do think there's a chance the vertical line artifact is actually a bug in Flux. As far as I know, the right hand side of the render should be no different to the bottom edge, but the bottom edge doesn't show that artifact. Makes me think there might be a missing +1 or -1 somewhere in the Flux training code, maybe even at Black Forest Labs. (It might not be a bug of course, maybe the fact the lines are vertical comes from the training data which has more vertical objects in it.)

araleza:
Hey, so I just made a new discovery regarding those vertical lines, and I thought I'd share it here.

I noticed that no matter how low I set my LR, and how big my batch size, the lines would appear at some point, usually before the training concept had successfully been learned. And then I realized that this might be due to the fact my alpha_masks look like this:

image

Since the line artifacts run down the right hand side of the image, then they pretty much only appear in the 5% attention areas of my training images. (The reason I use 5% alpha and not 0% alpha was to try to prevent the network from learning to maybe fill nonsense into these areas, such as duplicating the object being trained, as there would be no negative cost from doing this).

I think there's something about the vertical line artifact that's a side effect of the network focusing on producing the center part of the image rather than the whole image. Maybe it's able to improve the quality of the center if it sacrifices the side.

I put the theory to the test by just swtiching off `--alpha_mask`, and yes, the line artifact quickly faded out, even though I had raised the LR.

I'm thinking that maybe introduce a strip of 100% alpha down the right hand side of my training images, wide enough to cover the area where the artifact usually appears (even after bucketing might chop the image to be a bit smaller).

FurkanGozukara:
in my experience vertical lines appear when i upscale lora

when i upscale dreambooth trained models up to 2x there are no lines

recris:
I've had the same artifact before, and adding a small nonzero weight to the non masked areas didn't solve the issue for me. The problem seems to be worse when training with aggressive learning rates and/or low batch sizes.

The way I "fixed" the problem was by tweaking the noise shift level during inference (ModelSamplingFlux node in ComfyUI)

Issue #1796, opened 2024-11-19
Prodigy + ScheduleFree


hablaba:
Hello! How might I use this optimizer? https://github.com/LoganBooker/prodigy-plus-schedule-free

Would I use the schedulefree wrapper with it somehow?

rockerBOO:
Probably need the schedule-free branch. Then add the optimizer module from pip `pip install prodigy-plus-schedule-free`. Then 

```bash
--optimizer_type "prodigyplus.ProdigyPlusScheduleFree"
```

https://github.com/LoganBooker/prodigy-plus-schedule-free/blob/main/prodigyplus/prodigy_plus_schedulefree.py#L47C1-L99C28

And the options can be passed in optimizer_args

```bash
--optimizer_args "prodigy_steps=1000"
```

It should output the optimizer it's using in the logs to confirm.

rockerBOO:
I tested it and needed some fixes. I made #1799 for that. Seems to work at least.

hablaba:
Oh awesome, appreciate your work here. Yeah when I had tried it I got that error that it wasn’t supported and then I just had other things I was working on. I do think there’s a ton of potential though using schedule free prodigy

bghira:
that learning rate doesn't look very "dynamic", it never goes back down?

PR #1799, opened 2024-11-20
Allow unknown schedule-free optimizers to continue to module loader

rockerBOO:
To support [ProdigyPlusScheduleFree](https://github.com/LoganBooker/prodigy-plus-schedule-free)
```bash
pip install git+http://github.com/LoganBooker/prodigy-plus-schedule-free
```

```
--optimizer_type prodigyplus.ProdigyPlusScheduleFree
```

```
                    INFO     use prodigyplus.ProdigyPlusScheduleFree | {'weight_decay': 0.1}                                         train_util.py:4111
[Prodigy+ScheduleFree] Optimiser contains single param_group -- 'split_groups' has been disabled.
```

Related #1796

PR #1811, opened 2024-11-29
Allow unknown schedule-free optimizers to continue to module loader

rockerBOO:
To support [ProdigyPlusScheduleFree](https://github.com/LoganBooker/prodigy-plus-schedule-free)
```bash
pip install git+http://github.com/LoganBooker/prodigy-plus-schedule-free
```

```
--optimizer_type prodigyplus.ProdigyPlusScheduleFree --optimizer_args "prodigy_steps=700"
```

```
optimizer_type = "prodigyplus.ProdigyPlusScheduleFree"
optimizer_args = ["prodigy_steps=700"]
```

`prodigy_steps` indicates when it will stop adapting the LR. Suggested 25% of training steps or down to 5-10%. The value will depend on your dataset and may not be necessary. Helps prevent some amount of increasing d over the training run. 

Full set of usage options: https://github.com/LoganBooker/prodigy-plus-schedule-free?tab=readme-ov-file#usage

Related #1796 #1799

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -4600,7 +4600,7 @@ def resume_from_local_or_hf_if_specified(accelerator, args):
 
 def get_optimizer(args, trainable_params):
     # "Optimizer to use: AdamW, AdamW8bit, Lion, SGDNesterov, SGDNesterov8bit, PagedAdamW, PagedAdamW8bit, PagedAdamW32bit, Lion8bit, PagedLion8bit, AdEMAMix8bit, PagedAdEMAMix8bit, DAdaptation(DAdaptAdamPreprint), DAdaptAdaGrad, DAdaptAdam, DAdaptAdan, DAdaptAdanIP, DAdaptLion, DAdaptSGD, Adafactor"
-
+    
     optimizer_type = args.optimizer_type
     if args.use_8bit_adam:
         assert (
@@ -4874,6 +4874,7 @@ def get_optimizer(args, trainable_params):
         optimizer = optimizer_class(trainable_params, lr=lr, **optimizer_kwargs)
 
     elif optimizer_type.endswith("schedulefree".lower()):
+        should_train_optimizer = True
         try:
             import schedulefree as sf
         except ImportError:
@@ -4885,10 +4886,10 @@ def get_optimizer(args, trainable_params):
             optimizer_class = sf.SGDScheduleFree
             logger.info(f"use SGDScheduleFree optimizer | {optimizer_kwargs}")
         else:
-            raise ValueError(f"Unknown optimizer type: {optimizer_type}")
-        optimizer = optimizer_class(trainable_params, lr=lr, **optimizer_kwargs)
-        # make optimizer as train mode: we don't need to call train again, because eval will not be called in training loop
-        optimizer.train()
+            optimizer_class = None
+
+        if optimizer_class is not None:
+            optimizer = optimizer_class(trainable_params, lr=lr, **optimizer_kwargs)
 
     if optimizer is None:
         # 任意のoptimizerを使う
@@ -4989,6 +4990,10 @@ def get_optimizer(args, trainable_params):
     # for logging
     optimizer_name = optimizer_class.__module__ + "." + optimizer_class.__name__
     optimizer_args = ",".join([f"{k}={v}" for k, v in optimizer_kwargs.items()])
+
+    if hasattr(optimizer, 'train') and callable(optimizer.train):
+        # make optimizer as train mode: we don't need to call train again, because eval will not be called in training loop
+        optimizer.train()
 
     return optimizer_name, optimizer_args, optimizer
 
--- a/train_network.py
+++ b/train_network.py
@@ -61,6 +61,7 @@ class NetworkTrainer:
         avr_loss,
         lr_scheduler,
         lr_descriptions,
+        optimizer=None,
         keys_scaled=None,
         mean_norm=None,
         maximum_norm=None,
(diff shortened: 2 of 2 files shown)


rockerBOO:
> Thank you, this is great! Is the reason you install directly from git instead of from PyPi because it's better to use the latest version? Also, do you think it's okay to close #1796?

It wasn't on PyPi when I made this, so using that version would be suggestible. Closes #1796.

StableLlama:
It would be great when you'd also add a documentation here about the best way to use it.
Linking in this PR to a different source will not transfer the relevant information to the users of this repository.

rockerBOO:
> It would be great when you'd also add a documentation here about the best way to use it. 

This PR unblocks prodigyplus but I do not have a recommendation of how best to use it. With any suggestions I will update the original description to allow users to follow.  

Initially it's like prodigy, so setting the LR to 1.0 for text encoders and UNet. It is schedulefree so no LR schedulers are used or required. 

`--text_encoder_lr 1.0 --unet_lr 1.0`

araleza:
Hey, I gave this a try, combined with Huber loss, and I wanted to report my results.

I have around 200 training images, and I'm using batch size 1 / no gradient accumulation. I used your default settings as suggested (learning rate 1.0, I'm not training the text encoders).

The image quality was superb, and seemed to be learning slowly over time. But around the 3 hour mark (around step 1000), despite the images still being very high quality in appearance, I started getting the (incredibly annoying) grid artifact:

image

It appeared most strongly down the right hand side of the image, as it usually does when it appears.

I might restart and set the learning rate to 5e-1 instead of 1e-0 (1.0).

By the way, I'm getting pretty suspicious of the fact that this artifact appears most strongly at one side of the image in particular. Maybe some sort of offset bug somewhere? If someone was able to hunt down the true model-level cause of this artifact, that would be valuable. I don't think it's just 'your learning rate is too high', as I think using a lower learning rate might just be masking the problem, rather than the cure for it.

araleza:
Just in case that previous message of mine came across as a bit negative, the sample images I saw before the artifact showed up were very sharp and had lots of rich detail. I don't think this grid artifact is due to any error that @rockerBOO has made, but is just a pre-existing issue. I think this optimizer seems might be very useful for raising image quality.

araleza:
Yeah, even dropping the LR down to 0.3 still produces the vertical stripes near the right hand edge at around about 1000 steps:

image

rockerBOO:
> Given that `"prodigy_steps=700"` has been recommended as an optimizer parameter, and I see the stripes happen around around step 1000

I have removed the suggestion of prodigy_steps here as it's absolutely dependent on how it's training and how many steps your training is running. Suggested from the module is maybe 25% or 5-10% but it's a parameter if you want to disable it from continuing to modify it's d after that percentage of steps for your training. 

See the [linked repo](https://github.com/LoganBooker/prodigy-plus-schedule-free) for the actual recommendations, as this PR only unblocks this optimizer. Any suggestions or improvements to the algorithms would be best informed to the developer.

araleza:
Yeah, knocking down the prodigy_steps from 700 to 550 has gotten rid of that corruption for me. My total intended steps is 7400, so that's down from 9.5% to 7.4%.

Maybe it's lower for me than most people as most of my images are using an alpha_mask of around 5%, with typically only small parts at 100% alpha.

araleza:
I looked at the Prodigy docs, rather than just the schedule-free version, and I found this:

image
(From https://github.com/konstmish/prodigy/blob/main/README.md )

If people are experimenting with this optimizer, it might be worth trying setting use_bias_correction to true, as recommended there. Although I'm not certain whether these pages are talking about DiT when they say 'diffusion models', or if they specifically mean Unet-based diffusion. You could also try setting that weight decay value, as the default is zero.

You can set that value to true in the --optimizer_args.

deGENERATIVE-SQUAD:
I'm very sorry, but I don't know much about code. Could you explain step by step how to implement this optimizer in sd-scripts and use its arguments in a batch file? All my attempts were unsuccessful in either dev or sd3, only ValueError: too many values to unpack.

rockerBOO:
UPD. The problem in --optimizer_args if I use it with arguments. Without --optimizer_args the train process runs. How can I control args without --optimizer_args?

the arguments are passed like `--optimizer_args "prodigy_steps=1000" "weight_decay=0.01"`. or `--optimizer_args prodigy_steps=1000 weight_decay=0.01`. Note they are separated by spaces and no spaces in the arguments. Using the quotes allows you to use spaces if you want.

deGENERATIVE-SQUAD:
> the arguments are passed like `--optimizer_args "prodigy_steps=1000" "weight_decay=0.01"`. or `--optimizer_args prodigy_steps=1000 weight_decay=0.01`. Note they are separated by spaces and no spaces in the arguments. Using the quotes allows you to use spaces if you want.

Thank you for the answer, will check later! For the future, how can I find out from the optimizer settings how to write arguments in a right way?

Btw, could you tell me how to make the original ADOPT optimizer work from https://github.com/iShohei220/adopt to SD-scripts? I can see that ADOPT can be enabled in the ProdigyPSF settings, but I'd like to try the original as well.

rockerBOO:
> Thank you for the answer, will check later! For the future, how can I find out from the optimizer settings how to write arguments in a right way?

See the [train_network docs](https://github.com/kohya-ss/sd-scripts/blob/main/docs/train_README-ja.md#%E3%82%AA%E3%83%97%E3%83%86%E3%82%A3%E3%83%9E%E3%82%A4%E3%82%B6%E3%81%AE%E6%8C%87%E5%AE%9A%E3%81%AB%E3%81%A4%E3%81%84%E3%81%A6) but may need to translate it. 

> Btw, could you tell me how to make the original ADOPT optimizer work from https://github.com/iShohei220/adopt to SD-scripts? I can see that ADOPT can be enabled in the ProdigyPSF settings, but I'd like to try the original as well.

See the linked repo for the argument you need to pass it.

deGENERATIVE-SQUAD:
> See the [train_network docs](https://github.com/kohya-ss/sd-scripts/blob/main/docs/train_README-ja.md#%E3%82%AA%E3%83%97%E3%83%86%E3%82%A3%E3%83%9E%E3%82%A4%E3%82%B6%E3%81%AE%E6%8C%87%E5%AE%9A%E3%81%AB%E3%81%A4%E3%81%84%E3%81%A6) but may need to translate it.
Mm okay.
 
> See the linked repo for the argument you need to pass it.
I dont understand how, thats the problem.

Can you clarify my UPD in my recent post also, please?

There is also a question, in the original Prodigy you need to use decouple=True use_bias_correction=True safeguard_warmup=True, does it work with SF? Well, except for bias, it is in SF by default, but it is turned off.

deGENERATIVE-SQUAD:
And the last question for now: need I separate lr for unet and TE in SF or is it not important/no sense in this case?

deGENERATIVE-SQUAD:
And the last question for now: can I lower the separated lr for TE or is it not important/no sense in this case of SF using?

FurkanGozukara:
@rockerBOO @kohya-ss  when training with prodigyplus.ProdigyPlusScheduleFree

lets say 1 model trained 1000 steps and saved checkpoint at 500 and 1000

another one trained 500 steps and saved checkpoint at 500 steps

so the first model 500 steps checkpoint will be wildly different than the second model 500 steps?

i mean it works according to the total number of steps ? what is the logic

thank you

PR #1821, opened 2024-12-04
Unify controlnet parameters name and change scripts name.

sdbds:
Multiple versions of the controlnet training script use three different parameter names, so the parameters are unified.
`--controlnet_model_path`
As well as renaming the initial version of the controlnet script to `train_control_net.py`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,42 @@
+
+name: Python package
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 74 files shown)


sdbds:
> Thank you for this! This makes sense.
> 
> However, it seems that this will be a breaking change when it is merged into the main branch in the future. I think breaking changes are allowed on the dev/sd3 branch. Therefore, although the name is not good, I think it is better to use `controlnet_model_name_or_path` used in `train_controlnet.py`.

OK, I'll use  `controlnet_model_name_or_path`.

PR #1870, opened 2025-01-10
Add rank_stabilized for networks

rockerBOO:
https://arxiv.org/abs/2312.03732
https://huggingface.co/blog/damjan-k/rslora

Screenshot 2025-01-09 at 22-08-13 A Rank Stabilization Scaling Factor for Fine-Tuning with LoRA - 2312 03732v1 pdf
Screenshot 2025-01-09 at 22-08-00 A Rank Stabilization Scaling Factor for Fine-Tuning with LoRA - 2312 03732v1 pdf

Inspired by https://github.com/KohakuBlueleaf/LyCORIS/pull/172

### Usage 

`network_args=["rank_stabilized=True"]`
`--network_args rank_stabilized=True`

### DIFF
--- a/networks/dylora.py
+++ b/networks/dylora.py
@@ -31,7 +31,7 @@ class DyLoRAModule(torch.nn.Module):
     """
 
     # NOTE: support dropout in future
-    def __init__(self, lora_name, org_module: torch.nn.Module, multiplier=1.0, lora_dim=4, alpha=1, unit=1):
+    def __init__(self, lora_name, org_module: torch.nn.Module, multiplier=1.0, lora_dim=4, alpha=1, unit=1, rank_stabilized=False):
         super().__init__()
         self.lora_name = lora_name
         self.lora_dim = lora_dim
@@ -48,7 +48,10 @@ class DyLoRAModule(torch.nn.Module):
         if type(alpha) == torch.Tensor:
             alpha = alpha.detach().float().numpy()  # without casting, bf16 causes error
         alpha = self.lora_dim if alpha is None or alpha == 0 else alpha
-        self.scale = alpha / self.lora_dim
+        rank_factor = self.lora_dim
+        if rank_stabilized:
+            rank_factor = math.sqrt(rank_factor)
+        self.scale = alpha / rank_factor
         self.register_buffer("alpha", torch.tensor(alpha))  # 定数として扱える
 
         self.is_conv2d = org_module.__class__.__name__ == "Conv2d"
@@ -285,6 +288,7 @@ class DyLoRANetwork(torch.nn.Module):
         unit=1,
         module_class=DyLoRAModule,
         varbose=False,
+        rank_stabilized=False,
     ) -> None:
         super().__init__()
         self.multiplier = multiplier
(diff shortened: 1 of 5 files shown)


recris:
Does the scale in `apply_max_norm_regularization` also have to be changed to work correctly?

rockerBOO:
@recris good point, there is a scale value in there. I will update it and test.

araleza:
This sounds interesting. Has anyone tried it out and seen any noticeable improvements in output quality when training LoRAs with this PR in place?

I might give it a test myself...  I'm training a rank 52 LoRA cause that's what I can get working on my 4090 RTX and still be able to use my PC.  Not sure if that's a large enough rank to benefit from this change or not.

rockerBOO:
Haven't done A/B tests with it just yet but it should have marginal improvements likely. Maybe some better stability for larger ranks though. I use it all the time though, it's a marginal change so no harm in using it or not. Once I am on a faster setup for training I will do some more examples of a/b testing.

recris:
Maybe I am misunderstanding the paper & implementation, but isn't this the same as dividing the value of `network_alpha` by `sqrt(network_dim)` ahead of time, in the config? Or maybe just use `network_alpha = sqrt(network_dim)` all the time.

Or are we supposed to use an effective "alpha" value during training that is different during inference? Wouldn't that force the user to scale up the LoRA strength to compensate?

rockerBOO:
yeah i think you have a point in that at inference this scaling wouldn't be normally applied. So maybe we can do 

`network_alpha = (alpha / math.sqrt(rank)) * rank`

```
alpha = 16
rank = 32

# During training
rsLoRA_scale = alpha / math.sqrt(rank)  # ~2.83

# When saving for inference with standard LoRA code
# We need: network_alpha/rank = rsLoRA_scale
# So: network_alpha = rsLoRA_scale * rank

network_alpha = (alpha / math.sqrt(rank)) * rank
# network_alpha = alpha * (rank/sqrt(rank)) = alpha * sqrt(rank)
# network_alpha = 16 * sqrt(32) = 16 * 5.66 ≈ 90.5

# network_alpha / rank
# 2.82842712474619
```

recris:
This is very interesting... I've been using `network_dim = network_alpha` for a long time (Flux.1 dev), it gets good results but some times the output comes out degraded and I've never been really sure why.

Today I tried to use `network_alpha = sqrt(network_dim)` (dim=16), and the degradation appears to be gone?! It seems that using a proper ratio really has some impact in the learning.

MegaCocos:
> This is very interesting... I've been using `network_dim = network_alpha` for a long time (Flux.1 dev), it gets good results but some times the output comes out degraded and I've never been really sure why.
> 
> Today I tried to use `network_alpha = sqrt(network_dim)` (dim=16), and the degradation appears to be gone?! It seems that using a proper ratio really has some impact in the learning.

Sorry for nub's question, I would like to try it too, how should i implement it for training? I have standard sd-scripts from sd3 branch. Do I need to edit some script or can it be changed in training command by some options?

rockerBOO:
> This is very interesting... I've been using `network_dim = network_alpha` for a long time (Flux.1 dev), it gets good results but some times the output comes out degraded and I've never been really sure why.
> 
> Today I tried to use `network_alpha = sqrt(network_dim)` (dim=16), and the degradation appears to be gone?! It seems that using a proper ratio really has some impact in the learning.

Maybe a factor where the original implementations suggested half alpha to rank to deal with the scaling of the activations. So using 0.25 scaling in your example but different than 16/sqrt(16) = 4 scaling. Maybe different things here but maybe 8/sqrt(16) = 2. Probably vastly different results from this paper and your issues, but the idea I think would be good to implement if using  rank stabilized. So alpha=(16/sqrt(16) * 16) = 64. So would be like alpha = 64 rank = 16 without this PR.

Issue #1891, opened 2025-01-22
Add Support for Flex.1 Alpha


stepfunction83:
Creating an Issue from the related discussion: https://github.com/kohya-ss/sd-scripts/discussions/1882

Ostris has created a new Flux model that is both distilled to a lower parameter count and has a custom trained guidance module which allows the model to be much more easily and flexibly finetuned. 

To train this, the newly created guidance module needs to be bypassed during training and then restored afterwards. This is currently implemented in ai-toolkit and would need to be implemented in sd-scripts to be able to support training of the model.

This seems to be the key snippet of code to bypass the guidance block in ai-toolkit:

```
def guidance_embed_bypass_forward(self, timestep, guidance, pooled_projection):
    timesteps_proj = self.time_proj(timestep)
    timesteps_emb = self.timestep_embedder(
        timesteps_proj.to(dtype=pooled_projection.dtype))  # (N, D)
    pooled_projections = self.text_embedder(pooled_projection)
    conditioning = timesteps_emb + pooled_projections
    return conditioning

# bypass the forward function

def bypass_flux_guidance(transformer):
    if hasattr(transformer.time_text_embed, '_bfg_orig_forward'):
        return
    # dont bypass if it doesnt have the guidance embedding
    if not hasattr(transformer.time_text_embed, 'guidance_embedder'):
        return
    transformer.time_text_embed._bfg_orig_forward = transformer.time_text_embed.forward
    transformer.time_text_embed.forward = partial(
        guidance_embed_bypass_forward, transformer.time_text_embed
    )

# restore the forward function

def restore_flux_guidance(transformer):
    if not hasattr(transformer.time_text_embed, '_bfg_orig_forward'):
        return
    transformer.time_text_embed.forward = transformer.time_text_embed._bfg_orig_forward
    del transformer.time_text_embed._bfg_orig_forward
```

`bypass_flux_guidance` is applied before noise prediction, while `restore_flux_guidance` is applied after noise prediction and also before sampling.

stepfunction83:
This was also implemented into diffusion-pip through this commit: https://github.com/tdrussell/diffusion-pipe/commit/90ab3f957aae8ea862b379707d9e0cd8dfbea220

stepfunction83:
I've created a minimal working example here:

https://github.com/stepfunction83/sd-scripts/tree/sd3

With this commit just brute forcing in the relevant code snippets from ai-toolkit:

[...]

PR #1893, opened 2025-01-23
Implement Flex.1 Alpha Functionality

stepfunction83:
Create --bypass_flux_guidance argument and mechanism to allow training of the Flex.1-Alpha model created by Ostris (https://huggingface.co/ostris/Flex.1-alpha). This is an Apache 2.0 model based off of Flux Schnell with the distillation removed and a custom guidance module created, which allows the model to be more easily finetuned.

Initial testing I've done confirms that this is in fact the case and the model is extremely easy to finetune compared to Flux Dev, with a minor deterioration in quality.

The implementation here is based off of this commit of ai-toolkit: https://github.com/ostris/ai-toolkit/commit/8ef07a9c3662351328407d9249a5418639dd3052

Currently, I have an issue created here in reference to this with some initial testing: https://github.com/kohya-ss/sd-scripts/issues/1891

### DIFF
--- a/flux_minimal_inference.py
+++ b/flux_minimal_inference.py
@@ -20,6 +20,8 @@ from library import device_utils
 from library.device_utils import init_ipex, get_preferred_device
 from networks import oft_flux
 
+from library.flux_utils import bypass_flux_guidance, restore_flux_guidance
+
 init_ipex()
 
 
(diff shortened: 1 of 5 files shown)


stepfunction83:
Apparently this is a lot more difficult than I anticipated since the Flux model is stored differently in sd-scripts to how it is stored in ai-toolkit.

stepfunction83:
Apparently this might be a lot simpler than I thought since there is already a Flux.params.guidance toggle to turn off guidance selectively.

stepfunction83:
I would definitely love to hear comments or improvements. This is my first PR here and  my knowledge of the inner workings of sd-scripts and of coding for neural nets in torch is weak at best, so this is really just me poking around in the dark.

stepfunction83:
I can confirm from my testing that this does appear to be working correctly. My results from doing test runs of both full finetunes and LoRA seem to be consistent with each other and substantially better than a naive run without the bypass mechanism applied.

PR #1895, opened 2025-01-23
Add more dpmsolver++2m samplers to solve SDXL sampling artifacts

rockerBOO:
Visual artifacting for sampling using `dpmsolver++`. This adds more samplers but `dpmsolver++_2m_lu` should produce better results on SDXL. 

sample_2_880_1c033a0dba24b098ea6d

https://github.com/huggingface/diffusers/issues/5433
https://github.com/huggingface/diffusers/pull/5541

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -5980,6 +5980,31 @@ def get_my_scheduler(
     elif sample_sampler == "dpmsolver" or sample_sampler == "dpmsolver++":
         scheduler_cls = DPMSolverMultistepScheduler
         sched_init_args["algorithm_type"] = sample_sampler
+    elif sample_sampler == "dpmsolver++_2m":
+        scheduler_cls = DPMSolverMultistepScheduler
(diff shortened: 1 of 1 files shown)


rockerBOO:
I thought I did this correctly as referred to in the linked issues, but there are a few things I need to figure out. I can't easily test this outside of training so will take some time to fiddle with it. The schedulers are having a set value being passed but not sure why. I looked back in the history but did not find any indication of why, except it is using the defaults from one of the schedulers. 

```
    scheduler = scheduler_cls(
        num_train_timesteps=SCHEDULER_TIMESTEPS,
        beta_start=SCHEDULER_LINEAR_START,
        beta_end=SCHEDULER_LINEAR_END,
        beta_schedule=SCHEDLER_SCHEDULE,
       ...
```

So I am trying some experiments with it removed but it will cause samples to be different than before with the same settings, which can make it difficult to compare. At least find something that works better for SDXL sampling and then we can work back from there.

Issue #1897, opened 2025-01-24
Implement Stable Loss Calculation for Run Tracking


stepfunction83:
I propose creating a tracker to capture a stable loss measurement as proposed here: https://github.com/spacepxl/demystifying-sd-finetuning

Effectively, at regular intervals during training, a preselected image (or batch of images I imagine) and a preselected noise seed are used to calculate the loss. This ensures that over the course of training, the loss recorded can accurately show the progress of the training run.

By also incorporating a holdout set to use for calculating validation loss, this allows a proper evaluation of what point the model begins to overtrain.

stepfunction83:
It looks like it could be implemented as part of the standard train loop, potentially by creating two additional dataloaders: 

- `test_dataloader` for a test set selected from the dataset itself
- `val_dataloader` for a validation set held out from the dataset and specified separately

It may be additionally helpful to specify seeds for these dataloaders with additional seeds for their respective noise generation processes.

I imagine this could be done with something along the lines of:

`--calc_test_val_loss` (bool, enable test/validation loss calculation)
`--test_dataloader_seed 42`
`--val_dataloader_seed 42`
`--val_dataloader_dir "~/val_dataset_dir/"`

Finally, a frequency should be specified to run loss calculations on these sets:

`--test_val_loss_freq_steps 50` (calculate test/val loss every 50 steps)
`--test_val_loss_freq_epochs 1` (calculate test/val loss every epoch)

stepfunction83:
Implemented in #1899. I am working on a number of enhancements to it in #1900.

PR #1898, opened 2025-01-26
Implement Test and Validation Loss for Flux Finetuning and LoRA Training

stepfunction83:
As discussed in the issue I created here: https://github.com/kohya-ss/sd-scripts/issues/1897

This is based off of the work done by spacepxl here: https://github.com/spacepxl/demystifying-sd-finetuning

The general point of it is that traditionally, the loss that is calculated at each step is more or less fully random and conveys very little information about the actual progress of training and very little about the performance. 

Instead, we can calculate test and validation loss values as one would when doing traditional machine learning. By selecting a specific set of samples and running them with identical noise and timesteps over time, we can evaluate a test loss for the model. Further, by selecting a portion of the training set to hold out, we can calculate a validation error as well as identify when the model begins overtraining.

This means that until this capability becomes available, we are effectively just guessing at the progress of our training and have no real quantitative measurement of progress or of overtraining.

In this PR, I implement this capability for Flex Finetuning and LoRA training which allows for these two sets to be automatically identified and for the two loss values to be captured at regular intervals.

So far, I've done several runs to verify that it is working properly, and you can see a brief snippet of this here:
image
(Pictured: Test loss relative to initial value, Validation loss relative to initial value)

The test error relative to the initial test error is on the left, while the respective validation results are on the right. As opposed to the raw loss values (below), you can see they are much smoother and more directionally indicative of training progress.
image
(Pictured: Raw loss values, i.e. garbage)

To use the functionality, I have added 5 new arguments:
--test_step_freq 25 (The frequency at which test loss is calculated; 0 turns off feature)
--val_step_freq 25 (The frequency at which validation loss is calculated; 0 turns off feature)
--test_set_count 10 (The number of training samples used for the test set)
--val_set_count 10 (The number of training samples used for the validation set; These are then skipped in training since they are held out)
--test_val_repeat_count 6 (The number of noise/timestep iterations to use for each test and validation sample)

I will say that this required some fairly sizable changes to the training loops to accommodate the additional loss calculations seamlessly, so I would definitely like feedback regarding the approach. I'm also not very familiar with accelerate and pytorch (and am a middling programmer to boot), so I may be doing something wrong along the way and almost certainly could benefit from some best practices. From my initial tests however, things at least seem to be working appropriately and generating the desired result.

This is also limited to just Flux for now, but the general logic that I used could be applied more broadly to other training types.

### DIFF
--- a/flux_train.py
+++ b/flux_train.py
@@ -716,6 +716,7 @@ def train(args):
             if global_step % test_step_freq == 0 and test_step_freq > 0:
                 test_loss, test_fixed_states = train_util.calc_test_val_loss(dataset=test_set, loss_func=calculate_loss, repeat_count=test_val_repeat_count, fixed_states=test_fixed_states, test=True)
                 test_losses.append(test_loss)
+                accelerator.log({'test_loss':test_loss}, step=global_step)
 
             # CALCULATE LOSS ON VALIDATION SET AT TEST SET FREQUENCY
             if global_step==0:
@@ -724,6 +725,7 @@ def train(args):
             if global_step % val_step_freq == 0 and val_step_freq > 0:
                 val_loss, val_fixed_states = train_util.calc_test_val_loss(dataset=val_set, loss_func=calculate_loss, repeat_count=test_val_repeat_count, fixed_states=val_fixed_states, test=False)
                 val_losses.append(val_loss)
+                accelerator.log({'val_loss':val_loss}, step=global_step)
 
             # STANDARD LOSS CALCULATION
             loss, _ = calculate_loss(batch, accumulate_loss=True) # Loss should be accumulated when not running the test/validation samples though
--- a/flux_train.py
+++ b/flux_train.py
@@ -579,8 +579,11 @@ def train(args):
         # log empty object to commit the sample images to wandb
         accelerator.log({}, step=0)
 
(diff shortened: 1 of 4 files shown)


stepfunction83:
> Thank you for this! Unfortunately, it seems to be a duplicate of #1864. However, your approach for `calculate_loss` is very simple and great. I think we can simplify the code this way in the future.
> 
> However, I think there is one drawback to using the top n items of the dataset. That is, the data returned by the dataloader is shuffled every time. Therefore, if I understand this PR correctly, the current code will be testing/validating the same data as the training data, and the correct loss will not be obtained.
> 
> To address this issue, we need to have an independent validation dataset, which is what we do in #1864.

So in this, I am calculating loss for both a test and a validation set. Both are selected from the train_dataloader, and both are tested using a consistent set of timesteps and noise over time to ensure consistent loss calculation. The primary difference is that for the validation set, the train loop automatically skips over those samples when it encounters them in training, so they do not affect the final model. You can see that here at the beginning of the train loop:

```
if args.val_step_freq > 0:
                if step in val_set['steps']: # Skip validation steps, don't increment global step
                    logger.warning('SKIPPING BATCH IN VALIDATION SET')
                    continue
```

stepfunction83:
I'll close this for now and begin migrating some of the ideas into another PR using the new framework.

PR #1900, opened 2025-01-26
Validation Loss Enhancements

stepfunction83:
I am creating this PR to implement a number of additional enhancements to the base validation loss implementation. These include:

**High Effort**
- Implement for finetuning
- Capture loss calculation state to reduce variance of calculation
- Add test set loss calculation

**Medium Effort**
- Add multiple noise/timestep iterations per sample

**Lower Effort**
- Add specific validation sample count in addition to current % based selection
- Add relative loss (loss / initial loss value)
- (Complete)  Calc validation at start with `--validation_at_start` argument
- (Complete)  Clean up progress bars so the validation one sits neatly above the global one and is not recreated each time

**Bug Fixes**
- Make it so it doesn't recalculate train latents every run
- (Complete) Correct bug for not working for Flux LoRA training if latents are not cached

### DIFF
--- a/flux_train_network.py
+++ b/flux_train_network.py
@@ -323,7 +323,7 @@ class FluxNetworkTrainer(train_network.NetworkTrainer):
         self.noise_scheduler_copy = copy.deepcopy(noise_scheduler)
         return noise_scheduler
 
-    def encode_images_to_latents(self, args, accelerator, vae, images):
+    def encode_images_to_latents(self, args, vae, images):
         return vae.encode(images)
 
     def shift_scale_latents(self, args, latents):
--- a/train_network.py
+++ b/train_network.py
@@ -1387,7 +1387,7 @@ class NetworkTrainer:
                 # VALIDATION PER STEP
                 should_validate_step = (
                     args.validate_every_n_steps is not None
-                    and args.validation_at_start is not None
+                    and args.validation_at_start
                     and (global_step - 1) % args.validate_every_n_steps == 0 # Note: Should use global step - 1 since the global step is incremented prior to this being run
                 )
                 if accelerator.sync_gradients and should_validate_step:
--- a/train_network.py
+++ b/train_network.py
@@ -1163,6 +1163,18 @@ class NetworkTrainer:
                 args.max_train_steps > initial_step
             ), f"max_train_steps should be greater than initial step / max_train_stepsは初期ステップより大きい必要があります: {args.max_train_steps} vs {initial_step}"
 
+        if args.validate_every_n_steps is not None:
+            validation_steps = (
+                min(args.max_validation_steps, len(val_dataloader)) 
+                if args.max_validation_steps is not None 
+                else len(val_dataloader)
+            )
+            val_progress_bar = tqdm(
+                range(validation_steps), smoothing=0, 
+                disable=not accelerator.is_local_main_process, 
+                desc="validation steps"
+            )
+
         progress_bar = tqdm(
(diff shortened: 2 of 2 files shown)


rockerBOO:
I was going to be working on validation implementation for the training scripts because it requires implementing an abstraction for process_batch like train_network. We can work together for implementing those. Also there may be motivation to make abstractions of the training process because each script is done separately. 

For timesteps consider the work done in https://github.com/kohya-ss/sd-scripts/pull/1165 which added timestep validation process. You can see the progress I made on https://github.com/kohya-ss/sd-scripts/pull/1864/commits/bbf6bbd5ea27231066cec98b8bf2a65f162cb18f that I removed to get the PR merged. Would require a larger abstraction to get the random timestep or use the set timestamps. This timestep process would probably be good as it's on PR and on a smaller subset like LoRA training and then we can then use it in the training scripts. But because the process_batch aspect would need to be translated that may need to come before the training scripts.

Also for merging purposes it might be better to do in smaller pieces so it is easier to review. Each line item you presented could be presented and discussed before you do the work.

rockerBOO:
Also additionally look at https://github.com/kohya-ss/sd-scripts/pull/914 https://github.com/kohya-ss/sd-scripts/pull/1165 for more context of how things got to this point so we can be on the same page there as well.

stepfunction83:
That's fair. I have a lot I want to add to this to get it where I would like. Right now, I think some of the simpler items like the bug fixes, progress bar formatting, and calculating validation at start would be good to get in.

As far as training goes, I had already created a form of abstraction for the Flux finetuning loss calculation, which you see in #1898. I was going to migrate that over to this.

For the stable timesteps, I plan to implement a `state` argument for `calculate_loss()` which will effectively fix the variables for `noise`, `noisy_input_to_model`, `timesteps`, and `sigmas` after they are calculated for the first run. They will not be pre-determined, but will stored in the dataloader and manually set each time the function is called.

rockerBOO:
Yeah I think what you have right now in this PR could be good on it's own and good improvements. Loss calculation probably could be on it's own and Kohya appreciated it. And timesteps another PR would be good. Then we can get them merged faster and I can give feedback and help test.

stepfunction83:
I agree with that approach. I'd like to add one more change to have a fixed number of validation samples (`--validation_set_fixed_size`) as an additional option to this PR. With a large data set like the one I'm using (500 images), it can be a bit annoying to specify small validation sets (<20 items) as a % of the total.

After that we can probably pause there and work on some of the other items in additional PRs.

stepfunction83:
You know what? Looking through the code, that actually seems like a non-trivial workload to implement. It should probably be its own thing as well. I think it's important to be able to set a fixed validation set count because there is a non-trivial performance hit from running a lot of validation steps if the dataset is large.

But for now, you can just calculate the right split and set it accordingly.

stepfunction83:
Okay, take a look at what I implemented here. It's in the same spirit as what I implemented originally and accomplishes the goals of snapshotting the noise state and timesteps for each sample. The work to sample random timesteps is trivial and works _very_ well for creating a stable loss curve.

stepfunction83:
Sure! That sounds good. What I put together here is pretty straightforward, so it shouldn't be too hard to reimplement whenever. I'll look out for the update.

stepfunction83:
> I'm starting to implement it here, any suggestions are welcome: #1903.

I reverted the bulk change that included snapshotting. The remaining commits have some good cleanup items and utility which shouldn't conflict substantially with the changes you're making in the other PR.

Issue #1887, opened 2025-01-20
Insufficient RAM to load Models when using sdxl_gen_img.py in Free Google Colab Environment


DKnight54:
Hi @kohya-ss,

With recent versions of the SD-Scripts, I've been running into issues with loading SDXL models, even with the FP16 flag set, when attempting to generate images with the sdxl_gen_img.py script in the free Google Colab environment because there is not enough RAM available.

Is it possible to implement the -lowram loading option in sdxl_gen_img.py?

PR #1901, opened 2025-01-26
Fix for sdxl_gen_img.py unable to load in Google Colab Env and Distributed Inference Version (accel_sdxl_gen_img.py)

DKnight54:
For the Colab enviroment loading issue, it was pretty straight forward by implementing --full_fp16 and --full_bf16 when loading the model.
#1887 
Basically modifying the [linked lines](https://github.com/DKnight54/sd-scripts/blob/099cea2e24bf53941d2ad66d0c8195dfbf03f2ed/sdxl_gen_img.py#L1495-L1498).

For the distributed version [accel_sdxl_gen_img](https://github.com/DKnight54/sd-scripts/blob/099cea2e24bf53941d2ad66d0c8195dfbf03f2ed/accel_sdxl_gen_img.py).py, there are some major changes.

1. Chances are, interactive mode may be broken, but unable to test without a powerful enough GPU.
2. Instead of processing batches one at a time as it runs through the loop, [the code](https://github.com/DKnight54/sd-scripts/blob/099cea2e24bf53941d2ad66d0c8195dfbf03f2ed/accel_sdxl_gen_img.py) now does the following:
a. It now collects all prompts (prompts x number of images per prompt) in a list.
b. The prompt preparation loop runs only on the main process to prevent messy double printing of prompts, especially when dynamic prompts are used.
c. The list of prompts are then split into a list of batches of prompts.
d. The batches are then distributed onto available GPUs.
3. Simliar to the previous distributed sample generation, image generation order is mixed up.
a. Using similar workaround, `global_step` is used as image generation index.
b. An integer variable [`global_count`](https://github.com/DKnight54/sd-scripts/blob/099cea2e24bf53941d2ad66d0c8195dfbf03f2ed/accel_sdxl_gen_img.py#L1485) is added to the class BatchData
c. It is used during the inference to add an [index to the filenames](https://github.com/DKnight54/sd-scripts/blob/099cea2e24bf53941d2ad66d0c8195dfbf03f2ed/accel_sdxl_gen_img.py#L2415-L2418).
4. Something else that is likely to be broken now is the section that overrides [Deep Shrink and Gradual Latent](https://github.com/DKnight54/sd-scripts/blob/099cea2e24bf53941d2ad66d0c8195dfbf03f2ed/accel_sdxl_gen_img.py#L2736-L2764) as it would only be updated in the main process and I can't think of a way to push it to any sub processes.
a. I have a question about the logic though, because it appears that if the arguements are added to any prompts in a list of prompts, it would override and affect all prompts in that batch and any other batches after that.
b. Is this an intended behaviour? Especially since I don't see a way to unset it.
c. Okay, maybe I do have a way to update all processes. Maybe by using `gather_object()` to get a list of unet or pipes and iterate through them to update. Needs testing to see if it works., but since I don't understand fully Deep Shrink and Gradual Latent, not sure if I can correctly evaluate.

I've also left in some testing and troubleshooting log outputs as it's pretty useful to see what's going on.

I get that the distributed inference is probably not critical, but I have found it useful in jupyter notebooks enviroments with more than one GPU.


DKnight54:
Oooh. Did not realize that `gen_img.py` was intended to be catch all solution for image generation, hence I was using and modifying `sdxl_gen_img.py`.

Interestingly, so far it looks relatively straightforward to implement distributed inference and it should be model agnostic. The only issue is implementing it without breaking the interactive mode.

I've glanced at the code for `gen_img.py` and since it is mostly similar to `sdxl_gen_img.py`, I should be able to port over the modifications I've made to create `accel_sdxl_gen_img.py`, and I should be able to test the interactive mode using SD1.5 (but not on muliple GPUs).

One possible solution is to have the script run only on the main process in interactive mode, but let me figure out how interactive mode works first before deciding.

However, there is one question I hope you can answer, what's the logic for this following [code snippit](https://github.com/kohya-ss/sd-scripts/blob/6e3c1d0b58f03522f294dc2b0acbbbecc944d018/gen_img.py#L2919-L2921):
https://github.com/kohya-ss/sd-scripts/blob/6e3c1d0b58f03522f294dc2b0acbbbecc944d018/gen_img.py#L2919-L2921

So far it looks like it's supposed to start processing the current batch even if there is less than a full batch and there is a change in the BatchDataExt parameters, eg width and height and steps, then run the previous batch?

Interestingly, in my testing so far, I'm pretty sure I've accidentally run 1152x896 and 896x1152 in a batch together with no obvious negative effect, altho I see now that different steps should definitely have different batches.🤯

If it's okay with you, I can look into how to implement distributed inference on gen_img.py

PR #1903, opened 2025-01-27
Val loss improvement

kohya-ss:
- [x] train/eval state for the network and the optimizer.
- [x] stable timesteps
- [x] stable noise
- [x] support block swap

### DIFF
--- a/library/model_util.py
+++ b/library/model_util.py
@@ -643,16 +643,15 @@ def convert_ldm_clip_checkpoint_v2(checkpoint, max_length):
             new_sd[key_pfx + "k_proj" + key_suffix] = values[1]
             new_sd[key_pfx + "v_proj" + key_suffix] = values[2]
 
-    # rename or add position_ids
+    # remove position_ids for newer transformer, which causes error :(
     ANOTHER_POSITION_IDS_KEY = "text_model.encoder.text_model.embeddings.position_ids"
     if ANOTHER_POSITION_IDS_KEY in new_sd:
         # waifu diffusion v1.4
-        position_ids = new_sd[ANOTHER_POSITION_IDS_KEY]
         del new_sd[ANOTHER_POSITION_IDS_KEY]
-    else:
-        position_ids = torch.Tensor([list(range(max_length))]).to(torch.int64)
 
-    new_sd["text_model.embeddings.position_ids"] = position_ids
+    if "text_model.embeddings.position_ids" in new_sd:
+        del new_sd["text_model.embeddings.position_ids"]
+
     return new_sd
 
 
--- a/library/sdxl_train_util.py
+++ b/library/sdxl_train_util.py
@@ -344,8 +344,6 @@ def add_sdxl_training_arguments(parser: argparse.ArgumentParser, support_text_en
 
 def verify_sdxl_training_args(args: argparse.Namespace, supportTextEncoderCaching: bool = True):
     assert not args.v2, "v2 cannot be enabled in SDXL training / SDXL学習ではv2を有効にすることはできません"
-    if args.v_parameterization:
-        logger.warning("v_parameterization will be unexpected / SDXL学習ではv_parameterizationは想定外の動作になります")
 
     if args.clip_skip is not None:
         logger.warning("clip_skip will be unexpected / SDXL学習ではclip_skipは動作しません")
--- a/train_network.py
+++ b/train_network.py
@@ -1207,10 +1207,6 @@ class NetworkTrainer:
                 args.max_train_steps > initial_step
             ), f"max_train_steps should be greater than initial step / max_train_stepsは初期ステップより大きい必要があります: {args.max_train_steps} vs {initial_step}"
 
-        progress_bar = tqdm(
-            range(args.max_train_steps - initial_step), smoothing=0, disable=not accelerator.is_local_main_process, desc="steps"
-        )
-
         epoch_to_start = 0
         if initial_step > 0:
             if args.skip_until_initial_step:
@@ -1308,6 +1304,10 @@ class NetworkTrainer:
             logger.info(f"text_encoder [{i}] dtype: {param_3rd.dtype}, device: {t_enc.device}")
 
         clean_memory_on_device(accelerator.device)
+
+        progress_bar = tqdm(
+            range(args.max_train_steps - initial_step), smoothing=0, disable=not accelerator.is_local_main_process, desc="steps"
+        )
 
         validation_steps = (
             min(args.max_validation_steps, len(val_dataloader)) if args.max_validation_steps is not None else len(val_dataloader)


stepfunction83:
I love the approach to holding the rng_state and setting the validation state using the seed. It's much more elegant than tracking the state separately.

stepfunction83:
I would also add that once this is put in place, there won't be a need for a moving average to track the validation loss. Using consistent timesteps and noise will make it almost entirely stable, so displaying the mean of the validation loss amounts for each validation run should be all that's needed.

Since the validation set is subject to change if the core dataset changes, I've found tracking the validation loss relative to the initial loss is also helpful to make progress across different training runs comparable.

rockerBOO:
This looks great!

What are you using for formatting the code? I've been manually formatting but might be easier to align the formatting if I use the same formatting tool.

gesen2egee:
It seems that using 1/SNR correction for timestep sampling works better (I previously used debiased 1/√SNR, which is similar in meaning).
Perhaps averaging wouldn’t be necessary in this case.

Additionally, I have some thoughts on the args.
For validation_split, how about making it an integer greater than 1 to automatically represent the number of validation samples?
This would be more convenient.

gesen2egee:
image
https://github.com/[spacepxl/demystifying-sd-finetuning](https://github.com/spacepxl/demystifying-sd-finetuning)
Here's a suggestion for a function that, while not perfect, can normalize the losses across different timesteps to the same magnitude. I believe this approach is more reliable

spacepxl:
@gesen2egee you would need a different fit equation for each new model, and it's not really relevant when you make validation fully deterministic. I've tried applying it to training loss and it was extremely harmful.

rockerBOO:
>   File "/mnt/900/builds/sd-scripts/library/train_util.py", line 5968, in get_timesteps
    timesteps = torch.randint(min_timestep, max_timestep, (b_size,), device="cpu")
RuntimeError: random_ expects 'from' to be less than 'to', but got from=200 >= to=200

In get_timesteps maybe

```python
if min_timestep < max_timestep:
    timesteps = torch.randint(min_timestep, max_timestep, (b_size,), device="cpu")
else:
    timesteps = torch.ones(b_size, device="cpu") * min_timestep
```

I know this isn't completed but I tried it anyways.

rockerBOO:
Screenshot 2025-01-29 at 01-08-39 true-bee-59 cyberpunk-boo-kohya-lora – Weights   Biases

Validation loss (dashed) looking pretty smooth from a test.

stepfunction83:
I just realized that if there are any repeats in the dataset, validation error will not be calculated properly as the images used for validation can still exist in the training data.

rockerBOO:
Repeats are done after the split. I also made sure regularization images do not go into the validation dataset when splitting.

stepfunction83:
Nice! That's good to hear!

stepfunction83:
I am not able to get a stable loos from this PR:

image

In this, I calculate the average loss on each validation run and graph them out. As you can see, the line is quite volatile, and has been on all of the attempts I've tried so far. 

It could be user error. Here are the arguments I'm sending:

--validation_seed 43 --validation_split .02 --validate_every_n_steps 50

To calculate the error per cycle:
`
val_losses = [] # Before training loop
...
current_val_losses = [] # Before validation loop
...
current_loss = loss.detach().item()
current_val_losses.append(current_loss)
...
val_losses.append(sum(current_val_losses)/len(current_val_losses)) #After validation loop

 if is_tracking:
                        logs = {
                            "loss/validation/val_loss": val_losses[-1]
                        }
`

stepfunction83:
> In SDXL training, the random numbers are generated using the device, so this may be the cause.
> 
> https://github.com/kohya-ss/sd-scripts/blob/45ec02b2a8b5eb5af8f5b4877381dc4dcc596cb9/library/train_util.py#L5945
> 
> Also, torch.manual_seed initializes the random seed for all devices, but rng_state seems to only work for CPU.
> 
> https://pytorch.org/docs/stable/generated/torch.manual_seed.html#torch.manual_seed https://pytorch.org/docs/stable/generated/torch.get_rng_state.html#torch.get_rng_state
> 
> Although it would be a breaking change, it would be better to unify the way random numbers are generated.

I probably should have specified, but I am doing Flux LoRAs in this case, not SDXL. Also, if manual_seed is setting it for all devices, then it should be providing the desired consistency, so it's odd that it's not. I'll do some digging to confirm whether what I'm experiencing is even due to the noise, or whether it's something else.

There is always the possibility of caching the noise and using it in repeat runs, which doesn't interfere with any random state. Based on my initial testing, this didn't seem to have a significant overhead, but isn't quite as pretty as just setting the seed.

stepfunction83:
After some digging, the noise is absolutely consistent over time, but the noisy_model_input is the culprit here. I'm trying to understand why it is changing from one iteration to the next.

stepfunction83:
image

That's the culprit

stepfunction83:
I was able to resolve this by setting random.seed() in addition to setting torch.manual_seed():

```
rng_state = torch.get_rng_state()
                    random_seed_state = random.randint(0,100000)
                    torch.manual_seed(args.validation_seed if args.validation_seed is not None else args.seed)
                    random.seed(args.validation_seed if args.validation_seed is not None else args.seed)
```

and then

```
torch.set_rng_state(rng_state)
                    random.seed(random_seed_state)
```

It's a bit hacky, but storing the state for random as `random_seed_state = random.randint(0,100000)` does get the job done.

rockerBOO:
For the main progress_bar if you do `progress_bar.unpause()` after validation steps it will reset the timer for main training iterations. 

https://tqdm.github.io/docs/tqdm/#unpause

Otherwise has been working for me and block swap is working as well.

stepfunction83:
Not at my computer today, but were the fixes from #1900 implemented to fix the issue with running validation when latents are not cached and to fix bar positioning?

rockerBOO:
What is the expectation for tensorboard logs? I was mentioning that we may need a different approach for different "experiment trackers". Like tensorboard may be handled differently than wandb, and not have everything through the generic `accelerate.log`. We do this already for wandb specific features like sample images. Though I want to consider what seems appropriate for the other trackers they support. 

For Validation, the logging of the individual validation steps isn't being represented. This is due to them all happening on the same "step", which we are using global step. Iterating through step for each accelerate.log (and not setting steps) would work better to record them all. Meaning `accelerate.log(logs)` but do not pass the steps. Downside is the graphs will have gaps in their representation. 

Example from another trainer. Notice the flat parts. 
Screenshot 2025-02-12 at 12-32-19 noble-jazz-84 t5-distill – Weights   Biases
Screenshot 2025-02-12 at 12-32-06 noble-jazz-84 t5-distill – Weights   Biases

I do not know what is best here in the end. I think compromise and making things work for the trackers how we want would be the best. Make tensorboard work how you'd expect (via the specific experiment tracker) and then we can make it work for the other trackers maybe more generically.

stepfunction83:
To add on, I'd also like to recommend the metric of relative validation loss compared to the initial loss calculated.

Since the validation set is not constant, it is important to calculate the loss relative to where it started to more accurately track progress on different training runs.

`rel_val_loss=curr_val_loss / val_losses[0]`

rockerBOO:
Only issue with your approach for using the step in that way is it blocks those logs from being recorded on Wandb at all because they are not incrementing or they increment past the end preventing other logs using accelerate.log(*, step=global_step) from being logged as well. It's pretty annoying that Wandb works in this way but would probably require a pass to get them both working like how you have it presented. Otherwise if you use validation, normally logging would fail after the validation pass because global_step + val step would move the step increment past global_step.

stepfunction83:
I would agree with that. Logging the individual validation steps is almost entirely a waste of time. The only metric of interest is the average validation loss.

rockerBOO:
The secondary issue was setting the epoch, since the step would be lower than the global step it would not record the epochs as well. 

I think they generally want you to not set the steps at all, and set the "global_step" or "epoch" in as actual values. Then you can set them as "metric" and then you can use them in the graphs to organize them based on global_step or epoch. 

`accelerator.log({"global_step": global_step, "epoch": epoch + 1})`

So for tensorboard would need to be like:

```python
    if "tensorboard" in [tracker.name for tracker in accelerator.trackers]:
        tensorboard_tracker = accelerator.get_tracker("tensorboard")
        with tensorboard_tracker.as_default(step=global_step + val_step):
            tf.summary.scalar("loss/validation/step_current": current_loss)
```


```python
    if "wandb" in [tracker.name for tracker in accelerator.trackers]:
        wandb_tracker = accelerator.get_tracker("wandb")
        wandb_tracker.log(
            {
               "loss/validation/step_current": current_loss,
               "global_step": global_step,
               "val_step": val_step
            }
        )
```

Which then we would add `global_step` to the metric for wandb
```python
        if "wandb" in [tracker.name for tracker in accelerator.trackers]:
            import wandb 
            wandb_tracker = accelerator.get_tracker("wandb", unwrap=True)

            # Define specific metrics to handle validation and epochs "steps"
            wandb_tracker.define_metric("epoch", hidden=True)
            wandb_tracker.define_metric("val_step", hidden=True)

            wandb_tracker.define_metric("global_step", hidden=True)
```

I think we could make this happen for both but would need to drop `accelerator.log()` to be able to do it. Make a new function that handles these distinctions so we can utilize the trackers to their most efficient methods and not work around the lack of functionality of `accelerator.log`.

rockerBOO:
Looks good to me! I'll try to test it in a little bit. Thanks for doing this!

heinrichI:
Val loss not compatible with masked_loss?

rockerBOO:
Screenshot 2025-02-24 at 14-08-18 easy-mountain-428 women-kohya-lora – Weights   Biases

It is working for me with the 3 loss types, and then I can set the global_step and works. Only 4 epochs and every 32 steps with only 80 steps in the test. Looks good to me to merge.

PR #1906, opened 2025-01-28
Add uv/pyproject and black config

rockerBOO:
https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-format
https://docs.astral.sh/uv/

Feel free to close if not desired. Tried to match what you have for your tuner. uv.lock added but can make a PR without it. Maybe it will cause issues with people using as a library but maybe it can work a little better as a library this way too.


rockerBOO:
Seems the tests are working :) I'll need to consider the approach.

PR #1909, opened 2025-01-29
Move progress bar to account for sampling image first

rockerBOO:
Progress bar was including sampling images (--sample_at_first) and causing some amount of the time to be off what would be expected.

### DIFF
--- a/train_network.py
+++ b/train_network.py
@@ -1163,10 +1163,6 @@ class NetworkTrainer:
                 args.max_train_steps > initial_step
             ), f"max_train_steps should be greater than initial step / max_train_stepsは初期ステップより大きい必要があります: {args.max_train_steps} vs {initial_step}"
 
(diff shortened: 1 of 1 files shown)


stepfunction83:
I know that `tqdm()` is called multiple times for the progress bar to create it instead of calling `progress_bar.reset()`. Have you tried just resetting it instead of recreating it and seeing how that impacts its placement?

rockerBOO:
Could have it reset before it goes into training where I moved it. But I think it makes sense to put it where I have it right before the training. The progress is for the training steps so I think it makes sense to move it to capture this specifically. 

The other things you improved on validation progress may allow for different results and reset may be more appropriate there.

stepfunction83:
Oh, I see, you just pushed it down in the code. That makes perfect sense to me.

PR #1910, opened 2025-01-29
Optimize Latent Caching Speed with VAE Optimizations

alefh123:
PR Summary:

With some help from DeepSeek, this PR aims to improve latent caching speed.

Key Changes:

Mixed Precision Caching: VAE encoding now uses FP16 (or BF16) during latent caching for faster computation and reduced memory use.

Channels-Last VAE: VAE is temporarily switched to channels_last memory format during caching to improve GPU performance.

--vae_batch_size Utilization: This leverages the existing --vae_batch_size option; users should increase it for further speedups.

Benefits:

Significantly Faster Latent Caching: Reduces preprocessing time.

Improved GPU Efficiency: Optimizes VAE encoding on GPUs.

Impact: Faster training setup due to quicker latent caching.

This is much more concise and directly highlights the essential changes and their impact. Let me know if you would like it even shorter or with any other adjustments!

Based on the optimizations implemented—mixed precision and channels-last format for the VAE during caching—**a speedup of 2x to 4x** is a reasonable estimate.

### DIFF
--- a/sdxl_train.py
+++ b/sdxl_train.py
@@ -187,10 +187,6 @@ def train(args):
         )
         return
 
-    if cache_latents:
-        assert (
-            train_dataset_group.is_latent_cacheable()
-        ), "when caching latents, either color_aug or random_crop cannot be used / latentをキャッシュするときはcolor_augとrandom_cropは使えません"
 
     if args.cache_text_encoder_outputs:
         assert (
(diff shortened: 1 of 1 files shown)


rockerBOO:
I don't think this would be where it is doing the caching though. Converting SDXL VAE to fp16 also can cause NaN issues so we probably wouldn't want to convert it to fp16 regardless. 

The memory format might be what we can try from this as it might improve the performance.

feffy380:
The bigger issue is that VAE caching loads batches synchronously, so it blocks on *every single batch*.
Just move the image loading and preprocessing to a dataloader for a massive speedup: https://github.com/feffy380/sd-scripts/commit/c33b31c65f83c5cb187bf9dd5dca955c46051980

feffy380:
@rockerBOO No, asynchronous means loading the images in a separate thread. Right now the batch size doesn't matter because it has to wait while each image in the batch is loaded, one after the other, then encodes the batch, and only then can it start to load the next batch of images. Caching currently wastes most of its time blocked by IO because it's all being done in the same thread. A dataloader solves this.

PR #1913, opened 2025-01-30
Support loading diffusers models 

rockerBOO:
- Allow models to be loaded that are just the transformer https://huggingface.co/rockerBOO/Flux-Dev2Pro-fp8_e4m3fn
- Allow reduced models to be loaded (they may not have 1-3 shards). There is a index weight_map file that showcases what modules are mapped. We can use this to get a list of shards.  

Proposing this but needs further testing. For me it allows me to load a fp8 version of a model which has less shards but is in diffusers format.

### DIFF
--- a/library/flux_utils.py
+++ b/library/flux_utils.py
@@ -25,6 +25,21 @@ MODEL_NAME_DEV = "dev"
 MODEL_NAME_SCHNELL = "schnell"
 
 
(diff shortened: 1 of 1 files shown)


rockerBOO:
Closing in favor of #2007 which handles the different cases more appropriately.

PR #1940, opened 2025-02-17
Fix sizes for validation split

### LINKED ISSUE #1934: BUG: Validation breaks cached image sizes
The validation split helper function does not adjust the image `sizes` list, so images get stretched due to being assigned the wrong size.

https://github.com/kohya-ss/sd-scripts/blob/0778dd9b1df0d6aa33287ded3ce4195f3d03251b/library/train_util.py#L1939

https://github.com/kohya-ss/sd-scripts/blob/0778dd9b1df0d6aa33287ded3ce4195f3d03251b/library/train_util.py#L1976-L1998

@rockerBOO

rockerBOO:
Fixes #1934

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -148,10 +148,11 @@ TEXT_ENCODER_OUTPUTS_CACHE_SUFFIX_SD3 = "_sd3_te.npz"
 
 def split_train_val(
     paths: List[str], 
+    sizes: List[Optional[Tuple[int, int]]],
     is_training_dataset: bool, 
     validation_split: float, 
     validation_seed: int | None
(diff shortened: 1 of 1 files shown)


feffy380:
This doesn't fix the issue. You still shuffle `paths` without shuffling `sizes`, so each path is paired with the wrong size.

You need to do something like:

```python
# pair path with size
paired_paths = list(zip(img_paths, sizes))

# shuffle together
random.shuffle(paired_paths)

# create split from paired_paths...

# split paths and sizes into separate lists again
img_paths, sizes = list(zip(*paired_paths))
```

PR #1951, opened 2025-02-24
Fix sd 2.0/2.1 lora fine-tuning text encoder params mismatch issue

miracleyoo:
Fix sd 2.0/2.1 lora fine-tuning text encoder params mismatch issue. With the original text encoder and its corresponding parser, stable diffusion 2.0/2.1/2.1-unclip all cannot be properly loaded due to the text encoder transformer version difference and checkpoints' state_dict key name difference. After this fix, all these three versions are tested to work well when executing lora fine tuning.

Passed the Lora fine-tuning trial using the following SD base models:
1. https://huggingface.co/stabilityai/stable-diffusion-2-1/tree/main
2. https://huggingface.co/stabilityai/stable-diffusion-2-1-unclip/tree/main
3. https://huggingface.co/stabilityai/stable-diffusion-2/tree/main

### DIFF
--- a/library/model_util.py
+++ b/library/model_util.py
@@ -10,7 +10,7 @@ init_ipex()
 
 import diffusers
 from transformers import CLIPTextModel, CLIPTokenizer, CLIPTextConfig, logging
-from diffusers import AutoencoderKL, DDIMScheduler, StableDiffusionPipeline  # , UNet2DConditionModel
+from diffusers import AutoencoderKL, DDIMScheduler, StableDiffusionPipeline, StableUnCLIPImg2ImgPipeline  # , UNet2DConditionModel
 from safetensors.torch import load_file, save_file
 from library.original_unet import UNet2DConditionModel
 from library.utils import setup_logging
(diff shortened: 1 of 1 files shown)


miracleyoo:
Thanks for the great work and swift reply.

Actually there is another major difference other than the `embedder.model`, which is the addition of the entry `text_model.embeddings.position_ids`. SD 2.0/2.1/2.1-unclip doesn't seem to have this entry in their `state_dict`, that's why I removed that entry. 
I'm not sure why this key is added in the original `convert_ldm_clip_checkpoint_v2` function, and whether that's necessary for some other version of base model parsing. That's why I write a separate function to avoid potential conflict. 

Also, I noticed that the `transformers_version` is actually "4.28.0.dev0" for SD 2.0/2.1/2.1-unclip models, instead of the "4.25.0.dev0" in the original code. This also caused some issue. Basically that's all what I've corrected.

Please feel free to do any necessary modification on top of this PR, like merge the `convert_ldm_clip_checkpoint_v2` and `convert_ldm_clip_checkpoint_v2_fix`, if you can tell apart in which case `text_model.embeddings.position_ids` is needed and when it's not by checking some conditions. 

Plesae let me know if there's anything I can help or clarify.

Best Regard

PR #1960, opened 2025-02-28
Sd3 safetensors merge

kohya-ss:
Currently all models are loaded into the main RAM before saving, so it uses the total size of .safetensor files main RAM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,10 @@ The command to install PyTorch is as follows:
 
 ### Recent Updates
 
+Mar 6, 2025:
+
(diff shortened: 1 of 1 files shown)


FurkanGozukara:
> Fixed a bug that caused tensors to be converted to float32, and now saves as it is. Also, add `--save_precision` option, convert tensors to the specified type.
> 
> @FurkanGozukara Thanks for reporting it. I will merge this PR if you confirm the fix.

Just tested and works great

FP16 half size :) please merge as soon as possible

FurkanGozukara:
@kohya-ss can you merge this today so i can start new trainings with this feature and auto save with clip l and clip g while training?

FurkanGozukara:
> There is no automatic merging, so please run the script separately.

can you please implement automatic merge?

running this script is on each file extremely tiresome and impossible for many people

FurkanGozukara:
@kohya-ss which file i would need to modify to achieve this?

which file saves checkpoint and clip l and clip g during training so i can modify it to generate merged output and delete other outputs

Issue #1980, opened 2025-03-11
Flux: big jumps in key size seems to be due to min_snr_gamma not being hooked up


araleza:
While measuring sd-scripts' LoRA key lengths with Tensorboard, I noticed that from time to time there were big jumps.

The jumps seem to correspond to the noise timestep value for that step being >900. In the example below, the timestep value is 932. (I'm not talking about the training step number, which coincidentally happens to also be around 900 here). While looking into reducing the loss for high timestep values, I spotted that that discussion seems to have already taken place for Stable Diffusion and SDXL, with a pull request by @AI-Casanova for a feature called `--min_snr_gamma` here in March 2023:

https://github.com/kohya-ss/sd-scripts/pull/308#issuecomment-1524512544

(The implementation was later fixed by @drhead)

Flux does not seem to have this min-snr-gamma feature enabled, as the function that should call it is stubbed out (in `flux_train_network.py`):

```
    def post_process_loss(self, loss, args, timesteps, noise_scheduler):
        return loss
```

It would be nice to get an implementation of this working for Flux to stop the overly-large jumps in key sizes occuring when the timestep value is high. Flux does actually take the `--min_snr_gamma` parameter without complaint, but silently makes no use of the value that is set.

![Image](https://github.com/user-attachments/assets/6a35816f-563b-4944-9f04-21022c17394d)

araleza:
@drhead, are you maybe talking about `--timestep_sampling flux_shift`? I found out about that just a few weeks ago, and I have it in place on my command line. But I still see the big loss values for the high time steps. Unless these big steps are intentional and expected for Flux?

drhead:
I have not trained anything on Flux, but larger loss values for some timesteps sounds normal for Flux's timestep sampling.

Min-snr-gamma is a timestep weighting strategy.  It simply is a multiplier on loss values based on the timestep's noise level, and it proportionally decreases the impact of the gradients coming from that image (it reduces how much influence those have on the model).  It does so with the objective of making the influence of each timestep more equal. As a side note, you shouldn't take the lowered loss you get from min-snr-gamma too seriously.  You could simply multiply the loss values for all timesteps by 0.01, and that won't mean that your model is any better than before.

[...]

PR #1991, opened 2025-03-18
Support alpha cumulative product using shifted sigmas for Flux

rockerBOO:
> In many diffusion models, xsigma² = (1-α)/α where α is the cumulative product of alphas. So we can derive alphas_cumprod from sigmas

The sigmas are shifted in initialization of the noise_scheduler so we can use the shifted sigmas, unless the scheduler is using a dynamic shifting (based on resolution). I am not detecting if the scheduler is using dynamic shifting but we can add this clarification. 

Noise scheduler's from Diffusers have `index_to_timestep` which we can use to better align with future timestep alignment.

Supporting timesteps in ranges of 0 to 1 and 1 to 1000. 

Support in Flux network training:

- min SNR gamma `--min_snr_gamma 5.0`
- debiased estimation `--debiased_estimation`

Flux huber loss using `snr`:

- huber c SNR `--huber_schedule snr`

v-pred in this PR currently but not sure if it is appropriate for flux.

---

Testing of the parameters and making sure regression for other models due to changing the timestep process to account for finding alpha_cumprod where needed. We might want to preload alpha_cumprod to reduce the complexity where appropriate (maybe not appropriate if using dynamic shifting). 

Related #1980

### DIFF
--- a/flux_train_network.py
+++ b/flux_train_network.py
@@ -21,6 +21,13 @@ from library import (
     strategy_flux,
     train_util,
 )
+from library.custom_train_functions import (
+    prepare_scheduler_for_custom_training,
+    apply_snr_weight,
+    scale_v_prediction_loss_like_noise_prediction,
+    add_v_prediction_like_loss,
+    apply_debiased_estimation,
+)
 from library.utils import setup_logging
 
 setup_logging()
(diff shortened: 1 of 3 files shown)


rockerBOO:
- Updated loss modification functions to take image_size for dynamically shifting timesteps (timestep_sampling == "flux_shift"). 
- Added tests for loss modification in custom train functions

PR #1992, opened 2025-03-18
Add IP noise gamma for Flux

rockerBOO:
IP Noise gamma helps with exposure bias and is applicable for any model with noise. 

https://arxiv.org/pdf/2301.11706

- Add support for Flux models

### DIFF
--- a/library/flux_train_utils.py
+++ b/library/flux_train_utils.py
@@ -415,6 +415,16 @@ def get_noisy_model_input_and_timesteps(
     bsz, _, h, w = latents.shape
     sigmas = None
 
+    ip_noise_gamma = 0.0
+
+    # Add noise to the latents according to the noise magnitude at each timestep
+    # (this is the forward diffusion process)
+    if args.ip_noise_gamma:
+        if args.ip_noise_gamma_random_strength:
+            ip_noise_gamma = torch.rand(1, device=latents.device) * args.ip_noise_gamma
+        else:
+            ip_noise_gamma = args.ip_noise_gamma
+
     if args.timestep_sampling == "uniform" or args.timestep_sampling == "sigmoid":
         # Simple random t-based noise sampling
         if args.timestep_sampling == "sigmoid":
@@ -425,7 +435,7 @@ def get_noisy_model_input_and_timesteps(
 
         timesteps = t * 1000.0
         t = t.view(-1, 1, 1, 1)
-        noisy_model_input = (1 - t) * latents + t * noise
+        noisy_model_input = (1 - t) * latents + t * noise + ip_noise_gamma
     elif args.timestep_sampling == "shift":
         shift = args.discrete_flow_shift
         logits_norm = torch.randn(bsz, device=device)
@@ -435,7 +445,7 @@ def get_noisy_model_input_and_timesteps(
 
         t = timesteps.view(-1, 1, 1, 1)
         timesteps = timesteps * 1000.0
-        noisy_model_input = (1 - t) * latents + t * noise
+        noisy_model_input = (1 - t) * latents + t * noise + ip_noise_gamma
     elif args.timestep_sampling == "flux_shift":
         logits_norm = torch.randn(bsz, device=device)
         logits_norm = logits_norm * args.sigmoid_scale  # larger scale for more uniform sampling
@@ -445,7 +455,7 @@ def get_noisy_model_input_and_timesteps(
 
         t = timesteps.view(-1, 1, 1, 1)
         timesteps = timesteps * 1000.0
-        noisy_model_input = (1 - t) * latents + t * noise
+        noisy_model_input = (1 - t) * latents + t * noise + ip_noise_gamma
     else:
         # Sample a random timestep for each image
         # for weighting schemes where we sample timesteps non-uniformly
@@ -461,7 +471,8 @@ def get_noisy_model_input_and_timesteps(
 
         # Add noise according to flow matching.
         sigmas = get_sigmas(noise_scheduler, timesteps, device, n_dim=latents.ndim, dtype=dtype)
-        noisy_model_input = sigmas * noise + (1.0 - sigmas) * latents
+        noisy_model_input = sigmas * noise + (1.0 - sigmas) * latents + ip_noise_gamma
+
 
     return noisy_model_input.to(dtype), timesteps.to(dtype), sigmas
 
--- a/library/flux_train_utils.py
+++ b/library/flux_train_utils.py
@@ -471,7 +471,7 @@ def get_noisy_model_input_and_timesteps(
 
         # Add noise according to flow matching.
         sigmas = get_sigmas(noise_scheduler, timesteps, device, n_dim=latents.ndim, dtype=dtype)
-        noisy_model_input = sigmas * noise + (1.0 - sigmas) * latents + ip_noise_gamma
+        noisy_model_input = sigmas * noise +  ip_noise_gamma + (1.0 - sigmas) * latents
 
 
     return noisy_model_input.to(dtype), timesteps.to(dtype), sigmas


MegaCocos:
> IP Noise gamma helps with exposure bias and is applicable for any model with noise.
> 
> https://arxiv.org/pdf/2301.11706
> 
>     * Add support for Flux models

If i understand well i need to add `--ip_noise_gamma` to the command line to implement this? If i want to try it before it is merged

library/flux_train_utils.py
```diff
@@ -425,7 +435,7 @@ def get_noisy_model_input_and_timesteps(
 
         timesteps = t * 1000.0
         t = t.view(-1, 1, 1, 1)
-        noisy_model_input = (1 - t) * latents + t * noise
+        noisy_model_input = (1 - t) * latents + t * noise + ip_noise_gamma
```
recris, line 438: this looks like we're adding a constant offset to the noise, not actually random gaussian noise scaled by the constant.

shouldn't this be like
```
noisy_model_input = (1 - t) * latents + t * noise + ip_noise_gamma * torch.randn_like(noise)
```

like on `train_util.get_noise_noisy_latents_and_timesteps`

rockerBOO: Whoops, thanks for looking. I made it so it always calculates IP noise but ideally only when it is needed. Probably some way to have a no-op value instead or add a conditional.

library/flux_train_utils.py
```diff
@@ -415,6 +415,16 @@ def get_noisy_model_input_and_timesteps(
...
 
+    # Add noise to the latents according to the noise magnitude at each timestep
+    # (this is the forward diffusion process)
+    if args.ip_noise_gamma:
+        if args.ip_noise_gamma_random_strength:
+            ip_noise = (torch.rand(1, device=latents.device) * args.ip_noise_gamma) * torch.randn_like(latents)
+        else:
+            ip_noise = args.ip_noise_gamma * torch.randn_like(latents)
+    else:
+        ip_noise = torch.ones_like(latents)
```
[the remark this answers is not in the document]
rockerBOO: I'll check this value actually, I thought zeroes like but wouldn't that zero out the noise? I'll double check.

rockerBOO:
> If i understand well i need to add `--ip_noise_gamma` to the command line to implement this? I mean if i want to try it before it is merged. Just trying to apply now all your PR's to my actual installation as i want to start new training now.

```
  --ip_noise_gamma IP_NOISE_GAMMA
                        enable input perturbation noise. used for
                        regularization. recommended value: around 0.1 (from
                        arxiv.org/abs/2301.11706) / input perturbation
                        noiseを有効にする。正則化に使用される。推奨値: 0.1程度
                        (arxiv.org/abs/2301.11706 より)
  --ip_noise_gamma_random_strength
                        Use random strength between 0~ip_noise_gamma for input
                        perturbation noise./ input perturbation
                        noiseにおいて、0からip_noise_gammaの間でランダムな強度を使用します。
```

I probably need to fix some things with the approach here but thats the docs how to use it.

library/flux_train_utils.py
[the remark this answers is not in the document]
recris: it would be zeroing the noise only if multiplying... you're doing `(noise + ip_noise)`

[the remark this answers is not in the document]
rockerBOO: Doh, good point. Sorry I'm a little distracted but I was definitely doing multiplying in my head.

rockerBOO:
@MegaCocos https://github.com/rockerBOO/sd-scripts I have a version over here that has all my PRs in it if you want to try all them.

MegaCocos:
> @MegaCocos https://github.com/rockerBOO/sd-scripts I have a version over here that has all my PRs in it if you want to try all them.

Thank you, just ended up to recollect all the PR's files now:) If it fails then i just get the complete branch, surely. While i have some files edited for INFO during the training don't want to edit them again. Thank you a lot for your work and effort, i activated the notifications for all your posts so i will stay informed. I put `  --ip_noise_gamma 0.1`, **seems to be the best choice, right?** And `--min_snr_gamma 5.0 --debiased_estimation --huber_schedule snr --network_args "train_double_block_indices=0,1,17,18" "train_single_block_indices=all" "ggpo_sigma=0.03" "ggpo_beta=0.01"`

recris:
Hard to know if the proposed default (0.1) will be any good in Flux.1 and other flow match based models.

MegaCocos:
> Hard to know if the proposed default (0.1) will be any good in Flux.1 and other flow match based models.

Would you recommend any other value to try?

rockerBOO:
See the paper for their assessment of other models. They found 0.1 worked for them based on experiments but like recris mentioned it has not been tested on flow matching models.

rockerBOO:
@MegaCocos also min_snr_gamma and debaised_estimation are mutually exclusive they work in the same space so wouldn't do both.

MegaCocos:
> @MegaCocos also min_snr_gamma and debaised_estimation are mutually exclusive they work in the same space so wouldn't do both.

You said: "I implemented debiased estimation, huber snr loss, and ip noise gamma for flux which seems to perform really well" So which one gave you the better result?

rockerBOO:
that works, would need to set the loss type to huber too `--loss_type huber`

MegaCocos:
> that works, would need to set the loss type to huber too `--loss_type huber`

So: `--debiased_estimation --huber_schedule snr --loss_type huber --network_args "train_double_block_indices=0,1,17,18" "train_single_block_indices=all" "ggpo_sigma=0.03" "ggpo_beta=0.01"` Is it OK?

rockerBOO:
Reverted the noise after reviewing the paper. The noise and the perturbation of the noise should be separate. Added it at the end if perturbation is needed.

library/flux_train_utils.py
```diff
@@ -463,6 +463,16 @@ def get_noisy_model_input_and_timesteps(
...
 
+    # Add noise to the latents according to the noise magnitude at each timestep
+    # (this is the forward diffusion process)
+    if args.ip_noise_gamma:
+        xi = torch.randn_like(latents, device=latents.device, dtype=dtype)
+        if args.ip_noise_gamma_random_strength:
+            ip_noise_gamma = (torch.rand(1, device=latents.device, dtype=dtype) * args.ip_noise_gamma)
+        else:
+            ip_noise_gamma = args.ip_noise_gamma
+        noisy_model_input += ip_noise_gamma * xi
```
recris, line 474: In the paper, the perturbed version is stated as
$`y_t = \sqrt{\bar{\alpha}_t}x_0 + \sqrt{1 - \bar{\alpha}_t}(\epsilon + \gamma_t\xi)`$

Here we adding $`\gamma_t\xi`$ but not scaling by $` \sqrt{1 - \bar{\alpha}_t}`$ (in Flux code it would be just `t` or `sigmas`, I think)

recris: I've tested this latest change and I'm getting a significant quality degradation in image details.

recris: Refactoring suggestion:
```
def get_noisy_model_input_and_timesteps(
    args, noise_scheduler, latents: torch.Tensor, noise: torch.Tensor, device, dtype
) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
    bsz, _, h, w = latents.shape
    sigmas = None

    if args.timestep_sampling == "uniform" or args.timestep_sampling == "sigmoid":
        # Simple random t-based noise sampling
        if args.timestep_sampling == "sigmoid":
            # https://github.com/XLabs-AI/x-flux/tree/main
            t = torch.sigmoid(args.sigmoid_scale * torch.randn((bsz,), device=device))
        else:
            t = torch.rand((bsz,), device=device)

        timesteps = t * 1000.0
        s = t.view(-1, 1, 1, 1)
    elif args.timestep_sampling == "shift":
        shift = args.discrete_flow_shift
        logits_norm = torch.randn(bsz, device=device)
        logits_norm = logits_norm * args.sigmoid_scale  # larger scale for more uniform sampling
        timesteps = logits_norm.sigmoid()
        timesteps = (timesteps * shift) / (1 + (shift - 1) * timesteps)
        s = timesteps.view(-1, 1, 1, 1)
        timesteps = timesteps * 1000.0
    elif args.timestep_sampling == "flux_shift":
        logits_norm = torch.randn(bsz, device=device)
        logits_norm = logits_norm * args.sigmoid_scale  # larger scale for more uniform sampling
        timesteps = logits_norm.sigmoid()
        mu = get_lin_function(y1=0.5, y2=1.15)((h // 2) * (w // 2))
        timesteps = time_shift(mu, 1.0, timesteps)
        s = timesteps.view(-1, 1, 1, 1)
        timesteps = timesteps * 1000.0
    else:
        # Sample a random timestep for each image
        # for weighting schemes where we sample timesteps non-uniformly
        u = compute_density_for_timestep_sampling(
            weighting_scheme=args.weighting_scheme,
            batch_size=bsz,
            logit_mean=args.logit_mean,
            logit_std=args.logit_std,
            mode_scale=args.mode_scale,
        )
        indices = (u * noise_scheduler.config.num_train_timesteps).long()
        timesteps = noise_scheduler.timesteps[indices].to(device=device)
        s = sigmas = get_sigmas(noise_scheduler, timesteps, device, n_dim=latents.ndim, dtype=dtype)

    # Add noise to the latents according to the noise magnitude at each timestep
    # (this is the forward diffusion process)
    if args.ip_noise_gamma:
        xi = torch.randn_like(latents, device=latents.device, dtype=dtype)
        if args.ip_noise_gamma_random_strength:
            ip_noise_gamma = (torch.rand(1, device=latents.device, dtype=dtype) * args.ip_noise_gamma)
        else:
            ip_noise_gamma = args.ip_noise_gamma
        noisy_model_input = s * (noise + ip_noise_gamma * xi) + (1.0 - s) * latents
    else:
        noisy_model_input = s * noise + (1.0 - s) * latents

    return noisy_model_input.to(dtype), timesteps.to(dtype), sigmas
```

rockerBOO: Thank you for going through this! This fix looks appropriate now. 

I did a deep dive today to try and learn a little more about the sigmas and timesteps and their relationship. Sigmas would be the values between 0 and 1 here and timesteps should be the integer of the timestep. I will have a version up soon for this.

I am working through fixing the loss functions as well and the inconsistencies of when timestep and sigma is used is very confusing to read.

Also I noticed that for flux, at least, it isn't applying a sqrt to the sigmas and your assessment is accurate.

rockerBOO: Moved the broadcasting out of get_sigmas (which was only being used in this function) and then we can broadcast them all in 1 line. Sigmas are stated for clarity now instead of logits_norm/timesteps being mixed together. Added tests to make sure it's all working.

PR #2007, opened 2025-03-26
Support more checkpoint files for flux

rockerBOO:
Goal is to allow more flexibility in the file structure of models to be loaded. It is a little permissive in terms of getting more than it may need to but a starting point to consider. 

Added tests to make sure it is working as intended. 

Get checkpoint paths for flux models
- huggingface directory structure
    - huggingface sharded safetensors files
    - in transformer directory 
    - plain directory
- single safetensors files

### DIFF
--- a/library/flux_utils.py
+++ b/library/flux_utils.py
@@ -1,5 +1,7 @@
 import json
 import os
+from pathlib import Path
+import re
 from dataclasses import replace
 from typing import List, Optional, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


rockerBOO:
This has been working well for me, but it's hard to know if it's working in all cases. The tests are there to help but would like some other testing. Having a few people use this on my boo branch and haven't heard anyone have issues though.

PR #2010, opened 2025-03-27
Feature to improve training quality via detection of out-of-tolerance latent mean/std values

araleza:
This is a new feature, currently only for Flux LoRA training (although it could be applied to full fine-tune later too at least). It analyses the latents of training images, and checks that their mean (average) and standard deviation values are near 0.0 and 1.0 respectively.

It was requested as a feature here: [std/mean detection code.](https://github.com/kohya-ss/sd-scripts/discussions/294#discussioncomment-12506088)

Flux is a diffusion model that takes gaussian noise which is set to have a mean of 0.0 and an std of 1.0 as a starting image. So having training images with that same characteristic may offer improved training as the network does not have to learn how to adjust mean and std values. Current diffusion models don't seem to be good at that.

The default tolerances for detection can be set via `--latent_threshold_warn_levels=mean,std_max`, e.g. `--latent_threshold_warn_levels=0.15,1.40`. The test can be disabled using `--latent_threshold_warn_levels=disable`

If images do not pass the threshold test, then a warning message appears like this:

image

The std_max value sets the upper limit for the standard deviation. A lower limit is also set to 1.0 / std_max. For example, a std_max value of 1.40 also creates a lower threshold of around 0.714.

Sometimes it's not obvious why the mean and std values are not near 0 and 1. In that case, a parameter `--latent_threshold_visualizer` can be passed in which will show the latent average values in a window. (This has been tested on Ubuntu Linux. Please can someone try it on Windows? But it should probably work). 

image
image

Various changes that can move latent means/stds towards 1,0 are possible. e.g.:

* Adjusting image brightness/contrast
* Adjusting image gamma curve
* Adjusting image saturation
* Drawing lighter areas in black backgrounds / darker areas in white backgrounds
* Replacing backgrounds with inpainting
* Redrawing parts of the image with img2img
* Downsizing the image by 2x to remove noise / sharpness

Or even:

* Deleting the image from the training set

One thing I've found in the small number of days I've been training with images that are closer to mean/std 1,0 is that I've had to raise the alpha value of my training, and also reduce the LR. I think that might be because the 'gravity well' of the model remaining closer to base model quality is stronger due to it not being disrupted by training images that are outside the 1,0 distribution.

### DIFF
--- a/library/flux_train_utils.py
+++ b/library/flux_train_utils.py
@@ -550,6 +550,137 @@ def save_flux_model_on_epoch_end_or_stepwise(
 
 # endregion
 
+# region Latent analysis
+
+def check_latent_means_and_stds_against_thresholds(thresholds_string, latent_threshold_visualizer, image_data):
+
+    # Skip mean/std check?
+    if thresholds_string.lower() == "disable":
+        return
+
+    # Thresholds should be in 'mean,std' format; split on comma
+    parts = thresholds_string.split(',')
+    if len(parts) != 2:
+        logger.error(f"latent_threshold_warn_levels was set to '{thresholds_string}', "
+            "Expected latent threshold warning string to either be in 'mean,std' format, or 'disable'")
+        return
+    
+    mean_thresh = float(parts[0].strip())  # Magnitude
+    std_thresh_max = float(parts[1].strip())  # This threshold is the max value. The min value of 1.0 / std_thresh is also tested against
+
+    if std_thresh_max < 1.0:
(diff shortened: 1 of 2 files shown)


araleza:
I've now tested whether using `--ip_noise_gamma 0.1` and `--ip_noise_gamma_random_strength` is enough to make training work well even with images that have out-of-tolerance mean/stds. But no, they don't.

As soon as I bring the training images closer to mean 0.0 std 1.0, my training quality improved significantly. (e.g. eyes that were distorted became clear, slightly mushy lines became firmer and straighter etc.)

PR #2018, opened 2025-03-30
Resize interpolation small fix

kohya-ss:
Small fix to #1936.
- Use PIL for Lanczos and box interpolation.
- Update documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,9 @@ The command to install PyTorch is as follows:
 
 ### Recent Updates
 
+Mar 30, 2025:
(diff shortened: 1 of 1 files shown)


LexSong:
Bug:
```
resize_interpolation = "lanczos" if width > resized_width and height > resized_height else "area"
```
this line is assign resize_interpolation in wrong way.

And We should do area better if width >> resized_width or height >> resized_height 

better do it like:

```
if width > resized_width * 2 or height > resized_height * 2:
    resize_interpolation = "area"
else:
    resize_interpolation = "lanczos"

LexSong:
I made a PR to fix it: https://github.com/kohya-ss/sd-scripts/pull/2022

PR #2022, opened 2025-04-01
Fix size parameter types and improve resize_image interpolation

LexSong:
This PR fixes two issues in the `resize_image` function:

1. **Make sure size parameters to be integers**  
   - The function sometimes got width and height as non-integer types, (like Tensor), which caused crashes when calling cv2.resize() 
   - I added `int()` to make sure they’re always python integers. This stops the errors.

2. **Improve interpolation logic**  
   - The old code used 'lanczos' when shrinking images and 'area' when not, which was backwards.  
   - Now, it uses 'area' only when the image shrinks a lot (to half size or smaller), since 'area' is better for big reductions. Otherwise, it uses 'lanczos'.


LexSong:
I tried with 90% downscale and there is no obvious quality difference. As you said, let's simply use AREA for all downscaling.

May I ask why we're using PIL's lanczos not CV2's lanczos?

LexSong:
I see, thanks for the reply. It's possible caused by gamma correction step. which I think it's skipped in opencv implementation.

PR #2042, opened 2025-04-10
experimental: Documentation Updates with Gemini

kohya-ss:
Workflow:

1. Give the source code etc. to Gemini and have them write the documentation in Japanese or English.
2. Correct errors in documentation.
3. Add translation with Claude.
4. Final check.

Gemini seems to work well even with 200k tokens.


rockerBOO:
Thank you for doing this! Seems like a good idea. 

One thing I was doing just today was documenting the network parameters, like lora_flux parameters. They get hidden in the updates README but hard to point people to a place. Same idea I used for that, 

1. Provide `create_network()` function to a LLM
2. Correct any mistakes 

`create_network()` does a good job for all the networks I believe so could be a straight forward process and possibly aligned if provided a clean example after the first one.

rockerBOO:
I think this seems applicable in it's current state. I have been sharing lines from these docs but pointing into this PR is a bit more steps. If you want a further review I can give it a deeper look.

kohya-ss:
> I think this seems applicable in it's current state. I have been sharing lines from these docs but pointing into this PR is a bit more steps. If you want a further review I can give it a deeper look.

Thank you. Sorry for the late reply.

The need for documentation has increased even more with the addition of Lumina-Image etc. Although the documentation is not yet complete, I would like to merge this PR.

The documentation will continue to be improved.

PR #2053, opened 2025-04-19
fix: update hf_hub_download parameters to fix wd14 tagger regression

GlenCarpenter:
https://github.com/bmaltais/kohya_ss/issues/3178

This fixes a regression introduced the upgrade to `huggingface-hub==0.29.3`

- use named kwargs
- `force_filename` is now deprecated so it has been removed
- use `local_dir` kwarg


GlenCarpenter:
Hello, please advise as to what branch I should target, assuming `dev`

PR #2056, opened 2025-04-20
Added --blocks_to_swap_while_sampling, may allow faster sample image generation

araleza:
I've recently switched over to doing Flux full fine tuning instead of LoRA training. But I've found that sample image generation while training is very slow. I'm using `--blocks_to_swap 35` which lets me have a batch size of 5. This block-swapping persists during sample inference, increasing sample image generation time.

The reason that block swapping is useful while training is because it saves a lot of VRAM, allowing a larger batch size, and for the memory needed for the optimizer's structures, e.g. momentum. But these are not required when doing sample image inference. If I open up nvtop I can see that my VRAM is mostly unused during this time.

This new option allows the number of blocks to swap to be set to a lower number while generating sample images. This may allow faster image generation. e.g. On my current setup where I'm using 50 sampling steps per image, putting `--blocks_to_swap_while_sampling 2` reduces the time per image from around 3 minutes to around 1 minute 48 secs. That might not sound like a big difference at first, but if I generate around 100 images over a run, this saves around 2 hours in total.

### DIFF
--- a/library/custom_offloading_utils.py
+++ b/library/custom_offloading_utils.py
@@ -194,18 +194,25 @@ class ModelOffloader(Offloader):
 
         return backward_hook
 
-    def prepare_block_devices_before_forward(self, blocks: list[nn.Module]):
+    def prepare_block_devices_before_forward(self, blocks: list[nn.Module], override_blocks_to_swap = None):
         if self.blocks_to_swap is None or self.blocks_to_swap == 0:
             return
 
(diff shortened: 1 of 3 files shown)


araleza:
Thanks for the reviews, @rockerBOO and @kohya-ss . I'll take a look at the code soon and try to include your suggestions for improvement. :)

PR #2057, opened 2025-04-20
Feature: Support for resizing multiple format loras.

Symbiomatrix:
Various trainers use different naming conventions for loras, which break the resize script. A ubiquitous one is lora_A / lora_B instead of lora_down / lora_up. This change reads whichever convention was used in the given model and adjusts all hardcoded references to that name.
A bit of a patchwork fix using global variables since I don't know the full flow of the code, but I've tested it extensively over several types of base models' loras (from 1.5 to wan) and it works for every lora I've tested.

### DIFF
--- a/networks/resize_lora.py
+++ b/networks/resize_lora.py
@@ -20,6 +20,12 @@ logger = logging.getLogger(__name__)
 
 MIN_SV = 1e-6
 
+# Tune layers to various trainer formats.
+LORAFMT1 = ["lora_down", "lora_up"]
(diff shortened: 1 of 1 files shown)


woct0rdho:
In f-strings, we need to replace `LORAFMT[0]` and `LORAFMT[1]` with `{LORAFMT[0]}` and `{LORAFMT[1]}`

Symbiomatrix:
@woct0rdho Right, oops. Fixed. Eyeballing refactors in PR versions, bad habit of mine.

PR #2080, opened 2025-05-08
Add Partitioned VAE

rockerBOO:
Partitioned VAE allows the VAE to handle larger resolution images, and also allows us to adjust the packing to have larger patches. This might help with high resolution (2k-4k images) to be processed. 

https://github.com/zhang0jhon/diffusion-4k
https://arxiv.org/abs/2503.18352

In conjunction they also use Wavelet loss to capture the finer detail. This  is just the partitioned VAE aspects. 

Also applies in the sampling, where I refactored it so it makes a noisy latent like tensor and then makes the appropriate patches like we do in training. Instead of going straight to the patched dimensions and makes it a little clearer how it works, while allowing us to handle these differently.

### DIFF
--- a/flux_train_network.py
+++ b/flux_train_network.py
@@ -141,6 +141,10 @@ class FluxNetworkTrainer(train_network.NetworkTrainer):
 
         ae = flux_utils.load_ae(args.ae, weight_dtype, "cpu", disable_mmap=args.disable_mmap_load_safetensors)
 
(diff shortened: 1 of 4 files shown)


rockerBOO:
This is probably still very experimental as this implementation has a few parts and results in an image that seems a bit more "zoomed in". Which i don't think is ideal to it being implemented correctly.

PR #2108, opened 2025-06-03
Add tests for syntax checking training scripts

rockerBOO:
Add simple tests for testing that training scripts import and do not produce syntax errors or missing modules.


kohya-ss:
Thank you, this is really wonderful! 

It might be a good idea to write somewhere that contributors should run tests when they submit a PR.

PR #2121, opened 2025-06-13
Update IPEX libs

Disty0:
Makes IPEX libs compatible with Torch 2.7+. required for Battlemage.   
Removed gradscaler hijack module and uses the PyTorch built-in one instead.  
Updates the deprecated OpenVINO device=GPU_FP32 syntax to device=GPU + precision=FP32 syntax.  

Also side note, i do have my ipex libs as pip installable module via pip install git+https://github.com/Disty0/ipex_to_cuda  
I can convert the lib in this repo to use external package so users can update it without a PR to sd-script if you want but that can be a security issue for you too.


kohya-ss:
Thank you for this!

> Also side note, i do have my ipex libs as pip installable module via pip install git+https://github.com/Disty0/ipex_to_cuda
> I can convert the lib in this repo to use external package so users can update it without a PR to sd-script if you want but that can be a security issue for you too.

That would be convenient for users. However, as you said, dependent libraries are a security issue.
If you don't mind, please submit a pull request and we'll merge it as soon as possible. If you want to switch to an installable module, please let us know.

Disty0:
> If you don't mind, please submit a pull request and we'll merge it as soon as possible. If you want to switch to an installable module, please let us know.  

I don't really understand what you meant by that but i am fine with either of the option. I mentioned it in case if you prefer to use the ipex libs as a dependency instead of built-in to the codebase.

kohya-ss:
> I don't really understand what you meant by that but i am fine with either of the option. I mentioned it in case if you prefer to use the ipex libs as a dependency instead of built-in to the codebase.

Sorry for the confusion. I was wondering if opening a PR every time might be a burden for you. I think the current state is probably better for users because they can use it without having to do anything.

PR #2153, opened 2025-07-16
Fix alphas cumprod after add_noise for DDIMScheduler

### LINKED ISSUE #2102: ⛔[SD3 branch] Huber+SNR RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
Pytorch 2.7.0, CUDA 12.8

What: RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! when using Huber + SNR

Where: train_util.py, line 6025

Fix:

From
`alphas_cumprod = torch.index_select(noise_scheduler.alphas_cumprod, 0, timesteps.cpu())
`
To
`alphas_cumprod = torch.index_select(noise_scheduler.alphas_cumprod, 0, timesteps.to(noise_scheduler.alphas_cumprod.device))
`

rockerBOO:
Fixes #2102 

The issue is 

https://github.com/kohya-ss/sd-scripts/blob/7c075a9c8d234fccf8e0d66b9538a0b17bf4b13f/library/train_util.py#L5977-L6011

noise_scheduler.add_noise() moves the alphas_cumprod to the GPU

And you can see it how it is working in the DDIMScheduler

https://github.com/huggingface/diffusers/blob/a4df8dbc40e170ff828f8d8f79c2c861c9f1748d/src/diffusers/schedulers/scheduling_ddim.py#L474-L498

So this issue is for sd and sdxl models specifically. We can move the alphas_cumprod back to the CPU after.

### DIFF
--- a/library/train_util.py
+++ b/library/train_util.py
@@ -6008,6 +6008,8 @@ def get_noise_noisy_latents_and_timesteps(
     else:
         noisy_latents = noise_scheduler.add_noise(latents, noise, timesteps)
 
+    noise_scheduler.alphas_cumprod = noise_scheduler.alphas_cumprod.cpu()
+
     return noise, noisy_latents, timesteps
 
 


kohya-ss:
Thank you, this makes sense! I will merge this tomorrow.

rockerBOO:
Maybe we should add a note about why we are doing that there for the future.

kohya-ss:
I looked into why this hadn't been a problem until now, and it turns out Diffusers code has changed: https://github.com/huggingface/diffusers/pull/6704

The size of alphas_cumprod is 1,000, so I don't think the overhead is that large.

rockerBOO:
I think that PR might highlight that we might just want to move the timesteps to that device or just use that device in our huber lookup. We are moving the timesteps to the cpu there, and if it is having a big enough impact that they are showing, it might make sense to do that.

We can just move the `timesteps` to the `noise_scheduler.alphas_cumprod` device to match.

PR #2158, opened 2025-07-21
Feature chroma support per block LR specification

maybleMyers:
This enables you to specify per block learning rate multipliers. Chroma suffers from lines and other abnormalities if you train certain blocks too much. This is an example addition to your launch command with known good values:

--network_args 'lr_if_contains=double_blocks$$0$$:0' 'lr_if_contains=double_blocks$$1$$:0' 'lr_if_contains=double_blocks$$2$$:0.00456' 'lr_if_contains=double_blocks$$3$$:0.006' 'lr_if_contains=double_blocks$$4$$:0.00722' 'lr_if_contains=double_blocks$$5$$:0.00822' 'lr_if_contains=double_blocks$$6$$:0.009' 'lr_if_contains=double_blocks$$7$$:0.00956' 'lr_if_contains=double_blocks$$8$$:0.00989' 'lr_if_contains=double_blocks$$9$$:0.01' 'lr_if_contains=double_blocks$$10$$:0.00989' 'lr_if_contains=double_blocks$$11$$:0.00956' 'lr_if_contains=double_blocks$$12$$:0.009' 'lr_if_contains=double_blocks$$13$$:0.00822' 'lr_if_contains=double_blocks$$14$$:0.00722' 'lr_if_contains=double_blocks$$15$$:0.006' 'lr_if_contains=double_blocks$$16$$:0.00456' 'lr_if_contains=double_blocks$$17$$:0' 'lr_if_contains=double_blocks$$18$$:0'

### DIFF
--- a/README-ja.md
+++ b/README-ja.md
@@ -155,11 +155,12 @@ masterpiece, best quality, 1boy, in business suit, standing at street, looking b
 
   `#` で始まる行はコメントになります。`--n` のように「ハイフン二個＋英小文字」の形でオプションを指定できます。以下が使用可能できます。
 
-  * `--n` Negative prompt up to the next option.
-  * `--w` Specifies the width of the generated image.
-  * `--h` Specifies the height of the generated image.
-  * `--d` Specifies the seed of the generated image.
-  * `--l` Specifies the CFG scale of the generated image.
-  * `--s` Specifies the number of steps in the generation.
+  * `--n` ネガティブプロンプト（次のオプションまで）
+  * `--w` 生成画像の幅を指定
+  * `--h` 生成画像の高さを指定
+  * `--d` 生成画像のシード値を指定
+  * `--l` 生成画像のCFGスケールを指定。FLUX.1モデルでは、デフォルトは `1.0` でCFGなしを意味します。Chromaモデルでは、CFGを有効にするために `4.0` 程度に設定してください
+  * `--g` 埋め込みガイダンス付きモデル（FLUX.1）の埋め込みガイダンススケールを指定、デフォルトは `3.5`。Chromaモデルでは `0.0` に設定してください
+  * `--s` 生成時のステップ数を指定
 
   `( )` や `[ ]` などの重みづけも動作します。
--- a/README.md
+++ b/README.md
@@ -16,6 +16,13 @@ If you are using DeepSpeed, please install DeepSpeed with `pip install deepspeed
 
 ### Recent Updates
 
+Jul XX, 2025:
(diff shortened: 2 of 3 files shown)


kohya-ss:
Thank you, I think this is a very useful feature!

However, SDXL LoRA in sd-scripts already has a function to specify the learning rate per block, which is an option called `block_lr`. I think it would be better to unify the notation.

Also, in existing LoRA, the way to specify rank and alpha per block differed depending on the architecture (SD, SDXL, FLUX). I would like to be able to specify this using a unified notation (for example, a regular expression-based notation) in future, so I'd like to adopt the future-proof notation at this time.

I will consider how to implement this, so I would appreciate it if you could wait for a while.

kohya-ss:
I would like to implement the learning rate and dim (rank) specification using regular expressions using the following function. What do you think?

---

### 6.4. Regular Expression-based Rank/LR Configuration / 正規表現によるランク・学習率の指定

You can specify ranks (dims) and learning rates for LoRA modules using regular expressions. This allows for more flexible and fine-grained control than specifying by layer.

These settings are specified via the `network_args` argument.

*   `network_reg_dims`: Specify ranks for modules matching a regular expression. The format is a comma-separated string of `pattern=rank`.
    *   Example: `--network_args "network_reg_dims=single.*_mod.*=4,img_attn=8"`
    *   This sets the rank to 4 for modules whose names start with `single` and contain `_mod`, and to 8 for modules containing `img_attn`.
*   `network_reg_lrs`: Specify learning rates for modules matching a regular expression. The format is a comma-separated string of `pattern=lr`.
    *   Example: `--network_args "network_reg_lrs=single_blocks.*=1e-5,double_blocks.*=5e-5"`
    *   This sets the learning rate to `1e-5` for modules whose names start with `single_blocks`, and to `5e-5` for modules whose names start with `double_blocks`.

**Notes:**

*   Settings via `network_reg_dims` and `network_reg_lrs` take precedence over the global `--network_dim` and `--learning_rate` settings.
*   If a module name matches multiple patterns, the setting from the last matching pattern in the string will be applied.
*   These settings are applied after the block-specific training settings (`train_double_block_indices`, `train_single_block_indices`).

maybleMyers:
I think your solution is better xD. It will be a more condensed command and can accomplish all of the same things. It will allow more variations as well. It would be nice if we could also use it for fine tuning and not just lora.

kohya-ss:
> I think your solution is better xD. It will be a more condensed command and can accomplish all of the same things. It will allow more variations as well. It would be nice if we could also use it for fine tuning and not just lora.

Thank you for verification!

I've added the feature. Please see [the doc](https://github.com/kohya-ss/sd-scripts/blob/feature-chroma-support/docs/flux_train_network.md#65-regular-expression-based-ranklr-configuration--%E6%AD%A3%E8%A6%8F%E8%A1%A8%E7%8F%BE%E3%81%AB%E3%82%88%E3%82%8B%E3%83%A9%E3%83%B3%E3%82%AF%E5%AD%A6%E7%BF%92%E7%8E%87%E3%81%AE%E6%8C%87%E5%AE%9A) for details.

I'd like to close this PR.

PR #2159, opened 2025-07-23
Big image quality improvement! Kahan summation for Adafactor-optimized Flux FFT

araleza:
If you've been doing full fine tuning with Flux, you've likely seen the image quality improvements that the stochastic rounding feature (available when using `--fused_backward-pass` with the supplied Adafactor optimizer`) can offer. That feature improves the casting of the f32 'weight + gradient update' values back to bf16 during each training step by randomly moving the value either up or down to the next rounded bfloat16 value, depending on the distance the f32 value is between the two adjacent quantized bf16 values.

But although stochastic rounding offers a large improvement to bf16 training, its randomness doesn't always produce ideal results. As the "Revisiting BFloat16 Training" paper - https://arxiv.org/pdf/2010.06192 - notes:

> "using stochastic rounding alone could not fully match 32-bit training on all models."

This same paper then goes on to note a different technique - **Kahan summation** - that does match the performance of 32-bit training:

> "To address this, we show that Kahan summation for model weight updates closes remaining gaps on all the models we consider"

Kahan summation is a technique where the f32 values are rounded to the nearest bf16 value, and then the offset between the original f32 and the quantized bf16 value is recorded. This offset is then sent back to the CPU so it doesn't use up VRAM past that point. Then, on the next training step, instead of just taking the bf16 weight value plus the gradient update for that step, the offset that was lost from the previous step is added back on as well. And then that process is repeated.

It means that a bf16 value that is updating by (e.g.) 20% of the distance between that bf16 value and the next adjacent bf16 value will go up once in every 5 training steps, rather than stochastic rounding's random approximation of once in every 5 steps. Stochastic rounded values bounce up and down fairly unpredictably, but the Kahan summed values are more stable.

The technique comes at a price: training steps take nearly twice as long in my tests so far. (I'm training at batch size 5 on a 5090 RTX card). So it's probably best to use stochastic rounding for most of the training, and then switch to Kahan summation with `--kahan-summation` for the final polish phase. The slowdown comes from copying the values from the GPU to main memory and back again on each step.

Now I've implemented Kahan summation in sd-scripts, the quality improvements that it achieves are impressive. Flux.dev training does seem to be one of the cases where Kahan summation significantly exceeds stochastic rounding in terms of image quality. Very low LRs such as 2e-7 (which suffer quantization randomness with stochastic rounding) work great with Kahan summation.

If you (e.g. @kohya-ss) want to see what this feature can do, then why not grab this branch ,switch on `--kahan-summation`, set the LR to 2e-7 and the batch size to 5, and then try running a final polish pass on an already-trained FFT model that you have? The quality improvements are almost immediate, as it allows the weights to 'settle' into a more stable pattern.

### DIFF
--- a/flux_train.py
+++ b/flux_train.py
@@ -381,9 +381,24 @@ def train(args):
             raise ValueError("Schedule-free optimizer is not supported with blockwise fused optimizers")
         optimizer_train_fn = lambda: None  # dummy function
         optimizer_eval_fn = lambda: None  # dummy function
+
+        if args.optimizer_type == "adafactor" and args.full_bf16:
+            logger.warning("Use of --blockwise_fused_optimizers with Adafactor optimizer prevents stochastic/kahan weight updates.")
     else:
         _, _, optimizer = train_util.get_optimizer(args, trainable_params=params_to_optimize)
         optimizer_train_fn, optimizer_eval_fn = train_util.get_optimizer_train_eval_fn(optimizer, args)
+
+    # Pass any Kahan summation arg to the optimizer
+    if args.kahan_summation:
+        # Self check parameter compatibility
+        if args.optimizer_type != "adafactor":
+            logger.warning("Kahan summation has been requested, but currently this is only supported by the supplied Adafactor optimizer.")
+        if not args.full_bf16:
+            logger.warning("Kahan summation require --full_bf16")
+        if args.blockwise_fused_optimizers:
+            logger.warning("Kahan summation has been requested, but it is incompatible with --blockwise_fused_optimizer. "\
+                           "Perhaps try --fused_backward_pass instead.")
+    optimizer.use_kahan_summation = args.kahan_summation
 
     # prepare dataloader
     # strategies are set here because they cannot be referenced in another process. Copy them with the dataset
@@ -816,6 +831,12 @@ def setup_parser() -> argparse.ArgumentParser:
         help="enable blockwise optimizers for fused backward pass and optimizer step / fused backward passとoptimizer step のためブロック単位のoptimizerを有効にする",
     )
     parser.add_argument(
+        "--kahan-summation",
+        action="store_true",
+        help="Offloads to CPU the float parts lost during bf16 quantization, and re-adds them to the next step / "\
+            "bf16 量子化中に失われた浮動小数点部分を CPU にオフロードし、次のステップに再度追加します",
+    )
+    parser.add_argument(
         "--skip_latents_validity_check",
         action="store_true",
         help="[Deprecated] use 'skip_cache_check' instead / 代わりに 'skip_cache_check' を使用してください",
@@ -838,13 +859,3 @@ def setup_parser() -> argparse.ArgumentParser:
         help="[EXPERIMENTAL] enable offloading of tensors to CPU during checkpointing / チェックポイント時にテンソルをCPUにオフロードする",
     )
     return parser
-
-
-if __name__ == "__main__":
-    parser = setup_parser()
-
-    args = parser.parse_args()
-    train_util.verify_command_line_training_args(args)
-    args = train_util.read_config_from_file(args, parser)
-
-    train(args)
--- a/library/adafactor_fused.py
+++ b/library/adafactor_fused.py
@@ -28,6 +28,58 @@ def copy_stochastic_(target: torch.Tensor, source: torch.Tensor):
     del result
 
 
+# Kahan summation for bfloat16
+# The implementation was provided by araleza.
+# Base on paper "Revisiting BFloat16 Training": https://arxiv.org/pdf/2010.06192
+
+kahan_residuals = []
+tensor_index = 0
+prev_step = 0
+
+def copy_kahan_(target: torch.Tensor, source: torch.Tensor, step):
+    """
+    Copies source into target using Kahan summations.
+
+    The part of the float32 weight that is lost on conversion to bfloat16 is sent
+    to the CPU until the next step, where it is re-added onto that step's updated
+    weight.  This produces near float32-like weight behavior, although the copies
+    back and forth to main memory result in slower training steps.
+
+    Args:
+        target: the target tensor with dtype=bfloat16
+        source: the target tensor with dtype=float32
+    """
+    global kahan_residuals, tensor_index, prev_step
+
+    # Calculate the group index of the current residual Tensor. Tensors
+    # pass through this copy function in the same order at each step.
+    tensor_index += 1
+    if prev_step != step:  # Starting new step?
+        prev_step = step
+        tensor_index = 0
+
+    # Initialize residuals to 0.0 for first step
+    if len(kahan_residuals) <= tensor_index:
+        kahan_residuals += [torch.zeros_like(source)]
+
+    # Bring the residual from the previous step back from the cpu device, and add it to the
+    # float32 weights of the current step
+    summed = kahan_residuals[tensor_index].detach().to(source.device)  # Residual is float32 type
+    summed.add_(source)
+
+    # Mask off the lower 16 bits of the mantissa, adding 32768 in order to
+    # round-to-nearest when the lower bits are clipped off
+    summed_i32 = summed.view(dtype=torch.int32).detach().clone()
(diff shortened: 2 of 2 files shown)


araleza:
I've now accelerated the kahan summation function by only sending the 16 lower bits that are clipped from the f32 values to the cpu, instead of a floating point offset. With only half of the number of bytes needing to be transferred, it's much quicker. It might be higher quality too, as the bit identical f32 value that was quantized to bf16 is now restored.

FurkanGozukara:
@araleza this sounds amazing

i recently compared majority of optimizers and adafactor was king

definitely gonna test your branch 

can you share your example toml / json? so i need to add --kahan_summation what other?

araleza:
I don't usually use toml / json, so I don't have one to share @FurkanGozukara. But yeah, switching this on is literally just adding --kahan_summation to the command line and that's that.

I don't know how to add that to a toml / json file, but it's just a command line option like the rest, so however you add those, it should work for --kahan_summation too.

I should really add a debug output message that says it's switched on, so you can confirm it's working. Give me a minute and I'll check that in now.

FurkanGozukara:
> I don't usually use toml / json, so I don't have one to share @FurkanGozukara. But yeah, switching this on is literally just adding --kahan_summation to the command line and that's that.
> 
> I don't know how to add that to a toml / json file, but it's just a command line option like the rest, so however you add those, it should work for --kahan_summation too.
> 
> I should really add a debug output message that says it's switched on, so you can confirm it's working. Give me a minute and I'll check that in now.

whatever format you use fine if you put here as a txt file i can compare what i am missing. i will definitely run a comparison training

araleza:
I'm running directly from the command line like this, @FurkanGozukara:

`
python flux_train.py --clip_l /home/ara/Dev/sd3/clip_l.safetensors --t5xxl /home/ara/Dev/flux/flanT5XXLTextEncorder_fp16.safetensors --ae /home/ara/Dev/flux/ae.safetensors --cache_latents_to_disk --save_model_as safetensors --sdpa --persistent_data_loader_workers --max_data_loader_n_workers 2 --logging_dir="/home/ara/Dev/training/earthscape/kohya/log" --gradient_checkpointing --mixed_precision bf16 --save_precision bf16 --full_bf16 --cache_text_encoder_outputs --output_name earthscape --cache_text_encoder_outputs_to_disk --max_train_epochs 900 --save_every_n_steps 40 --save_last_n_steps 80 --enable_bucket --min_bucket_reso=64 --bucket_no_upscale --max_bucket_reso=1024 --resolution="1024,1024" --caption_extension=".txt" --sample_sampler=k_dpm_2 --train_data_dir="/home/ara/Dev/training/earthscape/kohya/img" --enable_wildcard --output_dir output --apply_t5_attn_mask --guidance_scale 1.0 --model_prediction_type raw --sample_prompts="/home/ara/Dev/training/earthscape/kohya/dreambooth/sample/prompt.txt" --sample_every_n_steps="10" --debiased_estimation_loss --alpha_mask --timestep_sampling flux_shift --seed 42 --kahan_summation --loss_type huber --huber_c 1.0 --huber_scale 2.5 --huber_schedule='exponential' --optimizer_type adafactor --optimizer_args "relative_step=False" "scale_parameter=False" "warmup_init=False" --train_batch_size 5 --blocks_to_swap 35 --fused_backward_pass --flip_aug --learning_rate 4e-7 --pretrained_model_name_or_path "/home/ara/Dev/flux/flux1-dev.safetensors"
`

(I switched out my already-trained FFT model for the base flux1-dev model in that command line. If you're actually training from the base with kahan summation, you probably want a higher LR of something like 2e-6, rather than a polishing LR of 4e-7)

Edit: I removed the `--ip_noise_gamma` and `--ip_noise_gamma_random_strength` parameters from my command line, and learning seems much stronger without them, at least for my training datasets.

blackmagic24:
> The technique comes at a price: training steps take around 40% longer in my tests so far.

The StableAdamW optimizer also supports Kahan summation and can be used with sd-scripts without any issues. However, I haven’t noticed any significant performance differences compared to other optimizers. It might still be worth a look.

https://pytorch-optimizers.readthedocs.io/en/latest/optimizer/#pytorch_optimizer.StableAdamW

optimizer_type = "pytorch_optimizer.StableAdamW"
optimizer_args = [
     "kahan_sum=True",
     "weight_decay=0.01",
     "weight_decouple=True",
]

araleza:
Hi @blackmagic24, thanks for your reply.

I did notice that the pytorch_optimizer's adamW implementation had a feature that it referred to as a kahan_sum, but I think it might be an unrelated feature that also shares the name of the technique, as kahan summation is an idea that's applicable to more than one area. When I looked at the source code for pytorch_optimizer's adamw implementation, the only changes that activating its kahan_sum feature does is this:

image

That's from:

https://github.com/kozistr/pytorch_optimizer/blob/5a985ade9c6a887c6507a061d1ce16673882a155/pytorch_optimizer/optimizer/adamw.py#L16

I'm not exactly certain what that section of code does, but it doesn't seem to be keeping the lower bits of the weights between steps, but rather it's doing something with the exponential averages. And it doesn't seem to be sending the bits to the CPU between training steps, so it doesn't seem like it has the same purpose as what my Kahan summation code is doing. But if you have any more information about this feature for adamW, I'd be interested to know about it.

araleza:
Actually, looking more at that AdamW code, it does seem to be doing a floating point sub_() of the bfloat16 value against the 'real' value, and an add_() to the kahan_comp variable. The parts with the exponential moving averages are the same between the kahan version and the non-kahan version, so it's not related to that at all.

I think this code is more like the first version of my pull request, where I kept an f32 offset on the CPU between training steps, rather than directly keeping the lower 16 bits that were lost. It isn't been sent to the CPU between steps though, so it wouldn't work in low-memory conditions like the Adafactor version in my pull request would.

An interesting thing is that that pytorch_optimizer's version of AdamW *doesn't* do stochastic rounding. While it's an interesting question whether Kahan summation beats stochastic rounding for Flux training or not, both of these should strongly outperform an optimizer that isn't doing either of them. So you should see a large improvement for that AdamW optimizer if you switch on kahan_sum. And the fact that you aren't is curious.

I'm sure this is unlikely, but you're not actually training a LoRA instead of doing full fine-tuning are you? Cause LoRAs are f32 all the time, so kahan_sum would have no effect there.

I might give that AdamW optimizer a try at some point soon, although I'll need to check if it even can be run on my 32 GB graphics card when doing full fine tuning.

FurkanGozukara:
i have recently compared AdamW  vs Adafactor and Adafactor yielded best realistic results

araleza:
> i have recently compared AdamW vs Adafactor and Adafactor yielded best realistic results

That might be because sd-scripts' current Adafactor does have stochastic rounding, and the AdamW implementation doesn't. AdamW should typically outperform Adafactor if both are using stochastic rounding - or both aren't - but AdamW uses much more memory, which is why projects often switch to Adafactor in low memory conditions.

araleza:
> I might give that AdamW optimizer a try at some point soon, although I'll need to check if it even can be run on my 32 GB graphics card when doing full fine tuning.

Yep, I can't run fine tuning with pytorch_optimizer.StableAdamW, even on my 32 GB graphics card. If I try to use that optimizer, I get the error message:
```
fused_backward_pass currently only works with optimizer_type Adafactor
```
If I switch off `--fused_backward_pass` to get the optimizer to work, I get a CUDA out of memory error, even with batch size 1 and 35 swapped blocks. So it seems a moot point for anyone fine tuning with a GPU that has 32 GB or less whether the StableAdamW optimizer supports Kahan summation or not.

FurkanGozukara:
> stochastic rounding

this makes sense ty for info

kohya-ss:
Thanks, sorry for the late response. This is very interesting. I'll take a closer look at the code.

araleza:
> I think the code would be more organized if kahan_residuals were stored in state, but is that possible?

Yes, I've now made this change - and the code is simpler now too.

I think that the step and tensor_index values that you mentioned were already correct, but perhaps the code was not clear here. However, now I'm using the optimizer state to store the kahan residuals, both step and tensor_index have been deleted from the copy_kahan_() function, so they are definitely not a problem now. 😊

araleza:
I've also included the Kahan summation code in my separate (and more impactful / higher image quality) AdamW for 5090 branch here:

https://github.com/kohya-ss/sd-scripts/pull/2187

PR #2168, opened 2025-08-03
Update model spec to 1.0.1.

rockerBOO:
Updates to latest [model spec 1.0.1](https://github.com/Stability-AI/ModelSpec)

Refactored a little bit of the code to allow for some more flexibility in adding new metadata parameters. 
- Added `ModelSpecMetadata` as a dataclass 
- `implementation_version`: `sd-scripts/{COMMIT_HASH}`
- `modelspec_thumbnail` can be added as data URL or path (though might be more of a after training thing) 
- Added tests
- New `--metadata_` arguments added for the new options. 

Written by AI but I guided what functions to make, abstractions, and goals.

### DIFF
--- a/library/sai_model_spec.py
+++ b/library/sai_model_spec.py
@@ -1,14 +1,19 @@
 # based on https://github.com/Stability-AI/ModelSpec
 import datetime
 import hashlib
+import argparse
+import base64
+import logging
+import mimetypes
+import subprocess
+from dataclasses import dataclass, field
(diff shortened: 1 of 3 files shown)


kohya-ss:
Thank you, this works quite well! I think this problem existed before this PR, the architecture has an issue in the finetuning of SD3 (like `stable-diffusion-3-3-medium`), so I will only fix that.

PR #2175, opened 2025-08-14
Resize ControlLoRA

woct0rdho:
This PR is based on #2057 . It supports resizing [ControlLoRA](https://github.com/HighCWu/ControlLoRA) (LoRA version of ControlNet). We can save a ControlLoRA using `ControlLoraSave` node in [stability-ComfyUI-nodes](https://github.com/Stability-AI/stability-ComfyUI-nodes), then resize it using `networks/resize_lora.py`.

### DIFF
--- a/networks/resize_lora.py
+++ b/networks/resize_lora.py
@@ -240,7 +240,7 @@ def resize_lora_model(lora_sd, new_rank, new_conv_rank, save_dtype, device, dyna
         for key, value in tqdm(lora_sd.items()):
             weight_name = None
             if LORAFMT[0] in key:
-                block_down_name = key.rsplit(f".LORAFMT[0]", 1)[0]
+                block_down_name = key.rsplit(f".{LORAFMT[0]}", 1)[0]
                 weight_name = key.rsplit(".", 1)[-1]
                 lora_down_weight = value
             else:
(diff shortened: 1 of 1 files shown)


kohya-ss:
Thank you, this is a good idea to generalize the converter. Could you please check the Copilot review?
I also think that passing LORAFMT as a parameter instead of globally is a good idea.

woct0rdho:
I agree that using global variables is a bad practice (it was a quick patch based on the previous PR). I'll refactor it.

networks/resize_lora.py
```diff
@@ -192,24 +199,11 @@ def rank_resize(S, rank, dynamic_method, dynamic_param, scale=1):
 
 
 def resize_lora_model(lora_sd, new_rank, new_conv_rank, save_dtype, device, dynamic_method, dynamic_param, verbose):
-    network_alpha = None
-    network_dim = None
+    max_old_rank = None
+    new_alpha = None
```
woct0rdho, line 203: `max_old_rank` and `new_alpha` are returned by this function and used in the report. They only make sense if the new lora is not dynamic. If the old lora is dynamic, we report the max rank among all parameters.

networks/resize_lora.py
```diff
@@ -192,24 +199,11 @@ def rank_resize(S, rank, dynamic_method, dynamic_param, scale=1):
...
-        if network_alpha is None and "alpha" in key:
-            network_alpha = value
-        if network_dim is None and "lora_down" in key and len(value.size()) == 2:
-            network_dim = value.size()[0]
-        if network_alpha is not None and network_dim is not None:
-            break
-        if network_alpha is None:
-            network_alpha = network_dim
-
-    scale = network_alpha / network_dim
```
woct0rdho, line 211: This `scale` is not used below.

woct0rdho:
I've refactored the code to avoid the mutable global variable (there is no need to rename `lora_down` and `lora_up` in `param_dict`). I've tested that the 3 LoRA formats can be converted by this script, and they work as intended in ComfyUI.

kohya-ss:
Thank you for update! This looks good.

PR #2178, opened 2025-08-16
Feat: Update libraries, remove warnings

kohya-ss:
- Upgrade PyTorch >= 2.6.0
- Upgrade the dependent libraries to versions that are confirmed to work with Musubi Tuner.
- Resolve all possible warnings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ If you are using DeepSpeed, please install DeepSpeed with `pip install deepspeed
 ### Recent Updates
 
 Aug 28, 2025:
-- In order to support the latest GPUs and features, we have updated the **PyTorch and library versions**. There are many changes, so please let us know if you encounter any issues.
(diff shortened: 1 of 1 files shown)


kohya-ss:
`pytorch-lightning` is finally removed... Also `gradio` and `altair` doesn't seem to be needed anymore.

rockerBOO:
Can add python 2.6.0 to the github workflow too. Maybe we keep both 2.4.0 and 2.6.0 for now.

kohya-ss:
> Can add python 2.6.0 to the github workflow too. Maybe we keep both 2.4.0 and 2.6.0 for now.

That's right. I believe the project will continue to work with PyTorch 2.4.0. I'll look into how to change the flow, but it might be a good idea to merge this PR first.

rockerBOO:
Would need to change this line, to add 2.6.0

https://github.com/kohya-ss/sd-scripts/blob/2bfda1271bcdbe13e823579fc406f3eaa229573b/.github/workflows/tests.yml#L25

kohya-ss:
The scripts updated to remove the warnings seem to work with previous library versions as well, so users can upgrade their environments at any time.

There seems to be no problem merging this PR.

PR #2190, opened 2025-08-28
Fix validation split for subsets

rockerBOO:
Subsets could not get a separate validation split but it is supported. Might also want to add `validation_seed` but I have not tested it.

### DIFF
--- a/library/config_util.py
+++ b/library/config_util.py
@@ -197,6 +197,7 @@ class ConfigSanitizer:
         "caption_prefix": str,
         "caption_suffix": str,
         "custom_attributes": dict,
+        "validation_split": float,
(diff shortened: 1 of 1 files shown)


kohya-ss:
https://github.com/kohya-ss/sd-scripts/blob/884e07d73e2a5e1a7a8051880690e7b08bc4d84b/library/train_util.py#L2007-L2023

This part references `dataset.validation_split` and `is_training_dataset`, so it looks like it needs some additional changes to work...

rockerBOO:
Sorry, I had assumed I had this working. But it is not, as you point out. Making it work will be a bit more work. I updated the documentation in #2196 for how to do this with the current system.

PR #2192, opened 2025-08-31
Doc update for latest features

kohya-ss:


kohya-ss:
Some documents still need to be fixed, but we will merge this before HunyuanImage-2.1.

PR #2194, opened 2025-09-04
Fix the 'off by 1' problem in dynamically resized LoRA rank

woct0rdho:
If I understand correctly, the index returned by `index_sv_cumulative` and `index_sv_fro` is larger by 1 than it should be. For example, `torch.searchsorted` works like this:
```python3
>>> cumulative_sums = torch.tensor([0.9, 0.95, 1.0])
>>> print(torch.searchsorted(cumulative_sums, 0.94))
tensor(1)
>>> print(torch.searchsorted(cumulative_sums, 0.95))
tensor(1)
>>> print(torch.searchsorted(cumulative_sums, 0.96))
tensor(2)
```
If the user wants to keep 94% or 95% of `S`, then `index` should be 1 (and `new_rank` in the function `rank_resize` should be 2).

This problem is particularly visible when the new LoRA is rank 1 (I've seen many LoRAs on the internet that are almost rank 1). In the example above, if the user wants to keep 80% or 90% of `S`, then `index` should be 0, and `new_rank` should be 1.

### DIFF
--- a/networks/resize_lora.py
+++ b/networks/resize_lora.py
@@ -78,7 +78,7 @@ def index_sv_fro(S, target):
 def index_sv_ratio(S, target):
     max_sv = S[0]
     min_sv = max_sv / target
-    index = int(torch.sum(S > min_sv).item())
+    index = int(torch.sum(S > min_sv).item()) - 1
     index = max(0, min(index, len(S) - 1))
 
     return index


kohya-ss:
Thank you for the fix!

The correction for `index_sv_cumulative` and `index_sv_fro looks` correct — removing the +1 from searchsorted results properly aligns the 0-based index with the + 1 in rank_resize.

However, I think `index_sv_ratio` may still have an off-by-one issue. `torch.sum(S > min_sv)` returns a count (1-based), not a 0-based index. Since `rank_resize` does `new_rank = index + 1`, the resulting rank is always 1 higher than expected.

For example, with `--dynamic_method sv_ratio --dynamic_param 2`:
```
S = [1.0, 0.4, 0.1, ...]
min_sv = 1.0 / 2 = 0.5
torch.sum(S > 0.5) = 1  # only S[0] qualifies → expected rank = 1
# but rank_resize computes: new_rank = 1 + 1 = 2
```
A possible fix would be:
```
index = int(torch.sum(S > min_sv).item()) - 1
```

This converts the count to a 0-based index, consistent with the other two functions.

I would be grateful if you could include this fix for consistency in the pull request. I can also make the update after the merge, so please let me know if you'd like me to do so.

woct0rdho:
You're right. `index_sv_ratio` also needs this.

kohya-ss:
Thank you for quick update!

PR #2196, opened 2025-09-08
Fix validation dataset documentation to not use subsets

rockerBOO:
Subsets are not yet supported for validation_split so updating the documentation to show how to organize the validation splits. 

Related #2190


kohya-ss:
Thank you for this! I'll merge this sooner.

Issue #2201, opened 2025-09-16
lumina reversed timesteps seems not handled in some funcs


urlesistiana:
During my training, I accidently found if I use the default `shift` timestep sampling method, even after thousands of steps, the model still learned nothing, the sample images are identical. Which is very strange. Because when using `nextdit_shift` it only needs ~200 steps to show the noticeable changes in sample images

Lumina uses t=0 as the noise and t=1 as the image. Which is reversed, unlike other models  (t=1 as the noise and t=0 as image). 

I found this causes some funcs not working as expected, since they expects the usual timesteps.

In

https://github.com/kohya-ss/sd-scripts/blob/f5d44fd4874ef19dfbd764158ad29187184c6fd0/library/lumina_train_util.py#L806

`nextdit_shift` reversed the timesteps in `time_shift()`. Other timestep sampling methods don't work properly. They still shift toward t=1.

`--min_snr_gamma` is also affected, maybe more.

I also found that diffusers still uses the usual timesteps in its api, and only reverse it before passing to the network.

https://github.com/huggingface/diffusers/blob/0f252be0ed42006c125ef4429156cb13ae6c1d60/src/diffusers/pipelines/lumina2/pipeline_lumina2.py#L698

Maybe we should do this as well?

https://github.com/kohya-ss/sd-scripts/blob/f5d44fd4874ef19dfbd764158ad29187184c6fd0/lumina_train_network.py#L271

PR #2215, opened 2025-09-29
Fix Lumina reversed timestep handling (#2201)

duongve13112002:
Description

This PR fixes the issue where Lumina's reversed timesteps (using t=0 as noise and t=1 as image) were not properly handled in some functions. As a result, certain timestep sampling methods (other than nextdit_shift) did not work as expected, causing the model to fail to learn even after thousands of steps.

The fix ensures that timestep handling is consistent with Lumina’s reversed convention.

In addition, this PR introduces a new timestep type named lognorm.

Changes
- Fixed reversed timestep handling in lumina_train_util.py and related functions.
- Adjusted affected methods so that they properly account for t=0 noise / t=1 image convention.
- Added support for a new timestep type: lognorm.

### DIFF
--- a/library/lumina_train_util.py
+++ b/library/lumina_train_util.py
@@ -479,7 +479,6 @@ def time_shift(mu: float, sigma: float, t: torch.Tensor):
     # Since we adopt the reverse, the 1-t operations are needed
     t = 1 - t
     t = math.exp(mu) / (math.exp(mu) + (1 / t - 1) ** sigma)
-    t = 1 - t
     return t
 
 
@@ -854,7 +853,7 @@ def get_noisy_model_input_and_timesteps(
         mu = get_lin_function(y1=0.5, y2=1.15)((h // 2) * (w // 2))
         t = time_shift(mu, 1.0, t)
 
-        timesteps = 1 - t * 1000.0
+        timesteps = t * 1000.0
         t = t.view(-1, 1, 1, 1)
         noisy_model_input = (1 - t) * noise + t * latents
     else:


library/lumina_train_util.py
```diff
@@ -857,6 +858,15 @@ def get_noisy_model_input_and_timesteps(
         timesteps = t * 1000.0
         t = t.view(-1, 1, 1, 1)
         noisy_model_input = (1 - t) * noise + t * latents
+        
+    elif args.timestep_sampling == "lognorm":
```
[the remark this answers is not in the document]
duongve13112002: Ok i will add it into add_lumina_train_arguments in lumina_train_util in the next pull request

duongve13112002:
I have made the adjustments as you requested. Plus, the new timestep type which i implemented is similar to sigmoid so i deleted it and dont add to this pull request. In addition, I have also fixed the issue related to fine-tune the lumina model with multi-GPU. Moreover, according to the current fine-tune model code for Lumina, when args.blockwise_fused_optimizers is enabled, the model’s parameters are not being updated. At the moment, I don’t know how to fix this, so I have disabled this feature to prevent errors for users. Sorry for committing multiple times; I’m making changes using my tablet.

library/lumina_train_util.py
```diff
@@ -854,7 +854,7 @@ def get_noisy_model_input_and_timesteps(
         mu = get_lin_function(y1=0.5, y2=1.15)((h // 2) * (w // 2))
         t = time_shift(mu, 1.0, t)
 
-        timesteps = t * 1000.0
+        timesteps = 1 - t * 1000.0
```
duongve13112002, line 857: I also reversed the sampling of the ‘nextdit_shift’ timestep to synchronize it with the current code.

duongve13112002: I think we don’t need to change anything here because another function calls this one. Changing the code in this function could potentially break the training pipeline, for example, in this function
https://github.com/duongve13112002/sd-scripts/blob/4d24b71c1647f674951f482857c12c74a5a46440/library/lumina_train_util.py#L507-L537

duongve13112002: Ok i will change it right now

kohya-ss:
This PR is replaced with #2225. Sorry for taking time an thank you for this again!

PR #2225, opened 2025-10-13
fix: lumina 2 timesteps handling

### LINKED ISSUE #2201: lumina reversed timesteps seems not handled in some funcs

urlesistiana:
fix #2201 

mainly replaced `get_noisy_model_input_and_timesteps` with the one from flux_train_utils.py.

This also added `flux_shift` timestep sampling method and `--ip_noise_gamma` support.

### DIFF
--- a/library/lumina_train_util.py
+++ b/library/lumina_train_util.py
@@ -475,11 +475,7 @@ def sample_image_inference(
 
 
 def time_shift(mu: float, sigma: float, t: torch.Tensor):
-    # the following implementation was original for t=0: clean / t=1: noise
-    # Since we adopt the reverse, the 1-t operations are needed
-    t = 1 - t
     t = math.exp(mu) / (math.exp(mu) + (1 / t - 1) ** sigma)
-    t = 1 - t
     return t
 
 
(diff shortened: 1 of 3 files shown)


duongve13112002:
@kohya-ss, Would you mind reviewing this and merging it into sd3? This issue could cause LoRA training or full fine-tuning to not work correctly.

kohya-ss:
@duongve13112002 Sorry for the late confirmation. Is it correct to understand that this replaces #2215?

duongve13112002:
@kohya-ss This implementation by @urlesistiana is more detailed and complete than the previous one I fixed. I’ve tested it and it works. I’m still fixing issues related to the full fine-tune of Lumina, and I’ll open a pull request once it’s done.

PR #2240, opened 2025-11-17
Use `torch.svd_lowrank` for large matrices in `resize_lora.py`

woct0rdho:
This hugely reduces the time to prune a LoRA, without visible quality loss compared to the full SVD. The time is reduced from 3 hours (Wan) and 1 hour (Qwen-Image) to a few seconds on my laptop.

See https://docs.pytorch.org/docs/stable/generated/torch.svd_lowrank.html for details. Maybe we can increase `niter` for even better quality but I think the default `niter=2` already gives good enough quality. Kijai's `LoraExtractKJ` node uses `niter=7` by default, see https://github.com/kijai/ComfyUI-KJNodes/blob/bb205d809b467307b8ec3bb1a22680a4873187f8/nodes/lora_nodes.py#L30 (but they don't yet do adaptive rank after svd_lowrank)

### DIFF
--- a/networks/resize_lora.py
+++ b/networks/resize_lora.py
@@ -87,7 +87,14 @@ def index_sv_ratio(S, target):
 # Modified from Kohaku-blueleaf's extract/merge functions
 def extract_conv(weight, lora_rank, dynamic_method, dynamic_param, device, scale=1):
     out_size, in_size, kernel_size, _ = weight.size()
-    U, S, Vh = torch.linalg.svd(weight.reshape(out_size, -1).to(device))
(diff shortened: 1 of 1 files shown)


kohya-ss:
Thank you, this is really nice! I'd like to merge this sooner. I'm a little busy so please wait a moment.

kohya-ss:
Thank you for this!

kohya-ss:
I've added an option `--svd_lowrank_niter` to control `niter` param and disable `svd_lowrank` at #2296.

PR #2255, opened 2026-01-19
Fix AttributeError for UNet2DConditionModel with newer diffusers versions

cgcalatrava:
Recent versions of diffusers moved UNet2DConditionModel to:
diffusers.models.unets.unet_2d_condition

This change fixes the following error that occurs when using newer diffusers versions
(e.g. diffusers[torch]==0.32.1):

AttributeError: module 'diffusers.models' has no attribute 'unet_2d_condition'

It also adds a backward-compatible import fallback so that sdxl_gen_img.py
works with both older and newer diffusers versions.

Tested with diffusers 0.24.x and 0.32.x.

### DIFF
--- a/sdxl_gen_img.py
+++ b/sdxl_gen_img.py
@@ -15,6 +15,12 @@ import random
 import re
 
 import diffusers
+
+# Compatible import for diffusers old/new UNet path
+try:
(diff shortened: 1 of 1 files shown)


kohya-ss:
Thank you for this! We recommend `gen_img.py` for SDXL image generation, but it's important that our older scripts continue to work.

PR #2262, opened 2026-02-12
Fix bug and optimization for Lumina model

duongve13112002:
Hi @kohya-ss 

Here is a summary of the changes I made to the Lumina model in sd-scripts. These updates are based on my experiments while training both LoRA and full models.

-  Updated the code in library/lumina_models.py to optimize training speed.
-  Fixed several bugs related to confusion between the Flux and Lumina models during implementation.

If you have any question feel free to ask.

### DIFF
--- a/library/lumina_models.py
+++ b/library/lumina_models.py
@@ -34,18 +34,18 @@ from library import custom_offloading_utils
 try:
     from flash_attn import flash_attn_varlen_func
     from flash_attn.bert_padding import index_first_axis, pad_input, unpad_input  # noqa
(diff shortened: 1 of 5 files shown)


kohya-ss:
Thank you for this! I will review this tomorrow.

kohya-ss:
Thank you. Sorry for the late review.

I'm not familiar with Lumina's architecture, but the training seems to be working without any problems.

kohya-ss:
I've fixed `is not` error in #2267.

PR #2272, opened 2026-02-18
Fix the LoRA dropout issue in the Anima model and implement support for network_reg_alphas during LoRA training.

duongve13112002:
Hi @kohya-ss 

- During testing, I discovered a bug where training immediately crashes when using rank_dropout.
The issue is related to the Flux LoRA implementation.

- This pull request fixes the rank_dropout issue and adds support for network_reg_alphas, allowing different alpha values to be set per module during LoRA training for the Anima model.

### DIFF
--- a/docs/anima_train_network.md
+++ b/docs/anima_train_network.md
@@ -60,11 +60,7 @@ Qwen-Image VAEとQwen-Image VAEは同じアーキテクチャですが、[Anima�
 * **引数:** DiTモデルのパスには共通引数`--pretrained_model_name_or_path`を、Qwen3テキストエンコーダーには`--qwen3`を、Qwen-Image VAEには`--vae`を使用します。LLM AdapterとT5トークナイザーはそれぞれ`--llm_adapter_path`、`--t5_tokenizer_path`で個別に指定できます。
 * **一部引数の非互換性:** Stable Diffusion v1/v2向けの引数（例: `--v2`, `--v_parameterization`, `--clip_skip`）は使用されません。`--fp8_base`はサポートされていません。
 * **タイムステップサンプリング:** FLUX学習と同じ`--timestep_sampling`オプション（`sigma`、`uniform`、`sigmoid`、`shift`、`flux_shift`）を使用します。
-<<<<<<< HEAD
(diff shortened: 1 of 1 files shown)


kohya-ss:
Thank you for the pull request, and sorry for the delayed review.

Regarding `network_reg_alphas`, this is an interesting idea, but I believe the same effect can be achieved by adjusting the learning rate per module. Adding this new option would expand the hyperparameter search space without a clear benefit, and could lead to confusion for users.

Would it be possible to remove the `network_reg_alphas` feature from this PR and keep only the `rank_dropout` fix?

Thank you for your understanding.

duongve13112002:
Yeah, I removed it. However, I think it would be great if we could customize network_reg_alphas for training different ranks for each LoRA component.

kohya-ss:
Thank you for the quick update!

Just to share the reasoning behind my decision: The effective scaling of LoRA in each module is:

$$\text{effective scale} = \text{lr}_i \times \frac{\alpha_i}{r_i}$$

Since we can already control both lr and r per module, adding per-module alpha does not provide a fundamentally new axis of control — changing alpha can always be replicated by adjusting lr in the opposite direction.

In my understanding, the original motivation for introducing alpha in the LoRA paper was as a stabilization mechanism to eliminate the need for hyperparameter re-tuning when changing rank. With a fixed alpha, you can change rank and keep the rest of your hyperparameters as-is. So I think alpha was not designed as a parameter to be actively tuned per module.

duongve13112002:
Oh, I think the code is clear now. Would you mind reviewing it? I agree with you about the alpha rank. If you have any question feel free to ask

kohya-ss:
Thank you for update again!

The intention of this process is to match the mask and lx shapes, so the following process may be sufficient.
```
            # rank dropout
            if self.rank_dropout is not None and self.training:
                # Simplify to avoid unsqueeze operations
                mask_shape = [lx.size(0)] + [1] * (len(lx.size()) - 2) + [self.lora_dim]
                mask = torch.rand(mask_shape, device=lx.device) > self.rank_dropout
                lx = lx * mask
```

duongve13112002:
I'm not sure. In my opinion, we should separate it, because if lora_down is a Conv2d layer, the training may be break.

kohya-ss:
Oh, I see. Sorry, I was wrong. You're right. Conv2d is channel first.

PR #2276, opened 2026-02-23
feat: backward compatibility for SD/SDXL latent cache

kohya-ss:
Prior to #2269, no error was raised if a latent without a resolution suffix differed from the expected latent shape (which occurs when the user defines multiple resolution datasets in the same cache directory). This PR implements that behavior for backward compatibility.

### DIFF
--- a/library/strategy_sd.py
+++ b/library/strategy_sd.py
@@ -145,7 +145,7 @@ class SdSdxlLatentsCachingStrategy(LatentsCachingStrategy):
         self.suffix = (
             SdSdxlLatentsCachingStrategy.SD_LATENTS_NPZ_SUFFIX if sd else SdSdxlLatentsCachingStrategy.SDXL_LATENTS_NPZ_SUFFIX
         )
-    
+
     @property
     def cache_suffix(self) -> str:
         return self.suffix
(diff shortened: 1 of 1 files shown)


woct0rdho:
In #2269, if the user sets a resolution (for example 512) and the cached latent without resolution suffix is different (for example 768), then `LatentsCachingStrategy._default_is_disk_cached_latents_expected` will return False, and the training script will compute the new latent and add it to the npz file with suffix 512/8.

I think this is a good default behavior and we don't need to show error or warning in this case. Just like the case if the user sets resolution 512 and the cached latent has resolution suffix 768/8 , then we compute the new latent and add it to the npz file with suffix 512/8, without error or warning.

The old latent will not be removed, and it will be used when the user wants to train again with resolution 768. If the user feels out of disk space, they can manually remove the npz files.

After this PR, the training script shows a warning and loads the latent 768 (not 512). Is it what you intend to do?

kohya-ss:
Thank you, you have a point. However, I would like to avoid behavior changes when applying a PR.

Before #2269 was applied, if the user set a resolution (e.g., 512) and the cached latent (without the resolution suffix) were different (e.g., 768), `LatentsCachingStrategy._default_is_disk_cached_latents_expected` would return True and the training script would use variables different from the resolution. This PR reproduces that.

However, this is probably not what you want, so we're displaying a warning.

In any case, this use of a different size latent only occurs in the rare case of a mismatch between the cached latent size and the latent size derived from the current dataset, so it seems simpler to just reproduce the previous behavior rather than checking the size using npz's internal API.

woct0rdho:
I see, now this PR looks good to me.

kohya-ss:
Thank you for your understanding and for your review! And many thanks again for submitting the original PR.

PR #2277, opened 2026-02-23
feat: Stability with fp16 for anima

kohya-ss:
This pull request introduces an improvement to the `_forward` method in `library/anima_models.py` to enhance numerical stability during computations involving residual connections.

Numerical stability improvements:

* In the `_forward` method, input tensor `x_B_T_H_W_D` is now cast to `float32` if its original type is `float16`, ensuring better numerical stability in residual connections. Each module will handle casting back to `float16` as needed using an autocast context.

---

Address the issue described in #2274.

### DIFF
--- a/library/anima_models.py
+++ b/library/anima_models.py
@@ -864,6 +864,10 @@ class Block(nn.Module):
         adaln_lora_B_T_3D: Optional[torch.Tensor] = None,
         extra_per_block_pos_emb: Optional[torch.Tensor] = None,
     ) -> torch.Tensor:
+        if x_B_T_H_W_D.dtype == torch.float16:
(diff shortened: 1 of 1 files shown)


RedRayz:
Even after this PR is merged, it still causing NaN losses in fp16 precision.

kohya-ss:
> Even after this PR is merged, it still causing NaN losses in fp16 precision.

Thank you for reporting. Does #2274 solve the issue?

RedRayz:
Yes, #2274 works fine without any problems.

kohya-ss:
Thank you, I will investigate. If I can't find the cause, I will consider adopting #2274 as is.

kohya-ss:
I've opened #2302, I hope this fixes the issue.

PR #2315, opened 2026-04-17
fix: replace deprecated CLIPFeatureExtractor with CLIPImageProcessor for transformers 5.x compatibility

marcus165090-spec:
## Problem

`CLIPFeatureExtractor` was deprecated in transformers 4.x (renamed to `CLIPImageProcessor`) and **fully removed in transformers 5.0**. Users running sd-scripts with `transformers>=5.0.0` get the following error:

```
ImportError: cannot import name 'CLIPFeatureExtractor' from 'transformers'
```

This affects both `lpw_stable_diffusion.py` and `sdxl_lpw_stable_diffusion.py`.

## Fix

Replace the import with `CLIPImageProcessor` aliased as `CLIPFeatureExtractor` to maintain full backward compatibility while supporting transformers 5.x:

```python
# Before
from transformers import CLIPFeatureExtractor, CLIPTextModel, CLIPTokenizer, ...

# After
from transformers import CLIPImageProcessor as CLIPFeatureExtractor, CLIPTextModel, CLIPTokenizer, ...
```

## Files changed

- `library/lpw_stable_diffusion.py`
- `library/sdxl_lpw_stable_diffusion.py`

## Testing

Verified working with:
- Python 3.13.11
- PyTorch 2.10.0+cu130
- transformers 5.3.0
- NVIDIA GeForce RTX 2060

### DIFF
--- a/library/lpw_stable_diffusion.py
+++ b/library/lpw_stable_diffusion.py
@@ -9,7 +9,7 @@ import numpy as np
 import PIL.Image
 import torch
 from packaging import version
-from transformers import CLIPFeatureExtractor, CLIPTextModel, CLIPTokenizer, CLIPVisionModelWithProjection
+from transformers import CLIPImageProcessor as CLIPFeatureExtractor, CLIPTextModel, CLIPTokenizer, CLIPVisionModelWithProjection
 
 import diffusers
 from diffusers import SchedulerMixin, StableDiffusionPipeline
--- a/library/sdxl_lpw_stable_diffusion.py
+++ b/library/sdxl_lpw_stable_diffusion.py
@@ -10,7 +10,7 @@ import PIL.Image
 import torch
 from packaging import version
 from tqdm import tqdm
-from transformers import CLIPFeatureExtractor, CLIPTextModel, CLIPTokenizer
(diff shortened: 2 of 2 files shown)


WhitePr:
This change isn’t enough for Anima; you’ll also need to modify the following code:
https://github.com/kohya-ss/sd-scripts/blob/main/library/strategy_anima.py#57
```
qwen3_encoding = self.qwen3_tokenizer.batch_encode_plus(
            text, return_tensors="pt", truncation=True, padding="max_length", max_length=self.qwen3_max_length
        )
```
https://github.com/kohya-ss/sd-scripts/blob/main/library/strategy_anima.py#64
```
t5_encoding = self.t5_tokenizer.batch_encode_plus(
            text, return_tensors="pt", truncation=True, padding="max_length", max_length=self.t5_max_length
        )
```
Change 'batch_encode_plus' to '_encode_plus'

I haven’t tested whether other training scripts work properly; scripts that use Qwen as the TE may also need to be modified.

kohya-ss:
Thank you for this PR!

As WhitePr mentioned, it seems that several additional modifications are needed to support transformers 5, so I will address this after the merge.

PR #2352, opened 2026-05-10
Improve resolution bucketing

### LINKED ISSUE #731: SDXL 1024 training - images assigned to wrong buckets
I'm training a SDXL Lora and I don't understand why some of my images end up in the 960x960 bucket. Shouldn't the square and square like images go to the 1024x1024 bucket, provided the img resolution is high enough? This might be a problem with the script or perhaps I'm misunderstanding how images are assigned to buckets.... Can anyone shed some light on this?

I've got 27 high-res images (see below). As you can see there is only a single image with width < 1024, but it's 988x1756 so it should go to one of the tall portrait buckets, right?

257044676-cd43610a-579e-4a02-a174-9742cddc3bcb

And this is how the buckets are reported (img repeat is set to 5), so 2 images are assigned to the 960x960 bucket:

> bucket 0: resolution (768, 1216), count: 5
> bucket 1: resolution (768, 1344), count: 5
> bucket 2: resolution (832, 1088), count: 5
> bucket 3: resolution (896, 1024), count: 10
> bucket 4: resolution (896, 1088), count: 5
> **bucket 5: resolution (960, 960), count: 10**
> bucket 6: resolution (960, 1024), count: 20
> bucket 7: resolution (960, 1088), count: 10
> bucket 8: resolution (1024, 896), count: 5
> bucket 9: resolution (1024, 960), count: 20
> bucket 10: resolution (1088, 832), count: 5
> bucket 11: resolution (1152, 832), count: 10
> bucket 12: resolution (1152, 896), count: 10
> bucket 13: resolution (1216, 832), count: 5
> bucket 14: resolution (1280, 768), count: 5
> bucket 15: resolution (1344, 768), count: 5

Below is my training command. The training resolution is 1024x1024, buckets are enabled, bucket upscale is disabled, bucket resolution is 64.

> accelerate launch --num_cpu_threads_per_process=2 "./sdxl_train_network.py" **--enable_bucket**
> --min_bucket_reso=512 --max_bucket_reso=2048
> --pretrained_model_name_or_path="E:/Automatic1111/stable-diffusion-webui/models/Stable-diffusio
> n/sd/sdXL_v10VAEFix.safetensors" --train_data_dir="E:/Automatic1111/datasets/mona/train_v2"
> **--resolution="1024,1024"** --output_dir="E:/Automatic1111/datasets/mona/output"
> --logging_dir="E:/Automatic1111/datasets/mona/logs" --network_alpha="1"
> --training_comment="mona" --save_model_as=safetensors --network_module=networks.lora
> --text_encoder_lr=0.0003 --unet_lr=0.0003 --network_dim=32 --output_name="sdxl_mona3"
> --lr_scheduler_num_cycles="10" --cache_text_encoder_outputs --no_half_vae --full_bf16
> --learning_rate="0.00035" --lr_scheduler="constant" --train_batch_size="1"
> --max_train_steps="1350" --save_every_n_epochs="1" --mixed_precision="bf16"
> --save_precision="bf16" --caption_extension="txt" --cache_latents --cache_latents_to_disk
> --optimizer_type="Adafactor" --optimizer_args scale_parameter=False relative_step=False
> warmup_init=False --max_data_loader_n_workers="0" **--bucket_reso_steps=64** --flip_aug --xformers
> **--bucket_no_upscale** --noise_offset=0.0357 --network_train_unet_only --sample_sampler=k_dpm_2
> --sample_prompts="E:/Automatic1111/datasets/mona/output\sample\prompt.txt"
> --sample_every_n_epochs="1"

I'm using kohya_gui on windows. I have already posted this issue a week ago [there](https://github.com/bmaltais/kohya_ss/issues/1299), but I didn't receive any answer, except someone else reporting similar issues. I believe kohya gui was synchronized with kohya scripts two days ago and the issue is still there.

Also, when testing on a larger dataset with 160+ images, and the same bucket/resolution settings, I get proper assignment and the square bucket is 1024x1024. I have no idea what makes this smaller dataset create such small buckets. Other than the buckets being wrong, everything else seems to be working fine but of course I'm worried that I'm training in suboptimal resolution.

Hirmuolio:
Resolution bucketing often chooses buckets that are not as good as they could be when no_upscale=true. It also prioritizes matching aspect ratio at the expense of resolution.

This PR gives it a bit leeway in choosing slightly mismatching aspect ratio in exchange of higher resolution.

`select_bucket()` with `bucket_no_upscale=true`

* It used to find bucket resolution with aspect ratio closest to original resolution without considering anything else. The math it used also sometimes resulted in smaller bucket than necessary. 
* Now it chooses highest resolution bucket that crops at most a predefined amount. If no such bucket exists then bucket with closest aspect ratio is chosen.
* TODO:Expose acceptable cropping treshold so it can be configured. Choose a default value.

Fixes: #731

---
Example 1:
Before 3072x3000 would be resized to 960x960. Now it resizes to 1024x1024

Examples 2:

With `bucket_no_upscale=false` there are now fewer buckets and the images are sometimes put into higher resolution bucket.

Before (576x576):
INFO     bucket 0: resolution (320, 896), count: 8  
INFO     bucket 1: resolution (320, 960), count: 2  
INFO     bucket 2: resolution (320, 1024), count: 3 
INFO     bucket 3: resolution (384, 768), count: 30 
INFO     bucket 4: resolution (384, 832), count: 14 
INFO     bucket 5: resolution (448, 704), count: 158
INFO     bucket 6: resolution (512, 640), count: 106
INFO     bucket 7: resolution (576, 576), count: 39 
INFO     bucket 8: resolution (640, 512), count: 25 
INFO     bucket 9: resolution (704, 448), count: 17 
INFO     bucket 10: resolution (768, 384), count: 2 
INFO     bucket 11: resolution (896, 320), count: 1 
INFO     mean ar error (without repeats): 0.046860775345269015

After (576x576):
INFO     bucket 0: resolution (320, 896), count: 8  
INFO     bucket 1: resolution (320, 960), count: 5  
INFO     bucket 2: resolution (384, 768), count: 30 
INFO     bucket 3: resolution (384, 832), count: 14 
INFO     bucket 4: resolution (448, 640), count: 151
INFO     bucket 5: resolution (448, 704), count: 56 
INFO     bucket 6: resolution (512, 576), count: 10 
INFO     bucket 7: resolution (512, 640), count: 51 
INFO     bucket 8: resolution (576, 512), count: 5  
INFO     bucket 9: resolution (576, 576), count: 33 
INFO     bucket 10: resolution (640, 448), count: 18
INFO     bucket 11: resolution (640, 512), count: 13
INFO     bucket 12: resolution (704, 448), count: 8 
INFO     bucket 13: resolution (768, 384), count: 2 
INFO     bucket 14: resolution (896, 320), count: 1 
INFO     mean ar error (without repeats): 0.02551882110922673

Example 3:

Before (1024x1024):
INFO     bucket 0: resolution (640, 1536), count: 1       
INFO     bucket 1: resolution (704, 1216), count: 1       
INFO     bucket 2: resolution (704, 1408), count: 2       
INFO     bucket 3: resolution (768, 1216), count: 1       
INFO     bucket 4: resolution (768, 1280), count: 2       
INFO     bucket 5: resolution (768, 1344), count: 3       
INFO     bucket 6: resolution (832, 1088), count: 3       
INFO     bucket 7: resolution (832, 1152), count: 4       
INFO     bucket 8: resolution (832, 1216), count: 5       
INFO     bucket 9: resolution (896, 1024), count: 16      
INFO     bucket 10: resolution (896, 1088), count: 10     
INFO     bucket 11: resolution (896, 1152), count: 9      
INFO     bucket 12: resolution (960, 960), count: 6       
INFO     bucket 13: resolution (960, 1024), count: 24     
INFO     bucket 14: resolution (960, 1088), count: 9      
INFO     bucket 15: resolution (1024, 768), count: 18     
INFO     bucket 16: resolution (1024, 896), count: 4      
INFO     bucket 17: resolution (1024, 960), count: 8      
INFO     bucket 18: resolution (1024, 1024), count: 25    
INFO     bucket 19: resolution (1088, 896), count: 4      
INFO     bucket 20: resolution (1152, 896), count: 29     
INFO     bucket 21: resolution (1216, 832), count: 12     
INFO     bucket 22: resolution (1344, 768), count: 10     
INFO     mean ar error (without repeats): 0.007779885784017744

After (1024x1024):
INFO     bucket 0: resolution (640, 1600), count: 1 
INFO     bucket 1: resolution (704, 1472), count: 2 
INFO     bucket 2: resolution (768, 1344), count: 7 
INFO     bucket 3: resolution (832, 1216), count: 5 
INFO     bucket 4: resolution (896, 1152), count: 16
INFO     bucket 5: resolution (960, 1088), count: 33
INFO     bucket 6: resolution (1024, 768), count: 18
INFO     bucket 7: resolution (1024, 1024), count: 65 
INFO     bucket 8: resolution (1088, 960), count: 8 
INFO     bucket 9: resolution (1152, 896), count: 29
INFO     bucket 10: resolution (1216, 832), count: 12 
INFO     bucket 11: resolution (1344, 768), count: 10 
INFO     mean ar error (without repeats): 0.021142901962640452

### DIFF
--- a/finetune/prepare_buckets_latents.py
+++ b/finetune/prepare_buckets_latents.py
@@ -99,12 +99,7 @@ def main(args):
     bucket_manager = train_util.BucketManager(
         args.bucket_no_upscale, max_reso, args.min_bucket_reso, args.max_bucket_reso, args.bucket_reso_steps
     )
-    if not args.bucket_no_upscale:
-        bucket_manager.make_buckets()
-    else:
-        logger.warning(
-            "min_bucket_reso and max_bucket_reso are ignored if bucket_no_upscale is set, because bucket reso is defined by image size automatically / bucket_no_upscaleが指定された場合は、bucketの解像度は画像サイズから自動計算されるため、min_bucket_resoとmax_bucket_resoは無視されます"
-        )
+    bucket_manager.make_buckets()
 
     # 画像をひとつずつ適切なbucketに割り当てながらlatentを計算する
     img_ar_errors = []
@@ -229,8 +224,8 @@ def setup_parser() -> argparse.ArgumentParser:
         default="512,512",
         help="max resolution in fine tuning (width,height) / fine tuning時の最大画像サイズ 「幅,高さ」（使用メモリ量に関係します）",
     )
-    parser.add_argument("--min_bucket_reso", type=int, default=256, help="minimum resolution for buckets / bucketの最小解像度")
-    parser.add_argument("--max_bucket_reso", type=int, default=1024, help="maximum resolution for buckets / bucketの最大解像度")
+    parser.add_argument("--min_bucket_reso", type=int, default=256, help="minimum side length for buckets / bucketの最小解像度")
+    parser.add_argument("--max_bucket_reso", type=int, default=3072, help="maximum side length for buckets / bucketの最大解像度")
     parser.add_argument(
         "--bucket_reso_steps",
         type=int,
--- a/finetune/prepare_buckets_latents.py
+++ b/finetune/prepare_buckets_latents.py
@@ -99,7 +99,12 @@ def main(args):
     bucket_manager = train_util.BucketManager(
         args.bucket_no_upscale, max_reso, args.min_bucket_reso, args.max_bucket_reso, args.bucket_reso_steps
     )
-    bucket_manager.make_buckets()
+    if not args.bucket_no_upscale:
+        bucket_manager.make_buckets()
+    else:
+        logger.warning(
+            "min_bucket_reso and max_bucket_reso are ignored if bucket_no_upscale is set, because bucket reso is defined by image size automatically / bucket_no_upscaleが指定された場合は、bucketの解像度は画像サイズから自動計算されるため、min_bucket_resoとmax_bucket_resoは無視されます"
+        )
 
     # 画像をひとつずつ適切なbucketに割り当てながらlatentを計算する
     img_ar_errors = []
@@ -224,8 +229,8 @@ def setup_parser() -> argparse.ArgumentParser:
         default="512,512",
         help="max resolution in fine tuning (width,height) / fine tuning時の最大画像サイズ 「幅,高さ」（使用メモリ量に関係します）",
     )
-    parser.add_argument("--min_bucket_reso", type=int, default=256, help="minimum side length for buckets / bucketの最小解像度")
-    parser.add_argument("--max_bucket_reso", type=int, default=3072, help="maximum side length for buckets / bucketの最大解像度")
+    parser.add_argument("--min_bucket_reso", type=int, default=256, help="minimum resolution for buckets / bucketの最小解像度")
+    parser.add_argument("--max_bucket_reso", type=int, default=1024, help="maximum resolution for buckets / bucketの最大解像度")
     parser.add_argument(
         "--bucket_reso_steps",
         type=int,
--- a/library/model_util.py
+++ b/library/model_util.py
@@ -1385,31 +1385,32 @@ def expand_unet_to_inpainting(unet) -> None:
     unet.in_channels = 9
 
 
-def make_bucket_resolutions(max_reso, min_size=256, max_size=1024, divisible=64):
+def make_bucket_resolutions(max_reso: tuple[int, int], min_size: int = 256, max_size: int = 4096, divisible: int = 64, no_upscale: bool = False ) -> set:
+
+    assert min_size    % divisible == 0, "Resolutions must be divisible by step"
+    assert max_size    % divisible == 0, "Resolutions must be divisible by step"
+    assert max_reso[0] % divisible == 0, "Resolutions must be divisible by step"
+    assert max_reso[1] % divisible == 0, "Resolutions must be divisible by step"
+
     max_width, max_height = max_reso
     max_area = max_width * max_height
-
     resos = set()
 
-    width = int(math.sqrt(max_area) // divisible) * divisible
-    resos.add((width, width))
+    if no_upscale:
+        # All resolutions smaller or equal to max_reso are accepted
+        size_diff = 0.0
+    else:
+        # Resolutions that are at most 20% smaller than max_reso are accepted
+        size_diff = 0.8
 
-    width = min_size
-    while width <= max_size:
-        height = min(max_size, int((max_area // width) // divisible) * divisible)
-        if height >= min_size:
-            resos.add((width, height))
-            resos.add((height, width))
-
-        # # make additional resos
-        # if width >= height and width - divisible >= min_size:
-        #   resos.add((width - divisible, height))
-        #   resos.add((height, width - divisible))
-        # if height >= width and height - divisible >= min_size:
-        #   resos.add((width, height - divisible))
-        #   resos.add((height - divisible, width))
-
-        width += divisible
+    # Get resolutions that are at most 20% smaller than max_reso
+    for width in range( min_size, max_size, divisible ):
+        for height in range( min_size, max_size, divisible ):
+            if width * height > max_area:
+                break
+            if width*height / max_area > size_diff:
+                resos.add((width, height))
+                resos.add((height, width))
 
     resos = list(resos)
     resos.sort()
--- a/library/model_util.py
+++ b/library/model_util.py
@@ -1385,31 +1385,31 @@ def expand_unet_to_inpainting(unet) -> None:
     unet.in_channels = 9
 
 
-def make_bucket_resolutions(max_reso: tuple[int, int], min_size: int = 256, max_size: int = 1024, divisible: int = 64, no_upscale: bool = False ) -> list:
-
-    assert min_size    % divisible == 0, "Resolutions must be divisible by step"
-    assert max_size    % divisible == 0, "Resolutions must be divisible by step"
-    assert max_reso[0] % divisible == 0, "Resolutions must be divisible by step"
-    assert max_reso[1] % divisible == 0, "Resolutions must be divisible by step"
-
+def make_bucket_resolutions(max_reso, min_size=256, max_size=1024, divisible=64):
     max_width, max_height = max_reso
     max_area = max_width * max_height
+
     resos = set()
 
-    if no_upscale:
-        # All resolutions smaller or equal to max_reso are accepted
-        size_diff = 0.0
-    else:
-        # Resolutions that are at most 20% smaller than max_reso are accepted
-        size_diff = 0.8
+    width = int(math.sqrt(max_area) // divisible) * divisible
+    resos.add((width, width))
 
-    for width in range( min_size, max_size, divisible ):
-        for height in range( min_size, max_size, divisible ):
-            if width * height > max_area:
-                break
-            if width*height / max_area > size_diff:
-                resos.add((width, height))
-                resos.add((height, width))
+    width = min_size
+    while width <= max_size:
+        height = min(max_size, int((max_area // width) // divisible) * divisible)
+        if height >= min_size:
+            resos.add((width, height))
+            resos.add((height, width))
+
+        # # make additional resos
+        # if width >= height and width - divisible >= min_size:
+        #   resos.add((width - divisible, height))
+        #   resos.add((height, width - divisible))
+        # if height >= width and height - divisible >= min_size:
+        #   resos.add((width, height - divisible))
+        #   resos.add((height - divisible, width))
+
+        width += divisible
 
     resos = list(resos)
     resos.sort()
--- a/library/model_util.py
+++ b/library/model_util.py
@@ -1385,7 +1385,7 @@ def expand_unet_to_inpainting(unet) -> None:
     unet.in_channels = 9
 
 
-def make_bucket_resolutions(max_reso: tuple[int, int], min_size: int = 256, max_size: int = 4096, divisible: int = 64, no_upscale: bool = False ) -> set:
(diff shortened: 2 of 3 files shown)


kohya-ss:
Thanks for the PR and the detailed examples — the analysis itself is genuinely useful, so I'd like to start with a direction-level discussion before going into specifics.

### The underlying trade-off

Bucket selection is fundamentally a trade-off between aspect-ratio matching and effective per-image resolution. Allowing more bucket choices (e.g. the `size_diff = 0.8` relaxation in this PR) gives better AR matching, but pushes some images into smaller-area buckets, lowering their effective training resolution. There is no objectively optimal point on this curve — it depends on the dataset and the user's priorities.

The current implementation picks one defensible point on that curve. It is not optimal for every case, but the alternative position chosen by this PR is also a judgment call, not a strict improvement.

### Existing levers for the "small bucket area" issue

In my understanding, the underlying issue this PR addresses is not extreme aspect ratios per se, but the fact that some bucket sizes can end up small in area — for example, `(256, 1024)` at max_reso=(1024, 1024) only uses 1/4 of the max area. This is already addressable from the dataset config side: `--min_bucket_reso` is the lower bound for **both** axes, so raising  it directly raises the minimum bucket area. At max_reso=(1024, 1024):

| `min_bucket_reso` | Smallest bucket | Area (vs max_area) |
| --- | --- | --- |
| 256 (default) | (256, 1024) | ~25% |
| 384 | (384, 1024) | ~37% |
| 512 | (512, 1024) | ~50% |

For modern architectures with smaller `bucket_reso_steps` (32 or 16) and larger working resolutions, this lever stays effective because the surviving bucket grid is still dense. So users who want to avoid small-area buckets already have a knob — better documentation might help more users than a default behavior change.

### `bucket_no_upscale` semantics

The original intent of `bucket_no_upscale` is twofold:

1. Do not upscale images smaller than the specified resolution.
2. For images larger than the specified resolution, the bucket is decided purely from the image's aspect ratio — width
 and height are each truncated to `bucket_reso_steps` multiples, **without** any `min_bucket_reso` / `max_bucket_reso` constraint. This is by design and means that very wide or tall images can land in extreme buckets such as `(128, 1024)`; the trade-off is that the original image is preserved as much as possible.
  
Under this PR, `bucket_no_upscale=True` instead forces every image into a predefined grid built from `min_bucket_reso` / `max_bucket_reso`. This is a meaningful semantic change for an existing flag, and I'd prefer not to alter it.

### Cache compatibility

This is the part I'm most concerned about. Any change to bucketing — including the default change of `max_bucket_reso`  from 1024 to 3072 — silently invalidates existing latent caches: the same image can be assigned to a different bucket  and therefore expects a different cached latent. For users with large pre-cached datasets (a common workflow), this risks either silent re-caching of everything or, worse, training on stale latents if the cache key happens to collide.  This makes any default-behavior change to bucketing in `main` very risky for downstream users.


### Where I land

Given the above, I don't think the `size_diff` approach is the right direction for this codebase, even with the implementation issues fixed. The current bucketing isn't optimal, but the position it picks on the AR-vs-area trade-off is defensible, and the existing parameters already let users tune that trade-off for their dataset. On top of that, any change that alters bucket assignments by default has serious consequences for existing latent caches, which makes the bar for default-behavior changes very high.

So I'm going to close this PR. That said, the underlying observation — that the current grid can be sparse in AR near the diagonal, especially for architectures with smaller `bucket_reso_steps` — is fair, and I wouldn't rule out a future change in this area if it (a) adds bucket density locally without redefining the existing grid, (b) keeps `bucket_no_upscale` semantics intact, and (c) is opt-in so existing caches stay valid. If you'd like to explore something along those lines, I'm happy to discuss.

Thanks again for the thoughtful analysis and the before/after tables — they're a useful reference regardless of the outcome here.

Hirmuolio:
> In my understanding, the underlying issue this PR addresses is not extreme aspect ratios per se, but the fact that some bucket sizes can end up small in area — for example, (256, 1024) at max_reso=(1024, 1024) only uses 1/4 of the max area.

No that is not the problem adressed by this PR.

The problem is that the current bucketing often results in incorrect, objectively bad buckets.

For example when training at 1024x1024 an image with 3072x3000 with old code goes into 960x960 bucket.  
With my code it correctly goes into 1024x1024 bucket.

Both of them have same aspect ratio. But the old code incorrectly resizes it into smaller size.

The code also introduces control for _how much croping is allowed_. Unlike the old code which was wholely uncontrolled.  
Simply setting `threshold: float = 1` will make the behavior same as before so it always chooses best aspect ratio and crops as little as possible.

Hence that number could be exposed so end user can change it to their liking.

> Under this PR, bucket_no_upscale=True instead forces every image into a predefined grid built from min_bucket_reso / max_bucket_reso. This is a meaningful semantic change for an existing flag, and I'd prefer not to alter it.

This is true. And the PR _could_ be made without this change.

> Cache compatibility

See point 1. 

A 3072x3000 image should go into 1024x1024 bucket, not into 960x960 bucket. And changing that means that cache compatibility must be broken to fix this.

kohya-ss:
Thanks for the pushback — I traced through your `3072x3000` example to make sure I understood what the code actually does today.

With the default (`bucket_no_upscale=False`), the image lands at `(1024, 1024)` as expected: `(960, 960)` is not even present in the predefined bucket set generated by `make_bucket_resolutions` for `max_reso=(1024, 1024)`, and the AR-nearest selection picks `(1024, 1024)` with an error of 0.024.

The `(960, 960)` outcome only appears under `bucket_no_upscale=True`. In that path, the continuous fixed-area resize is `(1036.22, 1011.93)`, the height-anchored candidate is chosen because it has the smaller AR error, and the subsequent step-aligned rounding lands the bucket at `(960, 960)` even though `(1024, 1024)` is within the same `max_area`. So the underlying observation is fair: the rounding in the `no_upscale=True` path can drop a near-square large image one `reso_steps` below the bucket it could otherwise occupy. That part is worth fixing.

What I don't think is needed is the scope of the changes in this PR. A minimal fix lives entirely inside the `no_upscale=True` rounding path — it doesn't require a new `size_diff` threshold, a semantic change to `bucket_no_upscale`, or a default change to `max_bucket_reso`. Most of the surface area in #2352 is independent of the underlying issue, which is why I'd like to keep this PR closed.

That said, I'd welcome a smaller, focused PR (or an issue first) that addresses only the `no_upscale=True` rounding loss. One caveat on timing: any change to bucket assignment invalidates existing latent caches for affected images, and the SD/SDXL cache loader currently tolerates shape mismatches for backward compatibility (see `strategy_base.py` around `is_disk_cached_latents_expected`). Before such a fix can be merged safely on `main`, the cache validity check needs to be tightened on our side to force re-caching on bucket change — the way multi-resolution architectures (FLUX/SD3/Lumina/etc.) already do. That work hasn't happened yet, so a minimal PR for the rounding fix wouldn't merge immediately; I wanted to flag the dependency up front so the timeline isn't a surprise.

Thanks again — the rounding observation is a useful one and I'm glad it surfaced.

Hirmuolio:
I have confined the changes to no_upscale=True part of the code only.

I'd like to add `--permitted_crop` argument so the extra cropping can be controlled. But those argument definitions seem to be scattered in so many places.

I also strongly recommend that the changes I did in `make_bucket_resolutions()` are later done too. The current implementation of it skips valid resolutions and it does not properly define at all where the line of acceptable resolutions is.

kohya-ss:
Thanks for narrowing the scope to the `no_upscale=True` path — that's a good direction, and I confirmed the updated code does land `3072x3000` in `(1024, 1024)` as intended.

That said, I'd like to keep this PR closed, and I want to be transparent about the reasoning so we don't end up iterating further on the same PR:

1. With `threshold = 0.90` hardcoded inside `select_bucket`, the default behavior of`bucket_no_upscale=True` now permits up to ~10% cropping, where the previous behavior preserved aspect ratio exactly. That is a default-semantics change to an existing flag — one of the constraints I asked to keep in the previous round ((b) in my earlier comment). I understand from your note that you wanted to expose this through a `--permitted_crop` argument but found the argument plumbing scattered; without that argument in place, there is no way for users to opt out of the new cropping.

2. On the algorithm itself: a narrower approach — anchor-based candidate generation (snap on each axis at floor and ceil, derive the other axis from the original aspect ratio) combined with an area<=max_area filter and an AR-then-area tie-break — reaches the same correct result for `3072x3000` in O(1), without iterating over the bucket grid. The current implementation works for the cases it covers, but it's more machinery than the problem requires.

3. There are also a couple of correctness issues in the new code:
- `best_res` is referenced as `reso = best_res` after the loops without being initialized.  If `max(image_width, image_height) // reso_steps * reso_steps == reso_steps`, the outer  `range(reso_steps, max_dim, reso_steps)` is empty and this raises `NameError`.
- `range(reso_steps, max_dim, reso_steps)` excludes `max_dim` itself, so when `image_width` or `image_height` is an exact multiple of `reso_steps`, that exact value is never considered as a candidate bucket dimension.

4. Given the items above and the cache-compatibility prerequisite I mentioned last round (`strategy_base.py` shape check for SD/SDXL), I think the most productive path is for me to fold this fix into the cache-loader robustness work on our side, rather than continuing to iterate here. Could I ask you to open a separate issue capturing the underlying observation — that the `bucket_no_upscale=True` path can drop a near-square large image to a smaller bucket than necessary — and we'll use that as the tracking point? I'll plan the bucket fix together with the SD/SDXL cache validity tightening so the change doesn't silently invalidate existing caches.

Thanks again for the iteration and for narrowing the scope. The observation that drove this PR is genuinely useful and I'll make sure it gets addressed; I just don't want to ask you to keep pushing on a PR that I'm going to merge from a different angle anyway.

PR #2360, opened 2026-05-18
Fix overlap GPU encoding with async disk writes for VAE and text encoder caching

duongve13112002:
Hi @kohya-ss 

Problem: after encoding each batch on GPU, the process synchronously called
np.savez / save_latents_to_disk, leaving the GPU idle for the full duration of
the disk write before starting the next encode. With multiple GPUs each writing
N/num_processes files simultaneously, total disk throughput was saturated, so
adding more GPUs did not reduce wall time.

Fix: add a background ThreadPoolExecutor (write_executor) in both
new_cache_latents and new_cache_text_encoder_outputs. Disk writes are submitted
to this pool immediately after encoding, allowing the GPU to begin the next
batch encode concurrently. All writes are flushed before the function returns.

strategy_base.py:
- LatentsCachingStrategy: add set_async_write_executor / wait_for_async_writes;
  split save_latents_to_disk into async dispatch + _save_latents_to_disk_impl;
  tensors are cloned on CPU before handoff to prevent races
- TextEncoderOutputsCachingStrategy: same helpers plus save_outputs_npz()
  which copies numpy arrays and dispatches to the write pool
- Module-level _write_npz helper (avoids lambda/closure issues)

train_util.py (new_cache_latents):
- Create write_executor sized to cpu_count//num_processes capped at batch_size
- Log [Process X/N] device=... will encode M/total images at start
- Warn when num_processes=1 but multiple GPUs are visible

train_util.py (new_cache_text_encoder_outputs):
- Same write_executor pattern as VAE caching

strategy_flux/sd3/sdxl/lumina/hunyuan_image/anima:
- Replace np.savez(...) with self.save_outputs_npz(...)

### DIFF
--- a/library/strategy_anima.py
+++ b/library/strategy_anima.py
@@ -235,7 +235,7 @@ class AnimaTextEncoderOutputsCachingStrategy(TextEncoderOutputsCachingStrategy):
             caption_dropout_rate = torch.tensor(info.caption_dropout_rate, dtype=torch.float32)
 
             if self.cache_to_disk:
-                np.savez(
+                self.save_outputs_npz(
                     info.text_encoder_outputs_npz,
                     prompt_embeds=prompt_embeds_i,
                     attn_mask=attn_mask_i,
--- a/library/strategy_base.py
+++ b/library/strategy_base.py
@@ -20,6 +20,10 @@ import logging
 logger = logging.getLogger(__name__)
 
 
+def _write_npz(npz_path: str, kwargs: dict):
+    np.savez(npz_path, **kwargs)
+
+
 class TokenizeStrategy:
     _strategy = None  # strategy instance: actual strategy class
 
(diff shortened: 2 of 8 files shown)


kohya-ss:
Hi @duongve13112002, thank you for the detailed PR and investigation.

After review, I'd like to decline this one, for a couple of reasons:

1. I'm planning to add `.safetensors` as a cache file format in the near future, and I'd prefer not to expand the `.npz` write path across strategy_base.py and the six per-model strategy files right before that migration. The async I/O design is something I'd rather build into the new format from the start.
2. As the PR description itself notes, with multiple GPUs the disk throughput is already saturated. Overlapping GPU
encoding with disk writes doesn't change the write-bound wall time in that case, and the unbounded `ThreadPoolExecutor` queue risks holding many cloned CPU tensors in memory when the disk is slow.

That said, the diagnostic logging in `new_cache_latents` (per-process image count and the multi-GPU-with-single-process warning) looks genuinely useful and is independent of the async write change. A repo-wide refactor is currently in progress, so once that settles, could I ask you to resubmit just that logging part as a separate PR?

Thanks again for the contribution and the careful write-up.

duongve13112002:
@kohya-ss, Ok when you finish, i will find the best way to handle this problem. With this method, I tested a simple use case with around 50k images using 4× GPUs. With the baseline, the whole process takes about 1 hour and 45 minutes. In my implementation, the caching step alone takes around 35 minutes.

kohya-ss:
Thank you again for the PR and for sharing the benchmark results (1h45m → 35m with 4 GPUs is a very convincing demonstration of the disk I/O bottleneck).

As discussed above, I'm closing this PR for now. The caching code has since been restructured in the recent refactoring, and I'm planning to migrate the cache files to the .safetensors format (#2290 or new one). I'd like to design the async write path into that new format from the start, rather than extending the current .npz path.

Your investigation and numbers will be a valuable reference when I work on that. I'd be happy to have your feedback on the new implementation once it's ready.

PR #2361, opened 2026-05-18
Support multi-caption embedding caching for all models

duongve13112002:
Hi @kohya-ss 

## Overview
Add support for multi-caption embedding caching. When using `--enable_wildcard` with a caption file containing multiple lines (separated by `\n`), each line is encoded as a separate embedding and stored stacked in the NPZ cache. During training, one variant is randomly selected per epoch — acting as caption augmentation at no extra training cost.

## Requirement
This feature requires `--enable_wildcard` to be enabled. Without it, `train_util.py` always picks only the first line of a multi-line caption at training time, making the other cached variants unused.

| | With `--enable_wildcard` | Without `--enable_wildcard` |
|---|---|---|
| **Caching phase** | All lines encoded & cached ✅ | All lines encoded & cached ✅ |
| **Training (cache hit)** | Random variant selected ✅ | Random variant selected ✅ |
| **Training (cache miss / fallback)** | Random line picked at runtime ✅ | Only first line used ❌ |

## How it works
- **Caching**: `cache_batch_outputs()` reads the raw multi-line caption from `info.caption`, splits by `\n`, encodes each line separately, and saves stacked arrays `(N, seq_len, dim)` to NPZ
- **Loading (disk cache)**: `load_outputs_npz()` detects multi-caption via `ndim == 3` and randomly selects one variant
- **Loading (memory cache)**: `__getitem__` selects one variant using `ImageInfo.num_caption_variants`
- **Backward compatible**: existing single-caption NPZ files (`ndim == 2`) are unaffected

## Models supported
SDXL · SD3 · Flux · Lumina · Anima · HunyuanImage

## Files changed
| File | Change |
|------|--------|
| `library/train_util.py` | Add `num_caption_variants` field to `ImageInfo`; add random selection in `__getitem__` |
| `library/strategy_sdxl.py` | Add `_encode_captions()`, refactor `cache_batch_outputs()`, update `load_outputs_npz()` |
| `library/strategy_flux.py` | Same as above |
| `library/strategy_sd3.py` | Same as above |
| `library/strategy_lumina.py` | Same as above |
| `library/strategy_anima.py` | Same as above (`caption_dropout_rate` is a per-image scalar shared across all variants, not indexed during selection) |
| `library/strategy_hunyuan_image.py` | Same as above |
| `tests/test_*.py` | 36 unit tests covering NPZ load/save, random selection, backward compat |

## Notes
- Caching is slightly slower (per-image encoding instead of batch) but training speed is unaffected
- `caption_dropout_rate` in Anima is cached inside NPZ and applied at training time via `drop_cached_text_encoder_outputs()` — unlike other models where dropout is applied at runtime in `train_util.py`

### DIFF
--- a/library/strategy_anima.py
+++ b/library/strategy_anima.py
@@ -201,7 +201,40 @@ class AnimaTextEncoderOutputsCachingStrategy(TextEncoderOutputsCachingStrategy):
         t5_input_ids = data["t5_input_ids"]
         t5_attn_mask = data["t5_attn_mask"]
         caption_dropout_rate = data["caption_dropout_rate"]
+
+        # multi-caption: prompt_embeds is (N, seq_len, hidden_size) with ndim==3
+        # caption_dropout_rate is a scalar shared across variants — do NOT index it
+        if prompt_embeds.ndim == 3:
+            idx = random.randint(0, prompt_embeds.shape[0] - 1)
+            prompt_embeds = prompt_embeds[idx]
+            attn_mask = attn_mask[idx]
+            t5_input_ids = t5_input_ids[idx]
+            t5_attn_mask = t5_attn_mask[idx]
+
         return [prompt_embeds, attn_mask, t5_input_ids, t5_attn_mask, caption_dropout_rate]
+
+    def _encode_captions(
+        self,
+        tokenize_strategy: TokenizeStrategy,
+        models: List[Any],
+        text_encoding_strategy: "AnimaTextEncodingStrategy",
+        captions: List[str],
+    ) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+        tokens_and_masks = tokenize_strategy.tokenize(captions)
+        with torch.no_grad():
+            prompt_embeds, attn_mask, t5_input_ids, t5_attn_mask = text_encoding_strategy.encode_tokens(
+                tokenize_strategy, models, tokens_and_masks
+            )
+
+        if prompt_embeds.dtype == torch.bfloat16:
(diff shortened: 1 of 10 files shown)


kohya-ss:
Hi @duongve13112002, thanks for the detailed PR and the test coverage.

I'd like to decline this one as is, for two main reasons:

1. Silent behavior change on existing datasets. The cache path triggers on "\n" in info.caption automatically, and load_outputs_npz selects a random variant whenever ndim == 3, regardless of `--enable_wildcard`. Users who currently have multi-line caption files with caching enabled get a deterministic "first line only" behavior today; after this PR  they would silently switch to per-epoch random selection. A change of this kind needs to be strictly opt-in via a dedicated flag.
2. `.npz` cache format migration. As noted on #2360, I'm planning to add `.safetensors` as a cache file format in the near future, and I'd prefer not to extend the `.npz` schema (single array → stacked `(N, seq_len, dim)`) across seven strategy files right before that migration.

On the feature itself: I do have a small in-progress experiment (`exp-use-multi-line-captions-as-they-are`) that interprets `\n` in caption files differently, so the meaning of a newline is something I'd like to settle on the design side before adding cache support for it. A cleaner approach for the augmentation use case might be a dedicated separator (e.g. `---===---`) gated behind an explicit option, which avoids overloading `\n` and keeps the behavior opt-in.

Thanks again for the thorough work — happy to revisit once the new cache format and the multi-line semantics have landed.

duongve13112002:
@kohya-ss, Oh, i am happy to hear that. This feature is very helpful, I hope this can be supported in the future.

PR #2368, opened 2026-06-05
Resize and crop conditioning image in the same way as input image

woct0rdho:
When bucketing is enabled, we resize the input image to maybe a bit larger than the bucket size, then crop it to be exactly the bucket size.

But the transformation of conditioning image looks buggy: it's resized to `target_size_hw`, then cropped to `target_size_hw` (which is a no-op). I think we can reuse `trim_and_resize_if_required` so it's transformed in the same way as the input image.

This may be one of the causes why many control nets and image editing models cannot edit the image in a 'pixel perfect' way, but distort the image by a few pixels.


kohya-ss:
Thank you for this PR. There is indeed a problem with the code, and this fix appears to fix it.

Since I am currently refactoring, please allow me some time to consider how to handle this pull request. I will resolve the issue as soon as possible.

kohya-ss:
I updated this to match the refactored codebase and merged. Thank you again for this!

PR #2369, opened 2026-06-06
Support Anzhc's Qwen 2D VAE

woct0rdho:
Qwen-Image VAE is used in Anima and it's notoriously slow. We can convert 3d convs into 2d convs while keeping the encoded latents unchanged within floating point errors. On my machine (RTX 3080) when encoding a 1024x1024 image, it's 2x faster and takes only 1/3 peak VRAM.

See the model https://huggingface.co/Anzhc/Qwen2D-VAE and the ComfyUI inference code https://github.com/Anzhc/anzhc-qwen2d-comfyui

In the training config we can set `vae = '/path/to/Qwen2D_VAE.safetensors'`, and it will automatically detect the 2D VAE architecture.

This is also worth porting to musubi-tuner.

### DIFF
--- a/library/qwen_image_autoencoder_kl.py
+++ b/library/qwen_image_autoencoder_kl.py
@@ -854,6 +854,373 @@ class QwenImageDecoder3d(nn.Module):
         return x
 
 
+class QwenImageResidualBlock2D(nn.Module):
+    def __init__(self, in_dim: int, out_dim: int, dropout: float = 0.0) -> None:
+        super().__init__()
+        self.norm1 = QwenImageRMS_norm(in_dim)
+        self.conv1 = ChunkedConv2d(in_dim, out_dim, 3, padding=1)
+        self.norm2 = QwenImageRMS_norm(out_dim)
+        self.dropout = nn.Dropout(dropout)
+        self.conv2 = ChunkedConv2d(out_dim, out_dim, 3, padding=1)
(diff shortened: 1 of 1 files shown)


kohya-ss:
Thank you for this. This 2D VAE is fantastic!

However, sd-scripts' Qwen VAE already incorporates a memory-saving mechanism, reducing memory usage to 1/5 of the original.

I also believe that maintaining weight compatibility is important for existing users.

Therefore, directly merging this PR may be difficult, but I will use this idea to improve the VAE implementation in sd-scripts.

Thank you again for letting me know this amazing idea!

woct0rdho:
Besides saving memory, the speedup is also real when we need to cache a lot of latents. I can also implement tiling with the 2D VAE if we want to further save memory.

For weight compatibility, I did not change the code to load the 3D VAE, and we can still use `vae = '/path/to/qwen_image_vae.safetensors'` to load the 3D VAE. It uses `is_qwen2d_state_dict` to detect the 2D VAE architecture in the safetensors file.

It's ok to close this PR if you're not planning to merge it.

kohya-ss:
Inspired by this, I created a 2D version of Qwen-Image VAE in #2382. I also added automatic weight conversion, allowing users to seamlessly utilize the 2D VAE.

Thank you again for sharing this idea!

PR #2376, opened 2026-06-12
Add more optimizer dependency support

gesen2egee:
## Summary

- Add optimizer dependency updates for additional optimizer support.
- Pin `pytorch-optimizer` to commit `3d08fa0`.
- Update `adv-optm` to `2.5.2`.
- Preserve Prodigy / DAdapt learning-rate logging while supporting `effective_lr` logging from upstream.


kohya-ss:
Thanks for the PR. Could you clarify why `schedulefree` and `pytorch-optimizer` need to be installed from Git commits instead of PyPI releases? For default  requirements, we generally prefer released version pins for reproducibility and install stability.

Also, `torch-optimi` and `adv-optm` appear to be newly added default dependencies, but this PR does not seem to add direct code paths that require them. Since arbitrary optimizer classes can already be used by installing optional packages separately, could you remove these from `requirements.txt` unless they are required for a default/supported path?

gesen2egee:
Thanks for the feedback. Here is a quick clarification:
schedulefree & pytorch-optimizer (Git commits): Because the latest PyPI releases lack new features—specifically, the Schedule-Free+ variants and several newly added optimizers.

torch-optimi & adv-optm: They provide Stochastic Rounding and Kahan Rounding implementations for common optimizers (which are fully supported and utilized by frameworks like onetrainer). They are essential for preventing gradient underflow in low-precision training

PR #2400, opened 2026-07-04
Add configurable LoRA-Squeeze training support

jlasz:
## Summary

This PR adds opt-in [LoRA-Squeeze](https://arxiv.org/abs/2602.10993) support to `train_network.py`.

LoRA-Squeeze starts LoRA training at a higher rank and compresses it to the configured deployment rank during the same training run. The completed output is a standard LoRA at `network_dim`; no special inference support is required.

Three main training patterns are supported:

- **In-Squeeze:** progressively reduce rank several times during training.
- **Post-Squeeze:** train entirely at a high rank, then compress once at the end.
- **Cont-Squeeze:** compress once before the end and continue training at the target rank.

The feature is disabled unless `lora_squeeze_start_dim` or `lora_squeeze_num_squeezes` is configured.

## Basic In-Squeeze example

```toml
network_dim = 8
network_alpha = 4
max_train_steps = 4000
lora_squeeze_start_dim = 64
lora_squeeze_num_squeezes = 3
```

With the defaults, this creates:

```text
64 -> 32 -> 16 -> 8
```

The 4000 configured training steps are divided equally:

```text
steps    0-1000: rank 64
steps 1000-2000: rank 32
steps 2000-3000: rank 16
steps 3000-4000: rank 8
```

`max_train_steps` remains the total training budget. LoRA-Squeeze redistributes those steps between ranks; it does not add extra training steps.

The default configuration is equivalent to:

```toml
lora_squeeze_rank_schedule = "geometric"
lora_squeeze_step_schedule = "equal"
lora_squeeze_alpha_schedule = "proportional"
lora_squeeze_train_after_final_squeeze = true
lora_squeeze_optimizer_mode = "per_squeeze"
lora_squeeze_scheduler_mode = "global"
lora_squeeze_first_segment_ratio = 1.0
lora_squeeze_final_segment_ratio = 1.0
```

## Post-Squeeze example

Post-Squeeze trains at the higher rank for the complete training budget and compresses once at the end:

```toml
network_dim = 8
network_alpha = 4
max_train_steps = 4000
lora_squeeze_start_dim = 64
lora_squeeze_num_squeezes = 1
lora_squeeze_train_after_final_squeeze = false
```

This produces:

```text
steps 0-4000: train at rank 64
end of run:   compress 64 -> 8
saved LoRA:   rank 8
```

There are no optimizer updates at rank 8 in this configuration.

## Cont-Squeeze example

Cont-Squeeze performs one rank reduction and then continues training at the deployment rank:

```toml
network_dim = 8
network_alpha = 4
max_train_steps = 4000
lora_squeeze_start_dim = 64
lora_squeeze_num_squeezes = 1
lora_squeeze_train_after_final_squeeze = true
```

With equal step distribution:

```text
steps    0-2000: rank 64
steps 2000-4000: rank 8
```

This allows the compressed LoRA to adapt further at its final rank.

## Configuration options

| Option | Default | Purpose |
|---|---:|---|
| `lora_squeeze_start_dim` | unset | Initial training rank. Must be greater than `network_dim`. |
| `lora_squeeze_num_squeezes` | `0` | Number of rank-reduction events. |
| `lora_squeeze_train_after_final_squeeze` | `true` | Whether to train at the final target rank. |
| `lora_squeeze_rank_schedule` | `geometric` | Controls how intermediate ranks are selected. |
| `lora_squeeze_step_schedule` | `equal` | Controls how training steps are divided between ranks. |
| `lora_squeeze_alpha_schedule` | `proportional` | Controls alpha at intermediate ranks. |
| `lora_squeeze_first_segment_ratio` | `1.0` | Adjusts the relative length of the first segment. |
| `lora_squeeze_final_segment_ratio` | `1.0` | Adjusts the relative length of the final segment. |
| `lora_squeeze_optimizer_mode` | `per_squeeze` | Controls optimizer-state continuity across rank changes. |
| `lora_squeeze_scheduler_mode` | `global` | Controls LR-scheduler continuity across rank changes. |

## Rank schedules

`lora_squeeze_rank_schedule` controls the intermediate ranks.

### Geometric

```toml
lora_squeeze_rank_schedule = "geometric"
```

Geometric scheduling aims for approximately equal compression ratios:

```text
rank(i) ≈ start_dim × (network_dim / start_dim)^(i / num_squeezes)
```

For a starting rank of 64, target rank of 8, and three squeezes:

```text
64 -> 32 -> 16 -> 8
```

### Linear

```toml
lora_squeeze_rank_schedule = "linear"
```

Linear scheduling aims for approximately equal rank differences:

```text
rank(i) ≈ start_dim - (start_dim - network_dim) × i / num_squeezes
```

Using the same configuration:

```text
64 -> 45 -> 26 -> 8
```

Intermediate values are converted to integer ranks while ensuring that every transition strictly decreases the rank.

## Step distribution

`lora_squeeze_step_schedule` controls how `max_train_steps` is divided between the training ranks.

Each rank receives a weight:

| Schedule | Weight for rank `r` |
|---|---:|
| `equal` | `1` |
| `rank_proportional` | `r` |
| `sqrt_rank_proportional` | `sqrt(r)` |
| `inverse_rank_proportional` | `1 / r` |
| `inverse_sqrt_rank_proportional` | `1 / sqrt(r)` |

Steps are allocated proportionally:

```text
steps(r) ≈ max_train_steps × weight(r) / sum(all segment weights)
```

Integer boundaries are calculated cumulatively so the segments always cover exactly `max_train_steps`.

For a `64 -> 32 -> 16 -> 8` schedule with 400 training steps:

### Equal

```toml
lora_squeeze_step_schedule = "equal"
```

```text
rank 64: 100 steps
rank 32: 100 steps
rank 16: 100 steps
rank  8: 100 steps
```

### Rank-proportional

```toml
lora_squeeze_step_schedule = "rank_proportional"
```

```text
rank 64: 213 steps
rank 32: 107 steps
rank 16:  53 steps
rank  8:  27 steps
```

This allocates more training to the more expressive high-rank stages.

The inverse schedules allocate more steps after the LoRA has been compressed. The square-root variants provide a milder version of either preference.

## Adjusting the first and final segments

The first and final segments can be emphasized independently:

```toml
lora_squeeze_first_segment_ratio = 2.0
lora_squeeze_final_segment_ratio = 3.0
```

These values multiply the corresponding segment weights before the steps are distributed.

For an equal `64 -> 32 -> 16 -> 8` schedule with 600 total steps:

```text
rank 64: 171 steps
rank 32:  86 steps
rank 16:  85 steps
rank  8: 258 steps
```

This can reserve additional time for both the initial high-rank solution and refinement at the final deployment rank.

## Alpha schedules

`network_alpha` remains the final deployment alpha.

Intermediate alpha values are calculated using `lora_squeeze_alpha_schedule`.

For:

```toml
network_dim = 8
network_alpha = 4
```

### Proportional

```toml
lora_squeeze_alpha_schedule = "proportional"
```

For rank `r`:

```text
alpha(r) = network_alpha × r / network_dim
```

This keeps `alpha / rank` constant:

```text
rank 64: alpha 32
rank 32: alpha 16
rank 16: alpha  8
rank  8: alpha  4
```

### Square-root

```toml
lora_squeeze_alpha_schedule = "sqrt"
```

For rank `r`:

```text
alpha(r) = network_alpha × sqrt(r / network_dim)
```

With `network_dim = 8` and `network_alpha = 4`:

```text
rank 64: alpha 11.314
rank 32: alpha  8.000
rank 16: alpha  5.657
rank  8: alpha  4.000
```

## Optimizer and scheduler modes

Optimizer and scheduler continuity can be configured independently.

### Default: reset optimizer, continue scheduler

```toml
lora_squeeze_optimizer_mode = "per_squeeze"
lora_squeeze_scheduler_mode = "global"
```

Each rank starts with fresh optimizer state, while the LR scheduler continues across the complete training run.

### Restart both for each rank

```toml
lora_squeeze_optimizer_mode = "per_squeeze"
lora_squeeze_scheduler_mode = "per_squeeze"
```

Each rank segment receives a fresh optimizer and a new scheduler curve based on that segment’s length. Scheduler warmup is therefore repeated for every segment.

### Continue both globally

```toml
lora_squeeze_optimizer_mode = "global"
lora_squeeze_scheduler_mode = "global"
```

Compatible optimizer state is transformed into the new LoRA factor coordinates, and the LR curve continues without restarting.

Global optimizer mode supports the major optimizer families used by sd-scripts, including PyTorch Adam-family optimizers, SGD, Adafactor, Lion, supported bitsandbytes variants, schedule-free optimizers, Prodigy, and D-Adaptation variants.

Unknown or incompatible optimizer state is rejected before training begins. `per_squeeze` mode can be used with any optimizer.

Optimizers that manage their own learning-rate or averaging schedule require optimizer and scheduler modes to match.

## Supported networks

Initial support is provided for:

- `networks.lora`
- `networks.lora_anima`

Supported factors include:

- Standard Linear LoRA factors.
- Conv2d LoRA factors with a 1×1 `lora_up`.
- LoRA-C3Lier when `conv_dim` and `conv_alpha` match the final `network_dim` and `network_alpha`.

All participating LoRA modules must use the same current rank and alpha.

The following are not currently supported:

- Block-specific or regex-specific ranks and alphas.
- Split-QKV or `ModuleList` LoRA factors.
- LoRA-FA.
- Multi-process or DeepSpeed training.
- `torch.compile`.
- `dim_from_weights`.
- `initial_step` and `initial_epoch`.

When `network_weights` is used, the loaded weights must match the current scheduled rank and alpha. For a new run this is the starting rank; for a resumed run it is the rank recorded in the saved LoRA-Squeeze state.

## Resume and observability

LoRA-Squeeze state is saved in `train_state.json`, including the generated rank schedule, step boundaries, current rank and alpha, and completed squeeze count.

Resumed runs continue from the saved absolute optimizer step and preserve the original squeeze boundaries and total `max_train_steps` budget.

The current schedule and progress are also written to model metadata and training logs. Transition logs include:

- Rank and alpha used for the completed training step.
- Newly installed rank and alpha.
- Retained spectral energy.
- Numerical rank.
- Optimizer-state transfer statistics.

## Implementation

Compression uses reduced QR decompositions and SVD of a rank-sized interaction core. It therefore avoids constructing the full model-sized LoRA weight update.

The feature is separated into schedule, network protocol, compression, optimizer-state, and training-lifecycle modules. Network modules explicitly declare support and expose the LoRA factors they allow to be replaced.

## Documentation and tests

A bilingual English/Japanese guide documents the available configurations, supported networks, optimizer behavior, resume semantics, and limitations.

Test coverage includes schedule generation, Linear and Conv2d compression, repeated squeezes, optimizer and scheduler modes, resume state, metadata, Accelerate integration, and supported network protocols.

### DIFF
--- a/.ai/context/01-overview.md
+++ b/.ai/context/01-overview.md
@@ -23,6 +23,11 @@ The shared training utilities (formerly a single large `train_util.py`) are spli
 | `library/args.py` | All `add_*_arguments` parser registrations, `verify_*` validation, `read_config_from_file` |
 | `library/accelerator_setup.py` | `prepare_accelerator`, `prepare_dtype`, `HIGH_VRAM` flag |
 | `library/optimizer.py` | `get_optimizer` dispatcher (AdamW / 8-bit / Lion / DAdaptation / Prodigy / Adafactor / schedule-free), LR schedulers |
+| `library/lora_squeeze_schedule.py` | LoRA-Squeeze schedules, progress budgets, validation, and resume serialization |
+| `library/lora_squeeze_network.py` | LoRA-Squeeze network-facing protocols and standard Linear/Conv2d factor validation/replacement |
+| `library/lora_squeeze_compression.py` | LoRA-Squeeze protocol/network validation, optimizer parameter-layout validation, and SVD compression |
+| `library/lora_squeeze_optimizer.py` | Optimizer-specific LoRA-Squeeze state transformations |
+| `library/lora_squeeze_training.py` | LoRA-Squeeze runtime integration, training budgets, metadata/logging, optimizer/scheduler rebuilding, and Accelerate lifecycle |
 | `library/model_io.py` | Checkpoint loading (`load_target_model`), model hashes, safetensors metadata, SAI model spec |
 | `library/checkpoint_io.py` | Save / rotate checkpoints and train state (`save_sd_model_on_*`, filename templates) |
 | `library/loss.py` | Per-step loss building blocks: timestep sampling, noise, `conditional_loss` |
--- a/docs/lora_squeeze.md
+++ b/docs/lora_squeeze.md
@@ -1,0 +1,299 @@
+# LoRA-Squeeze training / LoRA-Squeeze学習
+
+LoRA-Squeeze trains a LoRA at a higher starting rank and compresses its effective weight update into progressively lower ranks during the same training run. The implementation uses the memory-efficient QR and core-SVD transformation described in [LoRA-Squeeze](https://arxiv.org/abs/2602.10993), without materializing the full model-sized weight update.
+
+The final `network_dim` and `network_alpha` remain the deployment rank and alpha. `lora_squeeze_start_dim` selects the higher initial training rank.
+
+The implementation follows sd-scripts' library boundaries: CLI registration is in `library/args.py`, scheduling and resume state are in `library/lora_squeeze_schedule.py`, network-facing factor protocols are in `library/lora_squeeze_network.py`, compression is in `library/lora_squeeze_compression.py`, optimizer mathematics are in `library/lora_squeeze_optimizer.py`, and runtime integration, metadata/logging, and Accelerate lifecycle handling are in `library/lora_squeeze_training.py`.
+
+<details>
+<summary>日本語</summary>
+
+LoRA-Squeezeは高い初期ランクでLoRAの学習を開始し、同じ学習実行中に有効な重み更新を段階的に低いランクへ圧縮します。この実装では、モデルと同じ大きさの重み更新全体を生成せず、[LoRA-Squeeze](https://arxiv.org/abs/2602.10993)で説明されているメモリ効率の良いQR分解とコアSVD変換を使用します。
+
+最終的な`network_dim`と`network_alpha`は、配布時に使用するランクとalphaです。`lora_squeeze_start_dim`で学習開始時の高いランクを指定します。
+
+実装はsd-scriptsのライブラリ境界に従っています。CLI引数は`library/args.py`、スケジュールと再開状態は`library/lora_squeeze_schedule.py`、ネットワーク側の因子プロトコルは`library/lora_squeeze_network.py`、圧縮処理は`library/lora_squeeze_compression.py`、オプティマイザの計算は`library/lora_squeeze_optimizer.py`、実行時統合・メタデータ・ログ・Accelerateのライフサイクル処理は`library/lora_squeeze_training.py`にあります。
+
+</details>
+
+## In-Squeeze example / In-Squeezeの例
+
+```toml
+network_dim = 8
+network_alpha = 4
+
+lora_squeeze_start_dim = 64
+lora_squeeze_num_squeezes = 3
+```
+
+This produces a `64 -> 32 -> 16 -> 8` rank schedule. The total configured `max_train_steps` is divided between the four rank stages, and training continues at rank 8 after the final squeeze.
+
+The default LoRA-Squeeze behavior is:
+
+- `lora_squeeze_rank_schedule = "geometric"`
+- `lora_squeeze_step_schedule = "equal"`
+- `lora_squeeze_train_after_final_squeeze = true`
+- `lora_squeeze_optimizer_mode = "per_squeeze"`
+- `lora_squeeze_scheduler_mode = "global"`
+- `lora_squeeze_alpha_schedule = "proportional"`
+- `lora_squeeze_first_segment_ratio = 1.0`
+- `lora_squeeze_final_segment_ratio = 1.0`
+
+<details>
+<summary>日本語</summary>
+
+上の設定では、ランクは`64 -> 32 -> 16 -> 8`の順に変化します。`max_train_steps`で設定した総ステップ数は4つのランク区間に配分され、最後の圧縮後もランク8で学習を続けます。
+
+LoRA-Squeezeのデフォルトは、ランク配置が`geometric`、ステップ配分が`equal`、最終圧縮後の学習が有効、オプティマイザが`per_squeeze`、スケジューラが`global`、alphaスケジュールが`proportional`、最初と最後の区間倍率がともに`1.0`です。
+
+</details>
+
+## Post-Squeeze and Cont-Squeeze / Post-SqueezeとCont-Squeeze
+
+Post-Squeeze trains at the source rank for the full training budget and performs one final compression before saving:
+
+```toml
+network_dim = 8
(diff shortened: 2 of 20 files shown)


kohya-ss:
Thank you very much for this contribution, @jlasz. The engineering quality here is genuinely impressive — the QR + core-SVD compression that avoids materializing the full weight update, the balanced factor re-parameterization for continued training, the optimizer-state projection across rank changes, and the thorough tests and bilingual docs all reflect a lot of careful work. I really appreciate the effort.

After reviewing it carefully, though, I've decided not to merge this PR, for two main reasons:

**1. Maintenance cost.** This adds roughly 4,300 lines of new library code across six modules. In particular, the optimizer-state transfer path depends on the internal state formats of many optimizer families (Adam variants, Lion, Adafactor, Prodigy, D-Adaptation, schedule-free, bitsandbytes). That surface is tightly coupled to code that evolves frequently in this repo, and keeping it correct over time would be a significant ongoing burden.

**2. Unclear benefit for diffusion models.** The underlying idea — train at a higher rank and compress — is well motivated, but the paper's evidence is on text/VLM tasks. There's no empirical comparison here showing that, for an actual SD/SDXL/FLUX LoRA at an equal step budget, squeeze-to-target beats training directly at the target rank. Without that, it's a large amount of complexity resting on an unverified assumption for this domain.

That said, I find the core idea very interesting, and I don't want it to be lost. sd-scripts already has `networks/resize_lora.py`, which performs the same underlying operation (truncated-SVD low-rank approximation of the LoRA update). I'm considering adding a lightweight "continued-training-friendly" option to it — essentially the balanced `sqrt(S)` factor split and an explicit target alpha — so that the train-high → compress → continue-training workflow (your Cont-/In-Squeeze) can be reproduced with the existing training path and `--network_weights`, at a fraction of the maintenance cost. Your PR was what made this direction clear to me, so thank you again.

I'm going to close this for now, but I'd genuinely welcome your thoughts on that lighter-weight direction — and if you're able to share any A/B results on diffusion models, that would be very valuable.

jlasz:
Thank you, I really appreciate that, and I understand your concerns.

Would a reduced version of this be more acceptable if I removed the optimizer-state transfer path? I could also look into reworking the feature around resize_lora.py instead.

Personally, I do think the feature is promising, but I have not experimented with it thoroughly enough yet. Most of my testing so far has been on Anima.

Empirically, the main thing I have seen so far, is that at the same final rank, squeeze-trained LoRAs use the available rank much more evenly than direct target-rank training (roughly 1.5x higher mean effective/stable rank).

I can also prepare a more well-documented series of A/B tests. Is there anything in particular you would like to see?

